PR #1, opened 2024-07-26
rename BabylonLabs dependencies

SebastianElvis:
NOTE: `babylonchain` occurences in README can be renamed only after we finished all migrations.


KonradStaniec (commented):
So the only non-migrated parts I found (except readmes) are:
- ci jobs https://github.com/babylonlabs-io/babylon/blob/dev/.github/workflows/ci.yml#L10
- and wasmbining in test code -  https://github.com/babylonlabs-io/babylon/blob/dev/wasmbinding/testdata/Cargo.toml#L11

I suppose those can be migrated in separate pr though ?

RafilxTenfen (approved):
Tested a few deployments script and worked :100:

KonradStaniec (approved):

gitferry (approved):
Nice start!

PR #4, opened 2024-07-30
feat: e2e upgrade btc headers

RafilxTenfen:
Add upgrade to insert BTC headers

Should be merged after #16


app/upgrades/launchsignet/data/btc_headers.json
SebastianElvis, line 1: I know this PR is still WIP, but still wanted to ask: is it okay to have the BTC headers in Json rather than in Go code?  The json could not be a part of the binary of the Go code, so the upgrade no longer depends on the new binary, but also this json file. Also the path might be hardcoded in the code and it's not intuitive for people to move the json file to the right path and conduct upgrade

As an example, Osmosis wraps a lot of data into Go code https://github.com/osmosis-labs/osmosis/blob/main/app/upgrades/v4/prop12_data.go

RafilxTenfen: yeap, I was trying to use as json, but in the upgrade we start to have issues like 
```shell
panic: open /home/babylon/data/btc_headers.json: no such file or directory

goroutine 1 [running]:
github.com/babylonlabs-io/babylon/app/upgrades/signetlaunch.CreateUpgradeHandler.func1({0x3ddc6e8?, 0xc000c30700?}, {{0xc003726490, 0xd}, {0x0, 0x0, 0x0}, 0x19, {0xc0037264a0, 0x8}, ...}, ...)
        github.com/babylonlabs-io/babylon/app/upgrades/signetlaunch/upgrades.go:44 +0x19a
```

which since it is a json it does not compiles together with go... So I might need to research a good way to convert json to go code

RafilxTenfen: Thanks for the example, having it as a string might work pretty well, I will test that out

KonradStaniec: a bit of higher level comment, while working on this epic we need to define data format we expect from other tools to make the job of chain upgrade as easy as possible, and later build some tooling to automate transforming data we get into usable upgrade.

If we define it as string in go file, then this will make job of data transformation from indexer to upgrade a bit tricky as now we will need to:

1.  run indexer/bitcoind to get required block headers
2. retrieve those headers from indexer, probably by querying api and save it into json
3. somehow transform this json file/files into this go file with this long string

Ideally whole pipeline would be fully automated.

And this is only about btc headers, there is so much other data to collect. Some of this data will be collected by web services team (messages signed by stakers), and it would be great if we can transfrom it relatively easily into the upgrade.

tldr; we can for now leave this string in e2e test to progress forward, but ultimately we need to define:
- in what format we expect all this data to come to core team
- transformation function from data format to viable upgrade
- automate this process as much as possible

RafilxTenfen: Yeap, I totally agree with this 
that is why at first I was trying to have a json, but that creates a bit more trouble to put it together with the build. But, having the data as a string was very easy to handle 

So, my plan once this e2e upgrade to insert BTC Headers is working properly is to create a script for the deployments repo that runs this upgrade with a script that gets the data from the BTC and writes the content of the file `app/upgrades/signetlaunch/data_btc_headers.go` to run the upgrade 

The BTC Headers are the low-hanging fruit at the moment, but the next case might be harder as it could be the finality providers, I am thinking of using the same strategy, gathering it all as json, put in a file as a string and parsing it inside the upgrade itself 

What do you guys think?

LICENSE
```diff
@@ -8,10 +8,10 @@ License text copyright (c) 2017 MariaDB Corporation Ab, All Rights Reserved.
 
 Parameters
 
-Licensor:             Babylon Labs, Ltd.
+Licensor:             BabylonLabs, Ltd.
```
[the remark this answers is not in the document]
RafilxTenfen: good catch! did not notice this change

app/upgrades/signetlaunch/upgrades.go
```diff
@@ -0,0 +1,104 @@
...
+	sort.Slice(headers, func(i, j int) bool {
+		return headers[i].Height > headers[j].Height
+	})
+
+	for _, header := range headers {
+		if err := header.Validate(); err != nil {
+			return err
+		}
+	}
+	k.InsertHeaderInfos(ctx, headers)
```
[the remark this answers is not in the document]
RafilxTenfen: I think I got you, like this new inserted headers should continue from the height we have stored in the state. 
By example in the e2e test the header starts as height 0 by default in the btclightclient, so the ones inserted here should continue the height as expected ...1,2,3,4 

So, I should add a validation here to get the tip header and continue from there the verifications

app/upgrades/signetlaunch/upgrades.go
```diff
@@ -0,0 +1,104 @@
...
+	keepers *keepers.AppKeepers,
+) upgradetypes.UpgradeHandler {
+	return func(context context.Context, _plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+		ctx := sdk.UnwrapSDKContext(context)
+
+		if err := propLaunch(ctx, &keepers.BTCLightClientKeeper); err != nil {
+			panic(err)
+		}
+
+		return mm.RunMigrations(ctx, cfg, fromVM)
```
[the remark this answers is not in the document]
RafilxTenfen: https://github.com/babylonlabs-io/babylon/pull/4/commits/a56f49910f36c876222eae82b98f0921cbcffb92

LICENSE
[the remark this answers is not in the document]
RafilxTenfen: https://github.com/babylonlabs-io/babylon/pull/4/commits/801ac1f20eab6f6fc0ff85cc8c52ac8cd2d15214

app/upgrades/signetlaunch/upgrades.go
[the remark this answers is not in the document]
RafilxTenfen: Or just use the function you pointed, way easier '-'
Modified at https://github.com/babylonlabs-io/babylon/pull/4/commits/3c9ae6132c910d3ffed4c9d732bbcb68d0958ea3

app/upgrades/signetlaunch/upgrades_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+				// ensure the headers were inserted as expected
+				for i, btcHeaderInserted := range btcHeadersInserted {
+					btcHeaderInState := allBtcHeaders[oldHeadersLen+i]
+
+					s.EqualValues(btcHeaderInserted.Hash.MarshalHex(), btcHeaderInState.Hash.MarshalHex())
+					s.EqualValues(btcHeaderInserted.Header.MarshalHex(), btcHeaderInState.Header.MarshalHex())
+					s.EqualValues(btcHeaderInserted.Height, btcHeaderInState.Height)
+
+					// TODO: check why work does not match
```
RafilxTenfen, line 109: @KonradStaniec by using [`k.InsertHeaders`](https://github.com/babylonlabs-io/babylon/blob/3c9ae6132c910d3ffed4c9d732bbcb68d0958ea3/app/upgrades/signetlaunch/upgrades.go#L107) to insert the BTC Headers, the work value does not match with the work retrieved from BTC https://github.com/babylonlabs-io/babylon/blob/3c9ae6132c910d3ffed4c9d732bbcb68d0958ea3/app/upgrades/signetlaunch/data_btc_headers.go#L3

Is that expected?

KonradStaniec: uh good we cought it no.

So in structure:
```
    {
      "header": "00000020533b53ded9bff4adc94101d32400a144c54edc5ed492a3b26c63b2d686000000b38fef50592017cfafbcab88eb3d9cf50b2c801711cad8299495d26df5e54812e7914d5fae77031ecfdd0b00",
      "hash": "00000032bb881de703dcc968e8258080c7ed4a2933e3a35888fa0b2f75f36029",
      "height": 2,
      "work": "77414720"
    },
    {
      "header": "000000202960f3752f0bfa8858a3e333294aedc7808025e868c9dc03e71d88bb320000007765fcd3d5b4966beb338bba2675dc2cf2ad28d4ad1d83bdb6f286e7e27ac1f807924d5fae77031e81d60b00",
      "hash": "000000e8daac2a2e973ecaab46dc948181c638adecf5ae0fd5d3e13aa14364b0",
      "height": 3,
      "work": "77414720"
    },
```

work should be total work of BTC chain at each given height, not just work of given header.  So if we have two headers from this example the work of the second one should be `2 *  77414720`

Given that, maybe it is better if data to upgrade will contain raw headers ie.
```
   "header": "000000202960f3752f0bfa8858a3e333294aedc7808025e868c9dc03e71d88bb320000007765fcd3d5b4966beb338bba2675dc2cf2ad28d4ad1d83bdb6f286e7e27ac1f807924d5fae77031e81d60b00"
```

Then btc light client module will automatically calculates all the work for all the height and make sure all validations are passed as expected ?

RafilxTenfen: got it, makes sense 
I will update, that also requires modification at the generation of this headers at the `sid` btc indexer

RafilxTenfen: done at https://github.com/babylonlabs-io/babylon/pull/4/commits/e10c0324cae4bd87f0c6ec80fe044c39f5a89e5f

app/upgrades/signetlaunch/upgrades.go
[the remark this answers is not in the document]
RafilxTenfen: staking indexer updated at https://github.com/babylonlabs-io/staking-indexer/pull/13

app/upgrades/signetlaunch/upgrades.go
```diff
@@ -0,0 +1,107 @@
...
+		return errors.New("no headers to insert")
+	}
+
+	headersBytes := make([]bbn.BTCHeaderBytes, len(btcHeaders))
+	for i, btcHeader := range btcHeaders {
+		h := btcHeader
+		headersBytes[i] = *h.Header
+	}
+
+	if err := k.InsertHeaders(ctx, headersBytes); err != nil {
```
KonradStaniec, line 97: one note here, this will trigger also:
- emitting of the events
- call the hook light client hooks

Events sounds safe but hooks are a trickky imo as it means that we can modify other modules store. 

Wydt about adding  function to the keeper which will:
- perform all standard validations for headers, probably by using `func (l *BtcLightClient) InsertHeaders(readStore BtcChainReadStore, headers []*wire.BlockHeader) (*InsertResult, error) {` function 
- insert headers to the store without emitting any events or calling hooks
- return error if for some reason we will hit rollback

KonradStaniec (commented):
one more comment.

Though e2e test for some reason fail.

x/btclightclient/keeper/keeper.go
```diff
@@ -99,13 +99,49 @@ func (k Keeper) insertHeaders(
 		k.triggerRollBack(ctx, result.RollbackInfo.HeaderToRollbackTo)
 	}
 
+	k.insertHeadersFromInsertResult(ctx, result, true)
+	return nil
+}
+
+func (k Keeper) insertHeadersNoEventsAndRollback(
```
KonradStaniec, line 106: tbh there are 2 things about this approach which I do not like:
1. Part of insert result is handled in ` insertHeadersNoEventsAndRollback` and part of it is handled in `insertHeadersFromInsertResult`
2. `insertHeadersFromInsertResult` takes bool argument which is something to be avoided

What do you think about:
1. Introducing internal function:
```
func (k Keeper) insertHeadersInternal(
	ctx context.Context,
	headers []*wire.BlockHeader,
	handleInsertResult func(ctx context.Context, s headersState, result *types.InsertResult) error,
) error {

	headerState := k.headersState(ctx)

	result, err := k.bl.InsertHeaders(
		headerState,
		headers,
	)

	if err != nil {
		return err
	}

	return handleInsertResult(ctx, headerState, result)
}
```

then having two functions to create different insert result handlers:
- with hooks and events
- just with insertion and checking no rollback happened

The one with hooks and events
```
func (k Keeper) triggerEventAndHandleHooksHandler() func(ctx context.Context, s headersState, result *types.InsertResult) error {
	return func(ctx context.Context, s headersState, result *types.InsertResult) error {
		// if we have rollback, first delete all headers up to the rollback point
		if result.RollbackInfo != nil {
			// roll back to the height
			s.rollBackHeadersUpTo(result.RollbackInfo.HeaderToRollbackTo.Height)
			// trigger rollback event
			k.triggerRollBack(ctx, result.RollbackInfo.HeaderToRollbackTo)
		}

		for _, header := range result.HeadersToInsert {
			h := header
			s.insertHeader(h)
			k.triggerHeaderInserted(ctx, h)
			k.triggerRollForward(ctx, h)
		}
		return nil
	}
}

```

and the insertion one:
```
func (k Keeper) insertHandler() func(ctx context.Context, s headersState, result *types.InsertResult) error {
	return func(ctx context.Context, s headersState, result *types.InsertResult) error {
		// if we have rollback, first delete all headers up to the rollback point
		if result.RollbackInfo != nil {
			return fmt.Errorf("rollback should not happend %+v", result.RollbackInfo)
		}

		for _, header := range result.HeadersToInsert {
			h := header
			s.insertHeader(h)
		}
		return nil
	}
}
```

Then `insertWithHookAndEvents` becomes:
```
func (k Keeper) insertHeadersWithHookAndEvents(
	ctx context.Context,
	headers []*wire.BlockHeader) error {

	return k.insertHeadersInternal(
		ctx,
		headers,
		k.triggerEventAndHandleHooksHandler(),
	)
}
```

and option without events will be similiar but with different handler. 

This way both problems are fixed:
- handling the result is always in one function
- there is no bools

RafilxTenfen: understood, addressed at https://github.com/babylonlabs-io/babylon/pull/4/commits/c3de831e686256b621ad004f8f348c8b95c8b647 

Thanks for the suggestion

RafilxTenfen:
https://github.com/babylonlabs-io/babylon/pull/10
Would help the rate limit at integration-tests

KonradStaniec (approved):
looks good! thanks for all the updates! 👍 

We need to figure out though why the e2e tests are failing.

gitferry (commented):
Maybe I was missing some important context. Why do we want to upgrade btc headers on signet? Do we want to cover the case where we launch testnetwork upgraded from v0.9.0 on mainnet?

KonradStaniec:
>Maybe I was missing some important context. Why do we want to upgrade btc headers on signet? Do we want to cover the case where we launch testnetwork upgraded from v0.9.0 on mainnet?

Yup exactly that, there will be big passage of time between TGE version and phase2 version, so if we are doing upgrade already it is good occasion to insert a lot of headers at once instead of burning fees by reporter.

gitferry:
we have base header in the parameters. Can we cover this case by changing this parameter?

.github/workflows/ci.yml
```diff
@@ -6,20 +6,107 @@ on:
...
+          fi
+
+
+# docker_pipeline:
+#   uses: babylonlabs-io/.github/.github/workflows/reusable_docker_pipeline.yml@v0.2.1
+#   secrets: inherit
+#   with:
+#    publish: false
+#    dockerfile: ./contrib/images/babylond/Dockerfile
+#    repoName: babylond
```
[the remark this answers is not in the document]
RafilxTenfen: tests to fix e2e :cry:

RafilxTenfen:
> > Yup exactly that, there will be big passage of time between TGE version and phase2 version, so if we are doing upgrade already it is good occasion to insert a lot of headers at once instead of burning fees by reporter.
> 
> @KonradStaniec we have a base btc header in the parameters. Can we cover this case by changing this parameter? As for TGE, there won't be any btc headers relayed to Babylon. The db for storing btc headers is empty

I thought we needed the BTC headers to be included for the BTC delegations done in the past (phase-1) be valid

gitferry:
> I thought we needed the BTC headers to be included for the BTC delegations done in the past (phase-1) be valid

Ah, this makes sense. So, basically pior launch, we need to insert headers from some height ahead of the first activation height in phase-1 to the base height of the launch. Am I understanding it correctly?

RafilxTenfen:
> > I thought we needed the BTC headers to be included for the BTC delegations done in the past (phase-1) be valid
> 
> Ah, this makes sense. So, basically pior launch, we need to insert headers from some height ahead of the first activation height in phase-1 to the base height of the launch. Am I understanding it correctly?

This was my understanding, but I might be wrong 
do we have some correlation for btc staking tx block height of inclusion and btc headers in `x/btclightclient`? 
@SebastianElvis @KonradStaniec ?

gitferry (approved):
Great work!

app/test_helpers.go
```diff
@@ -354,3 +357,22 @@ func initAccountWithCoins(app *BabylonApp, ctx sdk.Context, addr sdk.AccAddress,
...
+	}`
+	buff := bytes.NewBufferString(btcHeaderZeroStr)
+
+	err := cdc.UnmarshalJSON(buff.Bytes(), &btcHeaderZero)
+	if err != nil {
+		return nil, err
+	}
+
+	return &btcHeaderZero, nil
+}
```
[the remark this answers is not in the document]
RafilxTenfen: real, sorry about that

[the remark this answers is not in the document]
RafilxTenfen: https://github.com/babylonlabs-io/babylon/pull/4/commits/6a4027db37a9c53969d74f91a94d47b37829fbe6

PR #12, opened 2024-08-06
feat(ADR-024): Introduce epoch number in public randomness commit

### LINKED ISSUE #8: BTC-timestamping public randomness

gitferry:
This PR closes #8. Specifically, this PR
* added `epoch_num` to public randomness commit
* upon receiving a new public randomness commit, the current epoch number will be fetched and stored in db
* upon receiving a finality signature, it will check whether the committed epoch has been finalized. reject the finality sig if so.
* fixed tests

A subsequent PR will exclude assigning voting power to finality providers that do not have finalized randomness for a particular block


x/finality/types/expected_keepers.go
```diff
@@ -17,6 +17,7 @@ type BTCStakingKeeper interface {
 	GetBTCStakingActivatedHeight(ctx context.Context) (uint64, error)
 	GetVotingPowerDistCache(ctx context.Context, height uint64) (*bstypes.VotingPowerDistCache, error)
 	RemoveVotingPowerDistCache(ctx context.Context, height uint64)
+	GetEpoch(ctx context.Context) uint64
```
KonradStaniec, line 20: I would say that instead adding this function to `BTCStakingKeeper` , finality provider keep should now have dependency to `EpochingKeeper` as it is epoching keeper job to inform other modules about epochs

x/finality/keeper/msg_server.go
```diff
@@ -206,6 +214,7 @@ func (ms msgServer) CommitPubRandList(goCtx context.Context, req *types.MsgCommi
 		StartHeight: req.StartHeight,
 		NumPubRand:  req.NumPubRand,
 		Commitment:  req.Commitment,
+		EpochNum:    ms.BTCStakingKeeper.GetEpoch(ctx),
```
KonradStaniec, line 217: So this is breaking change as described in https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/10945276/Babylon+Chain+upgrading+guidelines+Ready+for+review#Changing-DB-schema-of-the-existing-module

So in theory we should now do all the steps described here (bump consensus version, write migration etc)

But the thing is, we already know that database of the chain after TGE version will be almost empty (i.e only few transfers and module paramters), so maybe we can get away without doing this whole dance.

Could you do small reality check ie:
1. Create binary from TGE version https://github.com/babylonlabs-io/babylon/releases/tag/v0.9.0
2. Create binary from version in this pr
3. Sync few blocks on TGE version 
4. then try to do upgrade to the pr version by manual update - https://docs.cosmos.network/main/build/tooling/cosmovisor#adding-upgrade-binary
5. try to commit some randomness 

If this works this will prove, we do not need to write migrations/bump consensus versions for breaking changes in modules which db will be empty.

x/finality/types/expected_keepers.go
gitferry: Good point. I think this also makes sense to move `GetLastFinalizedEpoch` to `CheckpointingKeeper`

x/finality/keeper/msg_server.go
SebastianElvis: Why do we need to bump consensus version for this? This is to dev branch and the version is not necessarily deployed on any devnet/testnet/mainnet. So maybe we only need to increment the consensus version if 1) we have a new tagged version, and 2) this new tagged version makes breaking changes to consensus

Also, can such a reality check be wrapped into a CI? Tbh I don't think I can do this properly in a manual way :sweat:

KonradStaniec: >Why do we need to bump consensus version for this?

Because this is breaking change to the module schema and if this change is introduced it will break consensus. Remember, at this point we already 'released' mainnet (v0.9.0) and what we are working on is actually upgrade to this mainnet version.

>Also, can such a reality check be wrapped into a CI? Tbh I don't think I can do this properly in a manual way 

Ideally it would be. I imagine that:
- in this pr @gitferry will do the check manually
- after that all the steps will be written on confluence
- then we can dockerize this steps and automate them

SebastianElvis: > Remember, at this point we already 'released' mainnet (v0.9.0) and what we are working on is actually upgrade to this mainnet version.

But this version is not going to be deployed as a new network that inherits the ledger of the "mainnet". So from user's pov this "mainnet" is not upgraded upon this PR is merged

SebastianElvis: > For this specific version we probably don't need to bump the consensus version because when we run v0.9.0, the db related to btc staking would be empty. I need to do the reality checks to verify.

I was saying from v0.9.0 to phase-2 mainnet, the consensus version should only increment *once*. This is because none of the versions in between will go live. 

That is, even if this PR has breaking changes to the DB, if the consensus version is 1 bigger than the value in v0.9.0, then this PR should not increment the consensus version. As this PR is somehow "overrides" or "replaces" the version prior to this PR.

> I don't get it. Wouldn't the current version we are working on be the upgrade of v0.9.0?

Yes, but this does not mean every state-breaking PR needs to increment the consensus version. Rather, before the next public network (i.e., phase-2 mainnet), we keep the same consensus version which is 1 bigger than that in v0.9.0, and every PR overwrites the upgrade handler of the prior PR.

Maybe I'm missing some important context?

KonradStaniec: >But this version is not going to be deployed as a new network that inherits the ledger of the "mainnet". So from user's pov this "mainnet" is not upgraded upon this PR is merged

I also do not understand.

Ultimately, due to this pr ( and probably bunch of others) the next version we release, lets call it `v0.10.0`, will be breaking when compared to `v0.9.0` this means few things:
1. Everybody wanting to sync whole chain from block 1 of the mainnet will need to have 2 binaries: `v0.9.0` binary and `v0.10.0` binary and cosmovisor to to the proper swap of binaries at proper upgrade block
2. Now we have two ways of doing devnets for testing:
- we can deploy devnet of only version v0.10.0 and do not have any governance proposal, then there is not binary switch and we are only testing latest version
- we can deploy devnet on which we produce few blocks on v0.9.0 and then then do the governance  proposal to switch to v0.10.0, this way we test switch to and latest version
3. All this makes it necessary to properly upgrade module consensus versions
4. One observation we can make is that, when working on this next version v0.10.0 we only need to update consensus version only once with first breaking change, as we only do one upgrade from v0.9.0 to v0.10.0

KonradStaniec: >Yes, but this does not mean every state-breaking PR needs to increment the consensus version

I think we agree on this. Only first state breaking pr will need to:
- bump consensus version
- create migration function
- register migration function

Later state breaking pr-s, will need to only modify the migration function to adjust it for the made changes

gitferry: @SebastianElvis @KonradStaniec Thanks for the discussions. Makes a lot of sense! Let's wrap up the above to the upgrade guidelines on confluence as it is slightly out of scope of this pr imo.

SebastianElvis: OK, I think now I get it. So

- Consensus version of a module is incremented at most once between every two consecutive major versions. 
- This increment is needed only when the new major version introduces breaking changes to the given module.
- The increment happens at the 1st PR that introduces the 1st breaking change between the two consecutive major versions.

Is that correct? If yes then this sounds good to me. 

Another thing is that Cosmos SDK uses consensus version to [identify](https://github.com/cosmos/cosmos-sdk/blob/main/docs/learn/advanced/15-upgrade.md#overwriting-genesis-functions) the state migration function. This basically means that upon incrementing a consensus version, one needs to define a new state migration function for migrating the prior version to the new version. Will we follow this practice or we will only maintain 1 big state migration function for each module?

KonradStaniec: >Is that correct? If yes then this sounds good to me.

Yup, I will clarify it in guidelines so that we are on the same page

>  Will we follow this practice or we will only maintain 1 big state migration function for each module?

I would follow this practice tbh. This creates nice history of upgrades between versions and makes the code clearer imo.

So after phase-2 launches there will be some modules that will have `migrate1to2` function registred

And after phase-3 upgrade, some modules will have `migrate1to2` function, and some modules will have `migrate1to2` and `migrate2to3` registered.

x/btcstaking/keeper/keeper.go
```diff
@@ -38,7 +37,6 @@ func NewKeeper(
 
 	btclcKeeper types.BTCLightClientKeeper,
 	btccKeeper types.BtcCheckpointKeeper,
-	ckptKeeper types.CheckpointingKeeper,
```
SebastianElvis, line 41: We will have a [design](https://github.com/babylonlabs-io/pm/blob/main/adr/adr-024-timestamping-public-randomness.md#decision) for not assigning voting power to an FP if it does not have finalised pub rand at this height. Will this be implemented in a separate PR?

Also this seems to mean that the voting power table update will need to be moved to the finality module, as the voting power dist algorithm will depend on the pub rand KV store which is in finality module

gitferry: > We will have a [design](https://github.com/babylonlabs-io/pm/blob/main/adr/adr-024-timestamping-public-randomness.md#decision) for not assigning voting power to an FP if it does not have finalised pub rand at this height. Will this be implemented in a separate PR?

Exactly. This PR is to kick off the feature and focus on db schema change.

> Also this seems to mean that the voting power table update will need to be moved to the finality module, as the voting power dist algorithm will depend on the pub rand KV store which is in finality module

Would it cause any concerns if we let the btc staking keeper have some expected interface of finality keeper to this end?

SebastianElvis: Hmm ok, should this PR go to a feature branch then? as the PR itself might let a bunch of blocks unfinalised until we change the criteria of having voting power (and thereby definition of activated height)

> Would it cause any concerns if we let the btc staking keeper have some expected interface of finality keeper to this end?

I think if we do this, then BTC staking module is not repsonsible for inferring voting power of FPs anymore. Rather its sole functionality is to manage state FPs / BTC delegations. Voting power table then turns to be like a "view" of finality module over the BTC staking module.

There is an ancient discussion on this https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/5878176/ADR-018+BTC+staking+consumer+module+refactoring?focusedCommentId=5878896 and it seems that the timestamping pub rand feature mandates this refactor

gitferry: > Hmm ok, should this PR go to a feature branch then? as the PR itself might let a bunch of blocks unfinalised until we change the criteria of having voting power (and thereby definition of activated height)

My plan is to have this pr as the feature pr which collects all the sub-prs (defined in the TODO)

> I think if we do this, then BTC staking module is not repsonsible for inferring voting power of FPs anymore. Rather its sole functionality is to manage state FPs / BTC delegations. Voting power table then turns to be like a "view" of finality module over the BTC staking module.

Need to think about this more but seems this is out of scope of this feature? Better to have an RFC about this. Wdyt?

KonradStaniec: >There is an ancient discussion on this https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/5878176/ADR-018+BTC+staking+consumer+module+refactoring?focusedCommentId=5878896 and it seems that the timestamping pub rand feature mandates this refactor

Oh I am big fan of such refactoring, to treat `btcstaking` module as a mostly dumb storage for fps and delegations. Though I agree that combining big security related feature and such refactoring is no go. 

>Would it cause any concerns if we let the btc staking keeper have some expected interface of finality keeper to this end?

Ultimately it may work, but we already have finality keeper depending on btc staking keeper, so if we would introduce dependency of finality keeper on btc staking keeper it would create circular dependency. Such cases usually mean there is some design problem and can lead to subtle bugs, for example, one needs to remember which dependency must be initialised first. 

Ideally we would end up with some implementation without such circular dependency and make refactoring later, though I wonder whether it is possible.

gitferry: I don't think circular dependency is an issue because we have used `expected_interfaces` to avoid go module dependency, i.e., the finality module does not directly depend on the btc staking module.

KonradStaniec: >I don't think circular dependency is an issue because we have used expected_interfaces to avoid go module dependency, i.e., the finality module does not directly depend on the btc staking module.

Tbh I do not have in mind exact golang module dependency, but rather design dependency. If we now start to retrieve data from finality module in staking module, then it means:
- finality module relies on data from staking module
- staking module relies of data from finality module

This is circular and with time it can lead to some problems.

SebastianElvis: In fact, this circular dependency will be catched by compiler -- the constructor functions of two keepers will depend the other keeper and `app.go` will not compile or panick due to nil pointer. Thus if we want to make activation height depend on pub rand, then moving voting power table to finality module is necessary.

So given that the timestamping of pub rand involves multiple steps, how about we do the following:

- create a feature branch for timestamping of pub rand merge this PR to the feature branch
- do a PR for moving voting power table to finality module, and this PR goes to `dev` branch (the integration team will merge this PR to the base branch as well)
- rebase the feature branch
- do a PR for excluding FPs without pub rand from voting power table at each height, and this PR goes to the feature branch

gitferry: Good point. So, we need to do three things in sequence:
1. introduce committed epoch (this PR does)
2. move the voting power table to the finality module
3. exclude fps without timestamped pub rand from the active set

I think all of the three tasks are independent and can be merged to `dev` in sequence. Another good thing is that after (1) is merged to `dev`, the finality provider side can be implemented in parallel. Wdty?

gitferry: Let's do what you proposed then just in case we have some important add-on to deploy before the whole timestamping feature is ready

KonradStaniec: > In fact, this circular dependency will be catched by compiler -- the constructor functions of two keepers will depend the other keeper and app.go will not compile or panick due to nil pointer. Thus if we want to make activation height depend on pub rand, then moving voting power table to finality module is necessary.

I think you can get around it, by not using constructor, by just setting one keeper through pointer:
```
app.BTCStakingKepper.FinalityKepper = fk
```
it is ugly but it probably works. Though, I agree lets do it in proper way. 

>create a feature branch for timestamping of pub rand, and merge this PR to the feature branch
do a PR for moving voting power table to finality module, and this PR goes to dev branch (the integration team will merge this PR to the base branch as well)
rebase the feature branch
do a PR for excluding FPs without pub rand from voting power table at each height, and this PR goes to the feature branch

This sounds good, as Babylon is always ready to be released with the current set of auxiliary programs on devnet.

KonradStaniec: One thing to remember,  we need to rememver that moving voting table from btcstaking module to finality module is also an breaking change in which in theory, we need to have upgrade handler that moves all data from `btcstaking` key space to `finality` key space. 

Of course this is only necessary if we need to do the migration even if database is empty.

SebastianElvis (approved):
Great discussion here! Code lgtm

gitferry:
I have tested the upgrade and it worked fine against this PR. Merging this PR now. Testing with public randomness will be done after the finality provider side accommodates the change

KonradStaniec (approved):

PR #13, opened 2024-08-07
Fix shadowing in retry library 

KonradStaniec:
We have regression in our retry library. 

Due to this regression all errors returned by provided functions are shadowed by error from - `jitter, err := randDuration(int64(sleep))`

This makes vigilante tests fail - https://github.com/babylonlabs-io/vigilante/actions/runs/10281209916/job/28451257642?pr=5

### DIFF
--- a/types/retry/retry.go
+++ b/types/retry/retry.go
@@ -52,7 +52,8 @@ func isExpectedErr(err error) bool {
 }
 
 // Do executes a func with retry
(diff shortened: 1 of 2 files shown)


gitferry (approved):
Maybe we can add a workflow to check shadowing?

KonradStaniec:
good point! maybe lets decide what we want to run to do that and add feature request to our pipeline repo ?

PR #22, opened 2024-08-19
zoneconcierge: replace chain ID with client ID for consumers

SebastianElvis:
Fixes https://github.com/babylonlabs-io/pm/issues/12

This PR replaces the chain ID with the client ID for consumers in the BTC timestamping protocol. This is in line with [the IBC security model](https://tutorials.cosmos.network/academy/3-ibc/1-what-is-ibc.html) where each IBC light client corresponds to a chain, and with the BTC staking integration implementation.

The PR includes:

- renaming all chain ID to consumer ID, same as BTC staking integration does
- using client ID to identify {epoch, finalised} chain info
- changing all queries, fuzz tests, e2e tests, and doc accordingly

TODO before merging the PR:

- [x] wait until @gitferry verifies we can do software upgrade with empty DB and updated DB schema

Plan after the PR:

- Fixing compatiblity with BE/FE (need arch decisions here)


KonradStaniec (approved):
looks good!

SebastianElvis:
Putting this PR on hold for now. Another way of resolving this issue is to enforce each chain ID to only have 1 IBC client in record. This is followed by [ICS](https://github.com/cosmos/interchain-security/blob/a0e1e2df2a8b88ae525529d247190e21269066e4/x/ccv/provider/keeper/proposal.go#L143-L144). Will need to investigate further and provide a concrete design

gitferry (approved):
Great work!

PR #23, opened 2024-08-19
feat(ADR-024): Exclude voting power for fps who do not have timestamped pub rand

### LINKED ISSUE #8: BTC-timestamping public randomness

gitferry:
This closes the second part of #8. In particular, this PR:
* Implemented `HasTimestampedPubRand` in the finality keeper to check whether a pub rand at a given height is BTC-timestamped
* Added expected finality keeper to btcstaking keeper so that `HasTimestampedPubRand` can be called within the module
* When signing voting power to fps in `btcstaking`'s begin blocker, if the fp does not have timestamped pub rand, voting power will not be assigned

Note that this PR introduced circular dependency between the btc staking module and finality module, which should be addressed by moving the voting power table to the finality module, tracked by issue #24


x/btcstaking/keeper/power_dist_change.go
```diff
@@ -232,6 +238,11 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 
 		// add this finality provider to the new cache if it has voting power
 		if fpDistInfo.TotalVotingPower > 0 {
+			// voting power is not assigned if it does not have timestamped
+			// public randomness for this height
+			if !k.FinalityKeeper.HasTimestampedPubRand(ctx, fpBTCPK, height) {
+				fpDistInfo.TotalVotingPower = 0
+			}
```
SebastianElvis, line 245: This filtering only applies to new FPs, but not existing FPs. So if an existing active FP stops having timestamped pub rand the algorithm will keep it in the active set, which is not expected.

A proper fix could be filtering out all FPs without timestamped pub rand in `ApplyActiveFinalityProviders`. This function provides the exact definition of when a FP is "active"

x/finality/keeper/public_randomness.go
```diff
@@ -33,6 +34,30 @@ func (k Keeper) GetPubRandCommitForHeight(ctx context.Context, fpBtcPK *bbn.BIP3
 	return nil, types.ErrPubRandNotFound
 }
 
+// GetTimestampedPubRandCommitForHeight finds the public randomness commitment that includes the given
+// height for the given finality provider
+func (k Keeper) GetTimestampedPubRandCommitForHeight(ctx context.Context, fpBtcPK *bbn.BIP340PubKey, height uint64) (*types.PubRandCommit, error) {
```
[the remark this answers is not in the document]
gitferry: I prefer to have separate methods for different usages and the implementation is embedded. `HasTimestampedPubRand` is needed to return primitive values. Otherwise, `types.PubRandCommit` will cause circular dependency between btcstaking and finality modules

x/btcstaking/keeper/power_dist_change.go
gitferry: > This filtering only applies to new FPs, but not existing FPs. So if an existing active FP stops having timestamped pub rand the algorithm will keep it in the active set, which is not expected.

The filter is also applied in Line 202 for existing fps

> A proper fix could be filtering out all FPs without timestamped pub rand in ApplyActiveFinalityProviders. This function provides the exact definition of when a FP is "active". Wdyt?

I think it's a good place to put it. One minor concern is that this might introduce overhead in the sense that we need to iterate all the fp dist info and update the voting power

SebastianElvis: > One minor concern is that this might introduce overhead in the sense that we need to iterate all the fp dist info and update the voting power

This is already the case and is inevitable. `ApplyActiveFinalityProviders` needs to find out the top N FPs in terms of voting power and the most efficient way of finding them is sorting first.

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -196,6 +197,11 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 
 		// add this finality provider to the new cache if it has voting power
 		if fp.TotalVotingPower > 0 {
+			// voting power is not assigned if it does not have timestamped
+			// public randomness for this height
+			if !k.FinalityKeeper.HasTimestampedPubRand(ctx, fp.BtcPk, height) {
```
KonradStaniec, line 202: hmm with this addition `ProcessAllPowerDistUpdateEvents` is probably in dire need of optimization, as now for every finality provider we iterate over potentially all pub rand commits of this finality provider. 

This can be even used as attack vector is we do not set `MinPubRand` high enough i.e create a lot if finality providers with minimal amount of delegation and super small random commits, to make this method run a lot of time.

x/btcstaking/keeper/power_dist_change.go
KonradStaniec: 👍  on putting it in `ApplyActiveFinalityProviders.` having this check in two separate places can lead to potential errors in the future. (and we want to refactor this area so it is better not to put landmines for our futureselves)

x/btcstaking/keeper/power_dist_change.go
gitferry: True. But not sure it has a real impact because this method is only called when the fp has voting power. We can revisit potential optimization when we do the refactoring. Added a TODO for now

gitferry:
@SebastianElvis @KonradStaniec I agree that it's better to reset the voting power in a single place but not sure we should put it in `ApplyActiveFinalityProviders` because it's probably good to inject a keeper method to a method of `VotingPowerDistCache`. Please see the update in [bee1bbc](https://github.com/babylonlabs-io/babylon/pull/23/commits/bee1bbcb66c8f7feccf8e9d193def8a784daa621)

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -113,6 +113,7 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 	events []*types.EventPowerDistUpdate,
 	maxActiveFps uint32,
 ) *types.VotingPowerDistCache {
+	height := uint64(sdk.UnwrapSDKContext(ctx).HeaderInfo().Height)
```
RafilxTenfen, line 116:
```suggestion
	height := uint64(sdk.UnwrapSDKContext(ctx).BlockHeight())
```

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -236,6 +237,17 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 		}
 	}
 
+	// set voting power to 0 if the fp does not have timestamped pub rand
+	for _, fp := range newDc.FinalityProviders {
+		// TODO calling HasTimestampedPubRand potentially iterates
```
RafilxTenfen, line 242: We could index timestamped pub rand by `fpBtcPK||height` with the return of `0` or `1` for this 
but it would increase our db size a little bit, unless we clean up after passed epoch or something like this

RafilxTenfen (approved):
No blockers from me, agree with the changes 
only concern is with iterating pub rand from the fp store

x/btcstaking/keeper/power_dist_change.go
gitferry: Not sure this will eliminate iteration because, for each new epoch, we still need to iterate every `fpBtcPk || height` to update the value

x/btcstaking/keeper/power_dist_change.go
SebastianElvis: I remember `BlockHeight` will be deprecated?

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -236,6 +237,17 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
+		// all the pub rand committed by the fp, which might slow down
+		// the process, need optimization
+		if fp.TotalVotingPower > 0 &&
+			!k.FinalityKeeper.HasTimestampedPubRand(ctx, fp.BtcPk, height) {
+			fp.TotalVotingPower = 0
+		}
```
SebastianElvis, line 248: I just realised that instead of zeroising total voting power of FP to be 0, we would have to remove the FP out from `newDc`, such that `ApplyActiveFinalityProviders` won't count FPs with 0 voting power in the active set when there are very few FPs. This is also done for slashed FPs ([ref](https://github.com/babylonlabs-io/babylon/blob/5a4d7ddae3a93c8e660e8f03fba740538b9e9c6c/x/btcstaking/keeper/power_dist_change.go#L173-L176)). So it seems that https://github.com/babylonlabs-io/babylon/blob/5a4d7ddae3a93c8e660e8f03fba740538b9e9c6c/x/btcstaking/keeper/power_dist_change.go#L173-L176 is the best and the only needed place to filter out FPs without timestamped pub rand. Wdyt? sorry for being slow here

x/btcstaking/keeper/power_dist_change.go
RafilxTenfen: oh, did not know about this

x/btcstaking/keeper/power_dist_change.go
gitferry: Thanks for the reminder! Agree that we should not count fps with 0 voting power in the active set, but I don't think we should remove those who have delegations but don't have timestamped pub rand from the dist cache. If we do so, what if those fps have pub rand suddenly have timestamped pub rand in the next height? Seems we would lose the accumulated delegations to calculate the power in a future height. I think we should fix `GetNumActiveFPs` to return the number of fps that have positive voting power. Did I miss anything?

SebastianElvis: Oh good point. Then indeed the best place to fix this is `ApplyActiveFinalityProviders`. Agree that `GetNumActiveFPs` is probably the place as this function only returns the number of active FPs, although the function name becomes a bit confusing after this change. So how about we 

- refactor `GetNumActiveFPs` as a private keeper function that is only triggered once (It iterates over FPs in DB so is expensive)
- persist the number of active FPs returned by this private keeper function in the `VotingPowerDistCache`
- In `GetActiveFinalityProviderSet` we retrieve `numActiveFPs` from the `dc` object directly, rather than calling the private keeper function

x/btcstaking/keeper/power_dist_change.go
gitferry: @SebastianElvis is correct https://github.com/cosmos/cosmos-sdk/blob/c4d9a495052b78a02723ed1df1d336efe83a4e8d/types/context.go#L45. `BlockHeight()` would return `header.height` which is deprecated, but not sure if `HeaderService` refers to `headerInfo`

SebastianElvis: > but not sure if HeaderService refers to headerInfo

It is mentioned somewhere in Cosmos SDK code but I forget where :sweat:

x/btcstaking/keeper/power_dist_change.go
gitferry: I feel that we should not change the state in a getter function like `GetNumActiveFPs`. We should have a separate method `updateNumActiveFps` for that. Please see the update in [8e14b99](https://github.com/babylonlabs-io/babylon/pull/23/commits/8e14b994a747da7ec5892d985b319b26600c7924)

SebastianElvis: I was saying that instead of having this getter function, we can define a new private keeper function (say `calcNumActiveFPs(dc *VotingPowerDistCache)`) that invokes `HasTimestampedPubRand` and determines the number of active FPs, and then use this function to get `num_active_fps` and assign it to `dc`. 

The reasoning was to avoid setting `TotalVotingPower` to zero as we do in L239-247. It's possible that an FP does not have timestamped pub rand at current height but has one in the next height. In this case the `newDc` will give 0 voting power to this FP and the `newDc` is stored in DB. This will be carried to the next height and this FP won't have voting power, which is not expected. It's also a bit hacky as this mixed the concept of voting power and active FP (a FP out of the active FP set can still have positive voting power).

gitferry: Per discussion off-line, we can't set the voting power directly to zero as when the fp has timestamped pub rand in a future height, the voting power can't be set back. We need to figure out another way to do it

KonradStaniec: oh why it can't be set back ? 

I imagine it could be done by iterating over all `BtcDels` and summing once again, though probably the problem would be when to do that. (probably it would require some kind of event for new registred pub rand though which seems like big cahnge)

gitferry: > I imagine it could be done by iterating over all BtcDels and summing once again, though probably the problem would be when to do that.

I think we intentionally avoid doing this due to the overhead

gitferry: I was thinking we could solve the issue by adding a flag `is_timestamped` to the fp dist info. We define that an fp is active at a height as long as it has positive voting power and timestamped pub rand (voting_power > 0 and is_timestamped = true). The voting power in the dist info keeps it is but we don't set the voting power to the store if `is_timestamped=true`. This way, when the fp has timestamped pub rand in a future height, we don't need to re-calculate previous delegations. Please see the change in [b47d837](https://github.com/babylonlabs-io/babylon/pull/23/commits/b47d83735bbc13cbc56ad297a5ee06369b79c21f)

x/btcstaking/keeper/power_dist_change.go
RafilxTenfen: once an epoch is better than each begin block ^^

x/btcstaking/types/incentive.go
```diff
@@ -36,21 +36,39 @@ func (dc *VotingPowerDistCache) FindNewActiveFinalityProviders(prevDc *VotingPow
...
+	for _, fp := range dc.FinalityProviders {
+		if numActiveFPs == maxActiveFPs {
+			break
+		}
+		if fp.TotalVotingPower == 0 {
+			break
+		}
+		if !fp.IsTimestamped {
+			continue
+		}
```
SebastianElvis, line 56: If this happens then that means the rest FPs are all not timestamped, so can we
break here?

x/btcstaking/types/incentive.go
```diff
@@ -36,21 +36,39 @@ func (dc *VotingPowerDistCache) FindNewActiveFinalityProviders(prevDc *VotingPow
...
+
+	numActiveFPs := uint32(0)
+
+	for _, fp := range dc.FinalityProviders {
+		if numActiveFPs == maxActiveFPs {
+			break
+		}
+		if fp.TotalVotingPower == 0 {
+			break
+		}
```
SebastianElvis, line 53: I remember `dc` will exclude FPs with zero total voting power. Is this true?

x/btcstaking/types/incentive.go
```diff
@@ -36,21 +36,39 @@ func (dc *VotingPowerDistCache) FindNewActiveFinalityProviders(prevDc *VotingPow
...
+// and records them in cache
 func (dc *VotingPowerDistCache) ApplyActiveFinalityProviders(maxActiveFPs uint32) {
-	// reset total voting power
-	dc.TotalVotingPower = 0
-	// sort finality providers
-	SortFinalityProviders(dc.FinalityProviders)
-	// calculate voting power of top N finality providers
-	numActiveFPs := dc.GetNumActiveFPs(maxActiveFPs)
+	// sort finality providers with timestamping considered
+	SortFinalityProvidersWithTimestamping(dc.FinalityProviders)
```
SebastianElvis, line 43: This function is now very specific so how about we do the sorting inside
`ApplyActiveFinalityProviders` and remove `SortFinalityProvidersWithTimestamping`?

x/btcstaking/types/incentive.go
```diff
@@ -36,21 +36,39 @@ func (dc *VotingPowerDistCache) FindNewActiveFinalityProviders(prevDc *VotingPow
...
+		totalVotingPower += dc.FinalityProviders[i].TotalVotingPower
 	}
+
+	dc.TotalVotingPower = totalVotingPower
+	dc.NumActiveFps = numActiveFPs
 }
 
 func (dc *VotingPowerDistCache) GetNumActiveFPs(maxActiveFPs uint32) uint32 {
-	return min(maxActiveFPs, uint32(len(dc.FinalityProviders)))
+	return min(maxActiveFPs, dc.NumActiveFps)
```
SebastianElvis, line 71: Is it true that `dc.NumActiveFps <= maxActiveFPs`? If yes then we can remove
`GetNumActiveFPs` function

SebastianElvis (approved):
Very good idea on having a flag for timestamped status of FPs! Lgtm with some minor comments. Thanks for the patience of discussing and addressing all points :+1:

x/btcstaking/types/incentive.go
gitferry: Yes, but we don't have any rules to enforce this. This is a nice to have

x/btcstaking/types/incentive.go
gitferry: we can but this is also used as util in test

x/btcstaking/types/incentive.go
gitferry: Good point

x/btcstaking/types/incentive.go
```diff
@@ -36,28 +36,44 @@ func (dc *VotingPowerDistCache) FindNewActiveFinalityProviders(prevDc *VotingPow
 }
 
 // ApplyActiveFinalityProviders sorts all finality providers, counts the total voting
-// power of top N finality providers, and records them in cache
+// power of top N finality providers, excluding those who don't have timestamped pub rand
+// and records them in cache
 func (dc *VotingPowerDistCache) ApplyActiveFinalityProviders(maxActiveFPs uint32) {
```
KonradStaniec, line 41: minor: can we have some separate table driven test case for this function with few possbile scenarios like:
- few finality provider timestamped, few don't
- different number of active inactive 
...

A lot is riding on this function being correct so it would be nice to have those. I can can be done in separate pr.

KonradStaniec (approved):
LGTM! 

minor thing: I would add some specific table driven tests for `ApplyActiveFinalityProviders`

x/btcstaking/types/incentive.go
gitferry: Good point. I think the `VotingPowerDisCache` in general needs comprehensive tests for which we have none. Created an issue for this #29 and will address it in a separate PR

PR #25, opened 2024-08-19
chore: e2e upgrade refactory

RafilxTenfen:
- Remove Vanilla upgrade
- Read upgrade height from the prop file
- Write the upgrade prop file with new height based on the current node status height


app/e2e_include_upgrades.go
```diff
@@ -4,9 +4,8 @@ package app
 
 import (
 	"github.com/babylonlabs-io/babylon/app/upgrades/signetlaunch"
-	"github.com/babylonlabs-io/babylon/app/upgrades/vanilla"
 )
 
 func init() {
-	Upgrades = append(Upgrades, vanilla.Upgrade, signetlaunch.Upgrade)
+	Upgrades = append(Upgrades, signetlaunch.Upgrade)
 }
```
SebastianElvis, line 11: Is it possible to move this file under `test/e2e` so that all e2e stuff is there but nowhere else?

test/e2e/software_upgrade_e2e_signet_launch_test.go
```diff
@@ -23,17 +23,22 @@ func (s *SoftwareUpgradeSignetLaunchTestSuite) SetupSuite() {
 
 	btcHeaderGenesis, err := app.SignetBtcHeaderGenesis(app.NewTmpBabylonApp().AppCodec())
 	s.NoError(err)
-	s.configurer, err = configurer.NewSoftwareUpgradeConfigurer(s.T(), false, config.UpgradeSignetLaunchFilePath, []*btclighttypes.BTCHeaderInfo{btcHeaderGenesis})
+
+	cfg, err := configurer.NewSoftwareUpgradeConfigurer(s.T(), true, config.UpgradeSignetLaunchFilePath, []*btclighttypes.BTCHeaderInfo{btcHeaderGenesis})
```
SebastianElvis, line 27: Is it a good idea to let `NewSoftwareUpgradeConfigurer` return the concrete `UpgradeConfigurer` rather than the interface?

test/e2e/configurer/upgrade.go
```diff
@@ -252,3 +260,124 @@ func (uc *UpgradeConfigurer) upgradeContainers(chainConfig *chain.Config, propHe
...
+// UpgradeFilePath returns the local full path of the upgrade file
+func (uc *UpgradeConfigurer) UpgradeFilePath() (string, error) {
+	pwd, err := os.Getwd()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(pwd, uc.upgradeJsonFilePath), nil
+}
+
+// parseGovPropFromFile loads from the file the Upgrade msg.
```
SebastianElvis, line 304:
```suggestion
// parseGovPropFromFile loads from the file and parse it to the upgrade msg.
```

SebastianElvis (approved):
Great work :+1:

e2e failed though

app/e2e_include_upgrades.go
RafilxTenfen: good question, let me try 
Since the Upgrades var is public it should not be a problem

RafilxTenfen: It didn't work, the `Upgrades` var inside the app package did not contain the SignetLaunch upgrade :/ when moving this file to another package 
Not sure why

test/e2e/software_upgrade_e2e_signet_launch_test.go
RafilxTenfen: https://github.com/babylonlabs-io/babylon/pull/25/commits/746b5839f2edd6f95d13267a39089f07e0c9b57d

test/e2e/configurer/upgrade.go
RafilxTenfen: https://github.com/babylonlabs-io/babylon/pull/25/commits/746b5839f2edd6f95d13267a39089f07e0c9b57d

app/e2e_include_upgrades.go
SebastianElvis: Might be because when starting the chain, the `test/e2e` package is not loaded, and therefore the `init` function is not executed. Eitherway np with me for keeping it as is. Maybe we can add some docstrings to clarify this

RafilxTenfen: agree, added a small comment to remove the file when we are done testing

PR #28, opened 2024-08-21
feat: upgrade signed FPs msgs

RafilxTenfen:
- Load signed `MsgCreateFinalityProvider` from json string
- Add finality providers in upgrade
- Check finality providers in e2e upgrade signet


app/keepers/keepers.go
```diff
@@ -155,6 +156,10 @@ type AppKeepers struct {
 	keys    map[string]*storetypes.KVStoreKey
 	tkeys   map[string]*storetypes.TransientStoreKey
 	memKeys map[string]*storetypes.MemoryStoreKey
+
+	// encoding/decode helpers
+	Codec            codec.Codec
+	TxEncodingConfig client.TxEncodingConfig
```
RafilxTenfen, line 162: I needed this for the upgrade to parse the signed msgs, the option was to add it here or at the [CreateUpgradeHandler](https://github.com/babylonlabs-io/babylon/blob/5f8af8ced17d24f3f0c6172293cd37fb3d055807/app/upgrades/types.go#L29)

app/upgrades/signetlaunch/upgrades.go
```diff
@@ -79,6 +101,115 @@ func LoadBTCHeadersFromData() ([]*btclighttypes.BTCHeaderInfo, error) {
...
+			return nil, err
+		}
+
+		fps[i] = fp
+	}
+
+	// sorts all the FPs by their addresses
+	sort.Slice(fps, func(i, j int) bool {
+		return fps[i].Addr > fps[j].Addr
+	})
```
RafilxTenfen, line 138: This is not strictly necessary as all the finality providers would have the same order when we compile, but I still kept it...

app/upgrades/signetlaunch/data_signed_fps.go
```diff
@@ -0,0 +1,122 @@
+package signetlaunch
+
+const SignedFPsStr = `{
```
SebastianElvis, line 3: Is it a hard requirement that the string needs to contain the full tx rather
than just the message part?

app/upgrades/signetlaunch/upgrades.go
```diff
@@ -79,6 +101,115 @@ func LoadBTCHeadersFromData() ([]*btclighttypes.BTCHeaderInfo, error) {
...
+// LoadSignedFPsFromData returns the finality providers from the json string.
+// It also verifies if the msg is correctly signed and is valid to be inserted.
+func LoadSignedFPsFromData(cdc codec.Codec, txJSONDecoder sdk.TxDecoder) ([]*btcstktypes.MsgCreateFinalityProvider, error) {
+	buff := bytes.NewBufferString(SignedFPsStr)
+
+	var d dataSignedFps
+	err := json.Unmarshal(buff.Bytes(), &d)
+	if err != nil {
+		return nil, err
+	}
```
SebastianElvis, line 113: Hmm if we only have the message part then we don't need to unmarshal things into `any`

x/btcstaking/keeper/finality_providers.go
```diff
@@ -4,13 +4,48 @@ import (
...
 	"cosmossdk.io/store/prefix"
 	"github.com/cosmos/cosmos-sdk/runtime"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/babylonlabs-io/babylon/x/btcstaking/types"
 )
 
+// AddFinalityProvider adds the given finality provider to KVStore if it has valid
+// commission and it was not inserted before
+func (k Keeper) AddFinalityProvider(goCtx context.Context, msg *types.MsgCreateFinalityProvider) error {
```
SebastianElvis, line 17: Good refactoring! Can `SetFinalityProvider` be made private then?

app/upgrades/signetlaunch/upgrades.go
```diff
@@ -79,6 +101,115 @@ func LoadBTCHeadersFromData() ([]*btclighttypes.BTCHeaderInfo, error) {
 	return gs.BtcHeaders, nil
 }
 
+// LoadSignedFPsFromData returns the finality providers from the json string.
+// It also verifies if the msg is correctly signed and is valid to be inserted.
+func LoadSignedFPsFromData(cdc codec.Codec, txJSONDecoder sdk.TxDecoder) ([]*btcstktypes.MsgCreateFinalityProvider, error) {
+	buff := bytes.NewBufferString(SignedFPsStr)
```
gitferry, line 107: Could the json string load from a file?

app/upgrades/signetlaunch/upgrades.go
```diff
@@ -79,6 +101,115 @@ func LoadBTCHeadersFromData() ([]*btclighttypes.BTCHeaderInfo, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	msgs := tx.GetMsgs()
+	if len(msgs) == 0 {
+		return nil, fmt.Errorf("invalid msg, there is no msg inside the tx %+v", tx)
+	}
+
+	// each tx should only contain one msg
```
gitferry, line 154: why? we should explain here

app/upgrades/signetlaunch/upgrades.go
```diff
@@ -79,6 +101,115 @@ func LoadBTCHeadersFromData() ([]*btclighttypes.BTCHeaderInfo, error) {
...
+
+	msgV2 := msgsV2[0]
+	signers, err := cdc.GetMsgV2Signers(msgV2)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get signers from msg %+v: %w", msg, err)
+	}
+
+	if len(signers) == 0 {
+		return nil, fmt.Errorf("no signer at msg %+v", msgV2)
+	}
```
gitferry, line 176: Should we ensure the number of signers is 1?

app/upgrades/signetlaunch/upgrades.go
```diff
@@ -79,6 +101,115 @@ func LoadBTCHeadersFromData() ([]*btclighttypes.BTCHeaderInfo, error) {
...
+
+	// sorts all the FPs by their addresses
+	sort.Slice(fps, func(i, j int) bool {
+		return fps[i].Addr > fps[j].Addr
+	})
+
+	return fps, nil
+}
+
+func parseCreateFPFromSignedTx(cdc codec.Codec, tx sdk.Tx) (*btcstktypes.MsgCreateFinalityProvider, error) {
```
gitferry, line 143: should we also check the commission rate to be not less than the min value defined in params?

app/upgrades/signetlaunch/data_signed_fps.go
KonradStaniec: I think I will attach myself to this comments as this is probably closest to my concerns.

I think the question here becomes where do we validates `bbn1gwecky0m842kvjg7mcvt9z330rz96948aplqlm` is if fact the signer of the transactions:

1. We can do it off-chain, in some process /service and here include only the messages ?
2. W can do it on-chain, so gather transactions to some database and during upgrade perform whole transaction validation including transaction signaure. To do that we would need to move all of those transaction through message router as transactions signature validation is performed in `anteHandler`. Then we would ignore the ones that fail the sig validation (though this seems like bad ux ?)

app/upgrades/signetlaunch/upgrades.go
RafilxTenfen: Nop :/ otherwise, it doesn't compile together with the binary, so you would need to also send this json file to who wants to upgrade and put it aside the binary when the upgrade is happening

app/upgrades/signetlaunch/upgrades.go
```diff
@@ -79,6 +101,115 @@ func LoadBTCHeadersFromData() ([]*btclighttypes.BTCHeaderInfo, error) {
...
+	msg, ok := msgs[0].(*btcstktypes.MsgCreateFinalityProvider)
+	if !ok {
+		return nil, fmt.Errorf("unable to parse %+v to MsgCreateFinalityProvider", msg)
+	}
+
+	if err := msg.ValidateBasic(); err != nil {
+		return nil, fmt.Errorf("error validating basic msg: %w", err)
+	}
+
+	msgV2 := msgsV2[0]
```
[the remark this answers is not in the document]
RafilxTenfen: I will add a check

app/upgrades/signetlaunch/upgrades.go
RafilxTenfen: it is what we expect, for each signer to only submit one message per transaction 
this is the format we are going to expect the finality providers submit to us

app/upgrades/signetlaunch/upgrades.go
RafilxTenfen: agree

app/upgrades/signetlaunch/data_signed_fps.go
RafilxTenfen: yes, we can do it off-chain as option 1. 
or in option 2. I am validating during the upgrade https://github.com/babylonlabs-io/babylon/blob/5215030b9b8e4d51fb63a633cc531c7e4f445224/app/upgrades/signetlaunch/upgrades.go#L177-L178

x/btcstaking/keeper/finality_providers.go
RafilxTenfen: done at https://github.com/babylonlabs-io/babylon/pull/28/commits/1b6e6bafea91db6262dc7b40ae0875ad069023ad

app/upgrades/signetlaunch/upgrades.go
RafilxTenfen: The commission rate is already check at https://github.com/babylonlabs-io/babylon/blob/1b6e6bafea91db6262dc7b40ae0875ad069023ad/x/btcstaking/keeper/finality_providers.go#L23

app/upgrades/signetlaunch/upgrades.go
RafilxTenfen: done at https://github.com/babylonlabs-io/babylon/pull/28/commits/2bb47c4f305f489af2065402ba56576a1ced60cd

app/upgrades/signetlaunch/upgrades.go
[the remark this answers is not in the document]
RafilxTenfen: https://github.com/babylonlabs-io/babylon/pull/28/commits/2bb47c4f305f489af2065402ba56576a1ced60cd

app/upgrades/signetlaunch/data_signed_fps.go
SebastianElvis: Do we really need to validate the signature of the tx signed by the signer?

It's ourselves who determine `SignedFPsStr` and we will need to gather data and verify them offline anyway. Only after we are confident `SignedFPsStr` is correct we will perform this software upgrade.

So it seems that inserting messages directly is already secure?

RafilxTenfen: Updated at https://github.com/babylonlabs-io/babylon/pull/28/commits/79a07120715aa0096c7ddf9c2537721f5627eea5

One thing to note is, that I would still expect finality providers to give us the entire signed message during the registry and we convert it to the Finality provider information as needed 

Mostly as a guarantee that the fp owns that key and is confirming that he wants to be a FP for that BTC pk, now that the PoP only covers BTC owns address :sweat_smile:

gitferry (approved):
Code-wise lgtm!

app/upgrades/signetlaunch/data_signed_fps.go
KonradStaniec: what misled me, is that the line you linked before ie
```
cdc.InterfaceRegistry().SigningContext().AddressCodec().BytesToString(signers[0])
```
does not really verify signature. To verify signature easiest way would be copy code from ante handler:
https://github.com/cosmos/cosmos-sdk/blob/main/x/auth/ante/sigverify.go#L302

> One thing to note is, that I would still expect finality providers to give us the entire signed message during the registry and we convert it to the Finality provider information as needed
> 
> Mostly as a guarantee that the fp owns that key and is confirming that he wants to be a FP for that BTC pk, now that the PoP only covers BTC owns babylon address

totally agree on that we need to verify signature for provider Babylon address. 

One things that worries me, if we miss something and add transactions with invalid signature to this string, and now we won't validate it on node level. It seems risky.  Maybe, as this whole string is part of our binary, we can have unit test that goes over whole thing and verify signatures ?

RafilxTenfen: Would You prefer to have the full signed message in the string and validate the signature during the upgrade? 
Or validate it offline and simplify the upgrade with just the Finality Provider information? 

About the `sigverify`: 
this improvement in `sigverify` is still unreleased https://github.com/cosmos/cosmos-sdk/blob/0fe3115dcf2e42dac29364a70cca71d52df2a443/x/auth/CHANGELOG.md#L38

The one we currently have loads the account from the AccountKeeper 
and this is not good for us, since the account was not created yet by not submitting any transaction by the finality provider babylon key holder before

image

RafilxTenfen: As discussed with @KonradStaniec offline, the string with finality providers will contain the entire transaction with signature, but the signatures validity, PoP and other checks will not be verified during the upgrade, only by a unit test `TestValidateSignatureSignedFPsFromData` that should make sure everything is checked for the upgrade to happen on that front

So, during the upgrade we would still need to parse the msg to `Any` to be able to cast it to the `MsgCreateFinalityProvider` after, but without validating other things

cc @SebastianElvis

RafilxTenfen: I will wait for you guys approval on this before merge

app/upgrades/signetlaunch/data_signed_fps_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/tx/signing"
+	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	authtx "github.com/cosmos/cosmos-sdk/x/auth/tx"
+	txmodule "github.com/cosmos/cosmos-sdk/x/auth/tx/config"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateSignatureSignedFPsFromData(t *testing.T) {
```
KonradStaniec, line 22: hmm we could probably also add validation if the string do not containan duplicates

app/upgrades/signetlaunch/data_signed_fps_test.go
```diff
@@ -0,0 +1,109 @@
...
+	chainID := "bbn-1"
+
+	ctx := bbnApp.BaseApp.NewContextLegacy(true, tmproto.Header{Height: 1, ChainID: chainID, Time: time.Now().UTC()})
+	buff := bytes.NewBufferString(v1.SignedFPsStr)
+	simulateTx := false
+
+	var d v1.DataSignedFps
+	err := json.Unmarshal(buff.Bytes(), &d)
+	require.NoError(t, err)
+	require.Len(t, d.SignedTxsFP, 2)
```
KonradStaniec, line 35: we probably do not need this assertion, as I imagine the length will be highly dependent on current launch context (devnet, testnet, mainnet)

KonradStaniec (approved):
Codewise LGTM!

I would say next step would be to describe in `README.md` how those strings containing data should be updated i.e:
- we are launching `devnet` how to prepare those strings and configure `app.go` so the upgrade will happen.

app/upgrades/signetlaunch/data_signed_fps_test.go
RafilxTenfen: good idea, I will update

RafilxTenfen: done at https://github.com/babylonlabs-io/babylon/pull/28/commits/c503f3bf3ce59c673203f5b76b512c01f29397a4

app/upgrades/signetlaunch/README.md
```diff
@@ -2,3 +2,44 @@
 
 This folder contains a software upgrade for testing purposes.
 DO NOT USE IN PRODUCTION!
+
```
RafilxTenfen, line 5: There is one example of shell script running this upgrade with load of BTC Headers at https://github.com/babylonlabs-io/babylon-deployment/blob/main/deployments/local/bbn-start-and-upgrade-signet-launch.sh

Did not linked in the README.md because the babylon-deployment repo is private

app/upgrades/signetlaunch/README.md
```diff
@@ -2,3 +2,58 @@
...
+$SID_BIN btc-headers 1 1000 --output $EXPORT_TO
+btcHeadersJson=$(cat $EXPORT_TO)
+
+# writes the headers to babylon as go file
+echo "package signetlaunch
+
+const NewBtcHeadersStr = \`$btcHeadersJson\`" > $GO_BTC_HEADERS_PATH
+```
+
+### Signed Create Finality Provider
```
RafilxTenfen, line 44: https://babylonlabs.atlassian.net/wiki/x/ZoANAQ
Doc that FPs will need to follow, to create the signed MsgCreateFinalityProvider

app/upgrades/signetlaunch/README.md
```diff
@@ -2,3 +2,58 @@
 
 This folder contains a software upgrade for testing purposes.
 DO NOT USE IN PRODUCTION!
+
+## Compile signet launch upgrade
+
+This upgrade has two data points loaded from strings. BTC Headers from
```
SebastianElvis, line 8: "data point" does not seem to mean stuff in our scenario (btc headers, fps, ...)?

SebastianElvis (approved):

PR #33, opened 2024-08-27
rafactor/Refactor handler for staker message

KonradStaniec:
Refactor our staking handler.

Pre-requisite before implementing: https://github.com/babylonlabs-io/pm/blob/main/adr/adr-023-update-btcstaking-params.md

### DIFF
--- a/x/btcstaking/keeper/msg_server.go
+++ b/x/btcstaking/keeper/msg_server.go
@@ -343,7 +343,14 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 		return nil, types.ErrInvalidProofOfPossession.Wrapf("error while validating proof of posession: %v", err)
 	}
 
-	// 3. Check finality providers to which message delegate
+	// 3. Check if it is not duplicated staking tx
+	stakingTxHash := parsedMsg.StakingTx.Transaction.TxHash()
+	delgation := ms.getBTCDelegation(ctx, stakingTxHash)
+	if delgation != nil {
+		return nil, types.ErrReusedStakingTx.Wrapf("duplicated tx hash: %s", stakingTxHash.String())
+	}
+
+	// 4. Check finality providers to which message delegate
 	// Ensure all finality providers are known to Babylon, are not slashed,
 	// and their registered epochs are finalised
 	for _, fpBTCPK := range parsedMsg.FinalityProviderKeys.PublicKeysBbnFormat {
@@ -358,7 +365,7 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 		}
 	}
 
-	// 3. Validate parsed message against parameters
+	// 5. Validate parsed message against parameters
 	vp := ms.GetParamsWithVersion(ctx)
 
 	btccParams := ms.btccKeeper.GetParams(ctx)
@@ -369,7 +376,7 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 		return nil, err
 	}
 
-	// 4. Check:
+	// 6. Check:
 	// - timelock of staking tx
 	// - staking tx is k-deep
 	// - staking tx inclusion proof
@@ -408,7 +415,7 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 		return nil, types.ErrInvalidStakingTx.Wrapf("staking tx's timelock has no more than w(=%d) blocks left", btccParams.CheckpointFinalizationTimeout)
 	}
 
-	// 6.all good, construct BTCDelegation and insert BTC delegation
+	// 7.all good, construct BTCDelegation and insert BTC delegation
 	// NOTE: the BTC delegation does not have voting power yet. It will
 	// have voting power only when it receives a covenant signatures
 	newBTCDel := &types.BTCDelegation{
--- a/x/btcstaking/keeper/msg_server.go
+++ b/x/btcstaking/keeper/msg_server.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	btcckpttypes "github.com/babylonlabs-io/babylon/x/btccheckpoint/types"
+
 	errorsmod "cosmossdk.io/errors"
 	sdkmath "cosmossdk.io/math"
 	"github.com/babylonlabs-io/babylon/btcstaking"
@@ -13,6 +15,7 @@ import (
 	"github.com/babylonlabs-io/babylon/x/btcstaking/types"
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/cosmos/cosmos-sdk/telemetry"
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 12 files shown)


x/btcstaking/keeper/msg_server.go
```diff
@@ -437,19 +313,134 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
+	parsedMsg, err := types.ParseCreateDelegationMessage(req)
+
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%v", err)
+	}
+
+	// 2. Basic stateless checks
+	// - verify proof of possession
+	if err := parsedMsg.ParsedPop.Verify(parsedMsg.StakerAddress, parsedMsg.StakerPK.BbnPk, ms.btcNet); err != nil {
+		return nil, types.ErrInvalidProofOfPossession.Wrapf("error while validating proof of posession: %v", err)
```
gitferry, line 343:
```suggestion
		return nil, types.ErrInvalidProofOfPossession.Wrapf(err)
```
The message is already in the error text

x/btcstaking/keeper/msg_server.go
```diff
@@ -437,19 +313,134 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
+	// 3. Check if it is not duplicated staking tx
+	stakingTxHash := parsedMsg.StakingTx.Transaction.TxHash()
+	delgation := ms.getBTCDelegation(ctx, stakingTxHash)
+	if delgation != nil {
+		return nil, types.ErrReusedStakingTx.Wrapf("duplicated tx hash: %s", stakingTxHash.String())
+	}
+
+	// 4. Check finality providers to which message delegate
+	// Ensure all finality providers are known to Babylon, are not slashed,
+	// and their registered epochs are finalised
```
gitferry, line 355: we should cut the last line as we don't have the registered epoch anymore

x/btcstaking/keeper/msg_server.go
```diff
@@ -437,19 +313,134 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
+	for _, fpBTCPK := range parsedMsg.FinalityProviderKeys.PublicKeysBbnFormat {
+		// get this finality provider
+		fp, err := ms.GetFinalityProvider(ctx, fpBTCPK)
+		if err != nil {
+			return nil, err
+		}
+		// ensure the finality provider is not slashed
+		if fp.IsSlashed() {
+			return nil, types.ErrFpAlreadySlashed
```
gitferry, line 364: how about wrapping the slashed fp pk?

x/btcstaking/keeper/msg_server.go
```diff
@@ -437,19 +313,134 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
+	// - staking tx is k-deep
+	// - staking tx inclusion proof
+	stakingTxHeader := ms.btclcKeeper.GetHeaderByHash(ctx, parsedMsg.StakingTxProofOfInclusion.HeaderHash)
+
+	if stakingTxHeader == nil {
+		return nil, fmt.Errorf("header that includes the staking tx is not found")
+	}
+
+	// no need to do more validations to the btc header as it was already
+	// validate by the btclight client module
```
gitferry, line 390:
```suggestion
	// validated by the btclightclient module
```

x/btcstaking/keeper/msg_server.go
```diff
@@ -437,19 +313,134 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
+	if !proofValid {
+		return nil, types.ErrInvalidStakingTx.Wrapf("not included in the Bitcoin chain")
+	}
+
+	startHeight := stakingTxHeader.Height
+
+	endHeight := stakingTxHeader.Height + uint64(parsedMsg.StakingTime)
+
+	btcTip := ms.btclcKeeper.GetTipInfo(ctx)
+	stakingTxDepth := btcTip.Height - stakingTxHeader.Height
```
gitferry, line 409:
```suggestion
	stakingTxDepth := btcTip.Height - stakingTxHeader.Height + 1
```

Should we follow the convention that inclusion counts as 1 confirmation?

x/btcstaking/keeper/msg_server.go
```diff
@@ -160,275 +163,148 @@ func caluculateMinimumUnbondingValue(
...
-	kValue, wValue := btccParams.BtcConfirmationDepth, btccParams.CheckpointFinalizationTimeout
-
-	minUnbondingTime := types.MinimumUnbondingTime(vp.Params, btccParams)
+type ParamsValidationResult struct {
+	StakingOutputIdx   uint32
+	UnbondingOutputIdx uint32
+}
 
+// ValidateParams validates parsed message against parameters
+func ValidateParams(
```
gitferry, line 172: What do you think we move the params-related code to a separate file or a different package?

And would it be better if we make `ValidateParams` a method of `ParsedCreateDelegationMessage`?

x/btcstaking/types/create_delegation_parser.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func NewParsedPublicKey(key *bbn.BIP340PubKey) (*ParsedPublicKey, error) {
+	if key == nil {
+		return nil, fmt.Errorf("cannont parse nil *bbn.BIP340PubKey")
+	}
+	pk, err := key.ToBTCPK()
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to parer *bbn.BIP340PubKey: %w", err)
```
gitferry, line 29:
```suggestion
		return nil, fmt.Errorf("failed to parse *bbn.BIP340PubKey: %w", err)
```

x/btcstaking/types/create_delegation_parser.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/wire"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+type ParsedPublicKey struct {
+	PublicKey *btcec.PublicKey
+	BbnPk     *bbn.BIP340PubKey
```
gitferry, line 19:
```suggestion
	*btcec.PublicKey
	*bbn.BIP340PubKey
```
A bit confusing about `BbnPk` because we also call it a BTC public key

x/btcstaking/types/create_delegation_parser.go
```diff
@@ -0,0 +1,268 @@
...
+
+	return &ParsedPublicKey{
+		PublicKey: pk,
+		BbnPk:     key,
+	}, nil
+}
+
+type ParsedBIP340Signature struct {
+	Signature *schnorr.Signature
+	BbnSig    *bbn.BIP340Signature
```
gitferry, line 40:
```suggestion
	*schnorr.Signature
	*bbn.BIP340Signature
```
Same here. `BbnSig` is a bit confusing

gitferry (approved):
Great work! Mostly nits

x/btcstaking/keeper/msg_server.go
KonradStaniec: I think on Babylon node we are using depth everywhere i.e for being k-deep, and for w-deep. Of course we may discuss this change as it makes sense to be compatible with phase-1, but this is not something to be done in refactor pr.

For now I have created ticket to track this - https://github.com/babylonlabs-io/babylon/issues/39

x/btcstaking/keeper/msg_server.go
KonradStaniec: hmm I guess it can be moved to types as it really do not touch/modify the state.

As for making it a method on `ParsedCreateDelegationMessage`, I am slightly against mostly due to the fact validating against the parameters is not really a property of `ParsedCreateDelegationMessage` (I know this sounds weird 😅 ) but rather operation that is by using `ParsedCreateDelegationMessage` with other pieces of data.  

Maybe in other words, `ParsedCreateDelegationMessage` is just data transfer object without any assumptions how it will be used. Other functions, object by use it for its own purposes, but the dto object should be as bare as possible.

PR #37, opened 2024-08-27
feat(ADR-024): Enable BTC-timestamping public randomness

### LINKED ISSUE #8: BTC-timestamping public randomness

gitferry:
This PR collects sub-PRs to close #8 and complete the whole feature defined in [ADR-024](https://github.com/babylonlabs-io/pm/blob/main/adr/adr-024-timestamping-public-randomness.md)

TODOs
- [x] #12 
- [x] #23 
- [x] [changes in the finality provider repo](https://github.com/babylonlabs-io/finality-provider/pull/40)
- [x] #47 
- [x] [table-driven tests on VotingPowerDistCache](https://github.com/babylonlabs-io/babylon/pull/38)
- [ ] validate upgrade with db schema change in local deployment


gitferry:
in favor of #51

PR #40, opened 2024-08-30
POC: feature gate for zone conceirge module

SebastianElvis:
This PR provides a PoC for the feature gate of zone concierge. In particular,

- It introduces a new module parameter `enable_integration`. If set true then integration is enabled.
- Disabling the integration means that 1) post handler will do nothing, 2) IBC channel creation with zone concierge is disabled, and 3) inclusion proofs won't be generated. Here 1 disables phase-1 timestamping integration, and 2/3 disables phase-2 timestamping integration.

Excluding the proto/swagger stuff, this is a change of 30 LoCs. Some modifications might not be necessary but adding them for being extra cautious.

This feature gate is already reflected in failed unit/timestamping e2e tests -- by default zone concierge no longer indexes IBC headers, or approves any IBC channel creation request.


KonradStaniec:
so there are few specific things that worry me about this approach:
1. By putting new field in params we need to do the upgrade from `TGE` chain to now contain his modified version of params.
2. By putting the enable/disable flag to parameters we signal that integration can be enabled/disabled at will. So in theory it could be enabled in phase-2 if validators wanted it, even though this would be un-tested code path which we do not intend to support.
3. (This may be my hazy understading of timestamping integration, so correct me if I am wrong 😅 ) If we disable proofs generation, wouldn't be need to backfill those in upgrade eitherway ?


What your opinion on having `EnableIntegration` flag hardcoded in binary at app.go level ?  Then:
- we do not need upgrade
- there is not way to enable it without releasing new version
- for ibc messages we can disable routing at app.go level which seems like safer option i.e instead of:
```
	ibcRouter := porttypes.NewRouter().
		AddRoute(ibctransfertypes.ModuleName, transferStack).
		AddRoute(zctypes.ModuleName, zoneConciergeStack).
		AddRoute(wasmtypes.ModuleName, wasmStack)
```
have:
```
	ibcRouter := porttypes.NewRouter().
		AddRoute(ibctransfertypes.ModuleName, transferStack).
		AddRoute(wasmtypes.ModuleName, wasmStack)


	if !EnableIntegration {
		ibcRouter = ibcRouter.AddRoute(zctypes.ModuleName, zoneConciergeStack)
	}
```

SebastianElvis:
Good point. Since genesis params are not expected to be changed, maybe it's not a good idea to touch it via software upgrade. 

Compared to having one more storage for it, maybe having a public bool variable in `app.go` is more lightweight and cleaner? This achieves the same thing, does not require us to wire this storage in upgrade handlers, and allows us to enable testing integration stuff using tricks like https://github.com/babylonlabs-io/babylon/blob/7315714451e741fb9b92f92e0f5446f671400c70/app/e2e_include_upgrades.go. Wdyt?

KonradStaniec:
>Compared to having one more storage for it, maybe having a public bool variable in app.go is more lightweight and cleaner? This achieves the same thing, does not require us to wire this storage in upgrade handlers, and allows us to enable testing integration stuff using tricks like https://github.com/babylonlabs-io/babylon/blob/7315714451e741fb9b92f92e0f5446f671400c70/app/e2e_include_upgrades.go. Wdyt?

yep hardcoding constant `EnableIntegration` is probably easiest and cleanest way to feature gate this whole module.

SebastianElvis:
OK, modified this PoC to define a global boolean variable `var EnableIntegration = false` as feature gate, and gated the inclusion of zone concierge in IBC router. Seems to be much cleaner now

x/zoneconcierge/types/params.go
```diff
@@ -4,6 +4,10 @@ import (
 	"fmt"
 )
 
+var (
+	EnableIntegration = false
```
RafilxTenfen, line 8: this is not a const, because it can be modified by tests?

SebastianElvis: Yes this is intentional. I wanted to enable tests when the build tag `e2e` is enabled. This way we can keep all timestamping e2e tests running while disabling integration in real deployments

SebastianElvis:
I have updated this PR according to the comments, including:

- feature-gating the GRPC queries and messages in Zone Concierge module
- enabling integration for test environments so that all existing tests still work
- adding a new fuzz test for ensuring the correct behaviours of {IBC, PostHandler, queries, message servers} when the feature gate is true/false

Please feel free to take a look again. Let me know if you want to see an end-to-end test.

PR #43, opened 2024-08-30
chore: e2e init chain upgrade

RafilxTenfen:
- Did not modified params
- fix e2e-init-chain to specific version `544386204a860ee7159923b3d34174d0cf8c6049` current main before proto changes to check https://github.com/babylonlabs-io/babylon/pull/41

#42 contains also the changes from this PR + proto changes for retro compatibility


RafilxTenfen:
close in favor of https://github.com/babylonlabs-io/babylon/pull/54

PR #46, opened 2024-09-02
feat: add e2e-init-chain

RafilxTenfen:
Add e2e-init-chain, it is necessary for e2e testing in new releases to have the init chain with the same proto versions used at the v09 bin


contrib/images/e2e-initialization/init.Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM golang:1.21 as build-env
+
+ARG E2E_SCRIPT_NAME
+
+# Copy All
+WORKDIR /go/src/github.com/babylonlabs-io/babylon
+COPY ./ /go/src/github.com/babylonlabs-io/babylon/
```
KonradStaniec, line 7: doesn't it need to be:

```
# Version to build. Default is empty
ARG VERSION

# Copy All
WORKDIR /go/src/github.com/babylonlabs-io/babylon
COPY ./ /go/src/github.com/babylonlabs-io/babylon/

# Handle if version is set
RUN if [ -n "${VERSION}" ]; then \
        git fetch origin ${VERSION}; \
        git checkout -f ${VERSION}; \
    fi

```

i.e we should be able to configure binary version here ?

contrib/images/Makefile
```diff
@@ -1,20 +1,31 @@
...
+		--build-arg BUILD_TAGS="e2e"
 
 babylond-rmi:
 	docker rmi babylonlabs-io/babylond 2>/dev/null; true
 
+e2e-init-chain-rmi:
+	docker rmi babylonlabs-io/babylond-e2e-init-chain --force 2>/dev/null; true
+
+e2e-init-chain:
+	docker build -t babylonlabs-io/babylond-e2e-init-chain --build-arg E2E_SCRIPT_NAME=chain --platform=linux/x86_64 \
```
KonradStaniec, line 20: I wonder whether ` --platform` shouldn't be configurable by some env variable, as when I run test locally on arm mac those fail due to native binding to bls libraries and changing platform fixes that

test/e2e/configurer/chain/queries.go
```diff
@@ -413,3 +415,54 @@ func (n *NodeConfig) QueryWasmSmartObject(contract string, msg string) (resultOb
 	}
 	return resultObject, nil
 }
+
+func (n *NodeConfig) QueryProposal(proposalNumber int) govtypesv1.QueryProposalResponse {
```
KonradStaniec, line 419: I wonder whether all of those commands are necessary, from what I can tell they are not used in any way ?

contrib/images/e2e-initialization/init.Dockerfile
RafilxTenfen: since we want this fixed version for testing, we don't need to have the version argument

contrib/images/Makefile
RafilxTenfen: I did not know that, I will add --platform tag to docker builds

test/e2e/configurer/chain/queries.go
RafilxTenfen: nop, they are mostly not needed 
just changes were applied when cherry picked, I will remove unused funcs

RafilxTenfen: done at https://github.com/babylonlabs-io/babylon/pull/46/commits/4bf9af900da69442dcb80d365c8e43b064ce0b3a

contrib/images/Makefile
RafilxTenfen: Can you check if this https://github.com/babylonlabs-io/babylon/pull/46/commits/997de7cfcee2063e3b92b011da73c8d467da8b3a
solved?

contrib/images/e2e-initialization/init.Dockerfile
KonradStaniec: hmm correct me if I am wrong, but I have had imagined that this is needed similiary to `babylond-before-upgrade` step.

i.e for babylon before upgrade we call :
```
	docker build --tag babylonlabs-io/babylond-before-upgrade -f babylond/Dockerfile \
    --build-arg VERSION="${BABYLON_VERSION_BEFORE_UPGRADE}" ${BABYLON_FULL_PATH}
```
to actually build the image with some old version.

I thought this will happen here too i.e
1. First init chain with the `VERSION="${BABYLON_VERSION_BEFORE_UPGRADE}"`
2. then build image before upgrade also with `VERSION="${BABYLON_VERSION_BEFORE_UPGRADE}"` 
what am I missing ?

contrib/images/Makefile
KonradStaniec: hmm I have checked it locally and got:
```
> make build-docker
/Library/Developer/CommandLineTools/usr/bin/make -C contrib/images babylond
docker rmi babylonlabs-io/babylond 2>/dev/null; true
Untagged: babylonlabs-io/babylond:latest
Deleted: sha256:7cd3d4a92eff50f5490ea0636532b3fbbcea3f5585d98fb886b1e96423614339
docker build --tag babylonlabs-io/babylond -f babylond/Dockerfile /Users/konradstaniec/Work/babylon/babylon --platform=local
[+] Building 2.0s (3/3) FINISHED                                                                                                                                    docker:desktop-linux
 => [internal] load build definition from Dockerfile                                                                                                                                0.0s
 => => transferring dockerfile: 2.22kB                                                                                                                                              0.0s
 => ERROR [internal] load metadata for docker.io/library/debian:bookworm-slim                                                                                                       2.0s
 => ERROR [internal] load metadata for docker.io/library/golang:1.21                                                                                                                2.0s
------
 > [internal] load metadata for docker.io/library/debian:bookworm-slim:
------
------
 > [internal] load metadata for docker.io/library/golang:1.21:
------
Dockerfile:1
--------------------
   1 | >>> FROM golang:1.21 AS build-env
   2 |
   3 |     # Customize to your build env
--------------------
ERROR: failed to solve: golang:1.21: no match for platform in manifest: not found
make[1]: *** [babylond] Error 1
make: *** [build-docker] Error 2
```

so it seems `--platform=local` does not really work on m1 macs 😅

KonradStaniec: thinking out loud, why do we even need this `--platform=linux/x86_64` in this particular target ? Other docker building commands do not specify this

contrib/images/e2e-initialization/init.Dockerfile
RafilxTenfen: We will have `VERSION` argument on main branch, that is to go to another version, here it will not be used to move to another version because this is already the fixed version we need to test 

But if you feel more comfortable I could add it

KonradStaniec: lets add it for the sake of completeness, the worst case scenario it will be unused

RafilxTenfen: added version argument to dockerfile 
https://github.com/babylonlabs-io/babylon/pull/46/commits/26a3fc93d9cf559e695d76a676c519542dde35c5

At the makefile to build the docker I did not set any argument

PR #47, opened 2024-09-02
doc(ADR-024): update document related to BTC timestamping

gitferry:

PR #51, opened 2024-09-03
feat(ADR-024): Enable BTC-timestamping public randomness

### LINKED ISSUE #8: BTC-timestamping public randomness

gitferry:
This PR collects sub-PRs to close #8 and complete the whole feature defined in [ADR-024](https://github.com/babylonlabs-io/pm/blob/main/adr/adr-024-timestamping-public-randomness.md)

TODOs
- [x] #12 
- [x] #23 
- [x] [changes in the finality provider repo](https://github.com/babylonlabs-io/finality-provider/pull/40)
- [x] #47 
- [x] [table-driven tests on VotingPowerDistCache](https://github.com/babylonlabs-io/babylon/pull/38)
- [x] [validate upgrade with db schema change in local deployment](https://github.com/babylonlabs-io/babylon-deployment/tree/gai/upgrade-pub-rand-commit)


proto/babylon/btcstaking/v1/incentive.proto
```diff
@@ -9,9 +9,15 @@ option go_package = "github.com/babylonlabs-io/babylon/x/btcstaking/types";
 // VotingPowerDistCache is the cache for voting power distribution of finality providers
 // and their BTC delegations at a height
 message VotingPowerDistCache {
+    option (gogoproto.goproto_getters) = false;
+    // total_voting_power is the total voting power of all the finality providers
+    // in the cache
     uint64 total_voting_power = 1;
     // finality_providers is a list of finality providers' voting power information
     repeated FinalityProviderDistInfo finality_providers = 2;
+    // num_active_fps is the number of finality providers that have positive voting power
```
SebastianElvis, line 18:
```suggestion
    // num_active_fps is the number of finality providers that have active BTC delegations
```

a FP has positive voting power at a height only when it has active BTC
delegations at this height AND has timestamped public randomness at this height

test/e2e/configurer/chain/queries.go
```diff
@@ -180,7 +180,9 @@ func (n *NodeConfig) QueryListSnapshots() ([]*cmtabcitypes.Snapshot, error) {
 func (n *NodeConfig) QueryRawCheckpoint(epoch uint64) (*ct.RawCheckpointWithMetaResponse, error) {
 	path := fmt.Sprintf("babylon/checkpointing/v1/raw_checkpoint/%d", epoch)
 	bz, err := n.QueryGRPCGateway(path, url.Values{})
-	require.NoError(n.t, err)
+	if err != nil {
+		return nil, err
+	}
```
SebastianElvis, line 185:
```suggestion
	require.NoError(n.t, err)
```
Failing to query the node should not happen so we can panick here

test/e2e/configurer/chain/queries_btcstaking.go
```diff
@@ -94,15 +94,19 @@ func (n *NodeConfig) QueryUnbondedDelegations() []*bstypes.BTCDelegationResponse
...
+	if err != nil {
+		return 0, err
+	}
 
 	var resp bstypes.QueryActivatedHeightResponse
 	err = util.Cdc.UnmarshalJSON(bz, &resp)
-	require.NoError(n.t, err)
+	if err != nil {
+		return 0, err
+	}
```
SebastianElvis, line 107: We can panick here as this means there is something fishy in the setup

SebastianElvis (approved):
Great work putting everything together!

Only super minor comments but these can be done in subsequent PRs

KonradStaniec (approved):
LGTM!

proto/babylon/btcstaking/v1/incentive.proto
gitferry: Agreed. The two conditions combined lead to fp having voting power

RafilxTenfen (approved):
LGTM

PR #52, opened 2024-09-03
feat: add comprehensive test suite for staking msg validation

KonradStaniec:
Comprehensive table driven test suite to check validations on message validation

Update: upgrade e2e test may still fail, this will be tackled in next pr

### DIFF
--- a/x/btcstaking/types/validate_parsed_message_test.go
+++ b/x/btcstaking/types/validate_parsed_message_test.go
@@ -1,0 +1,773 @@
+package types_test
+
+import (
+	"math/rand"
+	"testing"
+	"time"
+
+	sdkmath "cosmossdk.io/math"
+	"github.com/babylonlabs-io/babylon/btcstaking"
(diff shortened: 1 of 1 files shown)


Lazar955 (approved):
Looks good, nice work!
1. It would be nice to have more comments in the helper functions
2. It would be nice to use `t.Parallel()` if possible in all self contained unit tests (this can be a setting in conf of `golangci-lint`

KonradStaniec:
>it would be nice to have more comments in the helper functions

Agreed! for now I have added docstrings to largest helper function in latest commit: https://github.com/babylonlabs-io/babylon/pull/52/commits/e00381713c9917adb56b75bf3803f9f9b3b0caec

>It would be nice to use t.Parallel() if possible in all self contained unit tests (this can be a setting in conf of golangci-lint

Agreed that ` t.Parallel()` would be good in is some heavier tests, thought here we only run some simple function calls, not sure there is big advantage of parallelising them. (and parallelising them makes the test results pritining a bit more ugly as those results as out order 😅 )

PR #55, opened 2024-09-04
Remove `x/zoneconcierge` module

KonradStaniec:
just a show case of changes necessary to remove zoneconciege module


app/upgrades/signetlaunch/upgrades.go
```diff
@@ -23,10 +23,17 @@ import (
...
+	ZoneConciergeStoreKey = "zoneconcierge"
+)
+
 var Upgrade = upgrades.Upgrade{
 	UpgradeName:          "signet-launch",
 	CreateUpgradeHandler: CreateUpgradeHandler,
-	StoreUpgrades:        store.StoreUpgrades{},
+	// Upgrade necessary for deletions of `zoneconcierge`
+	StoreUpgrades: store.StoreUpgrades{
+		Deleted: []string{ZoneConciergeStoreKey},
```
RafilxTenfen, line 35: Have we ever tested if adding this store back to another upgrade with the same key works? 
Asking this, because I never saw to delete a module to add it back later and not sure of the implications that might have 

Also, the worst case would be when adding it back with a different naming than `zoneconcierge`

KonradStaniec: tbh  later I would re-add it with different key, even `zoneconcierge1` would be fine just to avoid thinking about this case 😅

CHANGELOG.md
```diff
@@ -37,6 +37,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+* [#55](https://github.com/babylonlabs-io/babylon/pull/55) Remove `x/zoneconcierge`
```
SebastianElvis, line 40: This needs to be marked as consensus breaking

test/e2e/btc_timestamping_e2e_test.go
SebastianElvis, line 1: what do you think about removing this e2e test completely? it's not a part of user stories for phase-2

SebastianElvis (approved):
Lgtm. The following things can be removed as well

- `e2e-run-btc-timestamping` in CI (if we want to remove this e2e)
- relevant Makefile entries

test/e2e/btc_timestamping_e2e_test.go
KonradStaniec: tbh I am leaning to leaving it as:
- it test finalization of several epochs by `FinalizeSealedEpochs` in `Test4GenerateAndWithdrawReward` test
- tests that out cosmowasm custom queries work as expected in `Test5Wasm`
- tests rewards for vigilantes in `Test6InterceptFeeCollector` (though this will probably change when tokenomics story lands)

PR #56, opened 2024-09-05
Fix typo update README

Olexandr88:
Fix typo update README


RafilxTenfen:
I think that it is suppose to be `babylond` where it is the Babylon daemon

PR #57, opened 2024-09-05
tests: bcd setup in e2e

gusin13:
This pr setups up babylon<->bcd integration suite

### Background
Currently in the babylon consumer integration tests, babylon nodes are spun up one of which acts as provider, and other node (which is also babylon node) acts as consumer. The problem in this is babylon node doesn't have the custom integration [module](https://github.com/babylonlabs-io/babylon-sdk/tree/main/x/babylon) needed to test proper consumer behaviour (slashing etc) 

We need to have [bcd](https://github.com/babylonlabs-io/babylon-sdk) in the e2e tests for sure. Now for that I had initially tried to use vanilla babylon-sdk docker image and then making changes in the existing setup to incorporate this external chain. The problem is in the existing e2e [infra](https://github.com/babylonlabs-io/babylon/blob/main/test/e2e/configurer/factory.go) the tests work as following 
1. the suite starts
2. chain configurer is setup which defines the number of nodes/validators we need
3. then genesis/config/app yamls are created through go code for the above configured chains
4. finally the docker containers are started and the above config/genesis are injected in the containers

i had very hard time setting up bcd with above steps and ran into all kinds of problems namely encoding config related issues, node not starting, some issue in genesis file. i was doing this on a separate [pr](https://github.com/babylonlabs-io/babylon-private/pull/17) then abandoned that due the above mentioned issues

it was much easier to simply use the scripts in the integration deployment [repo](https://github.com/babylonlabs-io/babylon-integration-deployment/tree/main/deployments/btc-staking-integration-bitcoind)

### PR insights
1. Separate scripts, dockerfile, pre deployment setup, genesis config etc under test/e2e/bcd_integration 
2. New rpc/grpc query clients specially for this setup to query the nodes
3. New test file `bcd_integration_test.go` to host the tests, its almost same as `btc_staking_integration_e2e_test.go` but the queries now use the rpc client for bcd/babylon stuff

tl:dr in test/e2e/bcd_integration dir 
dockerfile, docker compose, pre deployment, setup, wrapper scripts are taken from [deployment](https://github.com/babylonlabs-io/babylon-integration-deployment) repo

clientcontroller, cosmwasm rpc client are taken from finality provider [repo](https://github.com/babylonlabs-io/finality-provider/tree/base/consumer-chain-support/itest) 

bcd_integration_test.go glues together everything and performs actual tests.

### DIFF
--- a/test/e2e/bytecode/README.md
+++ b/test/e2e/bytecode/README.md
@@ -1,1 +1,3 @@
-Contract built from https://github.com/babylonlabs-io/babylon-contract/tree/main/contracts
+# Contract Bytecode
+
+To build the WASM binaries, follow the instructions in this repository: https://github.com/babylonlabs-io/babylon-contract and copy the resulting files into this directory.


SebastianElvis:
Some offline discussion:

- The docker-compose setup can be moved to `contrib/`
- The cosmwasm client can be moved to a new repository as it's generic and not related to Babylon. FP and Babylon e2e can make use of this repository

test/e2e/bcd_integration/Dockerfile
```diff
@@ -0,0 +1,69 @@
+FROM debian:bullseye-slim AS build-env
```
RafilxTenfen, line 1: why not start from a golang versioned base image?

test/e2e/bcd_integration/Dockerfile
RafilxTenfen, line 1: this image will run two things: relayer and babylon-sdk?

test/e2e/bcd_integration/Makefile
RafilxTenfen, line 1: suggestion to move this to `contrib/images`

test/e2e/bcd_integration/cosmwasmclient/query/tendermint.go
RafilxTenfen, line 1: can we move this to test/e2e/configurer/chain
where there is already other queries

contrib/images/bcd/pre-deployment.sh
```diff
@@ -2,7 +2,7 @@
 
 # Create new directory that will hold node and services' configuration
 mkdir -p .testnets && chmod -R 777 .testnets
-docker run --rm -v $(pwd)/.testnets:/data babylonlabs-io/babylond \
+docker run --rm -v $(pwd)/.testnets:/data --user $(id -u):$(id -g) babylonlabs-io/babylond \
```
RafilxTenfen, line 5: this `--user $(id -u):$(id -g)` fixes the removal of wasm data?

Makefile
```diff
@@ -454,6 +457,8 @@ build-docker-e2e:
 	$(MAKE) -C contrib/images babylond-e2e
 	$(MAKE) -C contrib/images babylond-before-upgrade
 	$(MAKE) -C contrib/images e2e-init-chain
+	$(MAKE) -C test/e2e/bcd_integration build-integration-test
```
tgntr, line 460: these are now in contrib/images https://github.com/babylonlabs-io/babylon/pull/57/files#diff-b7a8cc1d7346184c68e119aba8c13a202872c6deccf916e6e2b5216b4a6778ceR49-R55

contrib/images/bcd/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
...
+      - "1318:1317"
+      - "9091:9090"
+      - "2346:2345"
+    volumes:
+      - .testnets/node1/babylond:/babylondhome:Z
+    networks:
+      localnet:
+        ipv4_address: 192.168.10.3
+
+  ibcsim-bcd:
```
SebastianElvis, line 42: What do you think of building a `bcd` Docker image and using a container in this
docker compose file? This way we can reuse the IBC relayer functionalities in
the e2e test suite. Probably not for this PR

test/e2e/bcd_consumer_integration/clientcontroller/babylon/babylon.go
SebastianElvis, line 1: Is there a way to reuse commands in `test/e2e/initializaion/chain` so that we
don't need to port the entire client controller here? This could be a bit
challenging though as one needs to create a container manager for the Babylon
node docker container, but we can think about this later

Makefile
```diff
@@ -454,6 +457,8 @@ build-docker-e2e:
 	$(MAKE) -C contrib/images babylond-e2e
 	$(MAKE) -C contrib/images babylond-before-upgrade
 	$(MAKE) -C contrib/images e2e-init-chain
+	$(MAKE) -C contrib/images build-bcd-consumer-integration
+	$(MAKE) -C contrib/images start-bcd-consumer-integration
```
SebastianElvis, line 461: Do we need to run `start-bcd-consumer-integration` for this makefile entry?

SebastianElvis (commented):
Great work!

Btw, after we finish the cascading slashing e2e test, the `BCDConsumerIntegrationTestSuite` will implement all user stories in `BTCStakingIntegrationTestSuite`. At that stage, what do you think about removing `BTCStakingIntegrationTestSuite`?

contrib/images/bcd/docker-compose.yml
SebastianElvis: The code tagged here includes two functions that use chain configs: `runHermesIBCRelayer` and `createBabylonPhase2Channel`. If we look at their implementation, it only uses channel ID, port ID, and some IBC configurations inside chain config. If we hardcode these values is this possible?

Makefile
SebastianElvis: I see, then maybe let's keep them separate

KonradStaniec (commented):
My main worry with this pr, is that having part of e2e tests executed by scripts and part by the framework we have now (with dockertest and dockerfiles) will be pretty terrible to maintain.

For now it is fine, but in the future when phase-3 should hit the main net and be part of babylon node `main` branch we should simplify this somehow.

SebastianElvis:
> My main worry with this pr, is that having part of e2e tests executed by scripts and part by the framework we have now (with dockertest and dockerfiles) will be pretty terrible to maintain.
> 
> For now it is fine, but in the future when phase-3 should hit the main net and be part of babylon node `main` branch we should simplify this somehow.

The main reasons of doing this are:

- To test cascased slashing one needs to have 2 different consumer chains. As one is `bcd` chain and the other is Babylon chain (probably will be replaced by `bcd` chain in the future), this is made very hard as Cosmos SDK provides little support to manage multiple Cosmos SDK chains in one program.
- Later on if we want to cover even more complicated cases of cascaded slashing, the Docker based setup could be much easier compared to using the current framework

My recommendation is that 

- For testing BTC staking integration, we start using Docker based setup
- Since we don't have BTC timestamping integration, later on we will merge the BTC timestamping integration e2e to BTC staking integration e2e
- BTC timestamping/staking e2e tests can follow the existing framework as they only involve one Babylon chain

Wdyt?

KonradStaniec:
> My recommendation is that
> 
> For testing BTC staking integration, we start using Docker based setup
> Since we don't have BTC timestamping integration, later on we will merge the BTC timestamping integration e2e to BTC staking integration e2e, and remove the existing Go framework based BTC timestamping/staking integration e2e tests.
> BTC timestamping/staking e2e tests can follow the existing framework as they only involve one Babylon chain
> Wdyt?

Sounds good on the high level and we can progress with this plan, although we are still left with two e2e approaches to testing, which imo will generate maintenance issue in the future. 

In my (personal) perfect world, we would use existing go framework for everything and extend it when needed. I know it may be a bit harder, but for me not dealing with bash scripts, docker compose etc. and controlling everything from go is big advantage.

contrib/images/bcd/Dockerfile
maurolacy, line 1: Nit: Perhaps the folder name should be `ibcsim-bcd` for clarity, and consistency with other repos.

test/e2e/bytecode/babylon_contract.wasm
maurolacy, line 1: Nit: Update the `version.txt` file in this folder with the commit id / tag info for reference.

maurolacy (commented):
Great work pulling all this together! Just some initial comments, will review the Go code next.

test/e2e/bcd_consumer_integration/clientcontroller/babylon/babylon.go
```diff
@@ -0,0 +1,538 @@
...
+		return nil, err
+	}
+
+	return res, nil
+}
+
+// TODO: only used in test. this should not be put here. it causes confusion that this is a method
+// that will be used when FP runs. in that's the case, it implies it should work all all consumer
+// types. but `bbnClient.QueryClient.FinalityProviders` doesn't work for consumer chains
+func (bc *BabylonController) QueryFinalityProviders() ([]*btcstakingtypes.FinalityProviderResponse, error) {
```
maurolacy, line 281: This looks like a duplicated query. Why not use the one in `configurer/chain/queries_btcstaking.go` by example? The same potentially with the other queries below.

test/e2e/bcd_consumer_integration/clientcontroller/config/cosmwasm.go
```diff
@@ -0,0 +1,90 @@
...
+		Debug:                     true,
+		Timeout:                   20 * time.Second,
+		BlockTimeout:              1 * time.Minute,
+		OutputFormat:              "direct",
+		SignModeStr:               "",
+		BtcStakingContractAddress: "",
+	}
+}
+
+func (cfg *CosmwasmConfig) ToQueryClientConfig() *config.CosmwasmConfig {
```
maurolacy, line 73: Why these two config formats?

maurolacy (approved):

test/e2e/bcd_consumer_integration/clientcontroller/cosmwasm/msg.go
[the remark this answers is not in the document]
maurolacy: The contract binaries could also be taken from `babylon-sdk`. So that by defining a version / tag of the SDK, you're left with a compatible version of the contracts, their messages, and their API.

.github/workflows/ci.yml
RafilxTenfen, line 1: :100:

contrib/images/bcd/wrapper.sh
```diff
@@ -0,0 +1,105 @@
...
+    bcd:
+        src:
+            chain-id: $BABYLON_CHAIN_ID
+        dst:
+            chain-id: $CONSUMER_CHAIN_ID
+EOT
+
+echo "Inserting the consumer key"
+CONSUMER_MEMO=$(cat $CONSUMER_CONF/$CONSUMER_CHAIN_ID/key_seed.json | jq .mnemonic | tr -d '"')
+rly --home $RELAYER_CONF_DIR keys restore bcd $CONSUMER_KEY "$CONSUMER_MEMO"
```
RafilxTenfen, line 89: I still think we should separate rly and bcd in diff docker images

test/e2e/bytecode/btc_staking.wasm
RafilxTenfen, line 1: can we add a readme with links to the code that generated this wasm files

contrib/images/ibcsim-bcd/docker-compose.yml
SebastianElvis, line 1: Let's create an issue on avoiding the use of Docker compose in the current setup?

SebastianElvis (approved):
great work figuring out the whole framework!

test/e2e/bytecode/version.txt
SebastianElvis, line 1: Btw, also what do you think about moving `bytecode/` and the Go structs for
contract <-> FP interaction to Babylon SDK repo, as suggested by @maurolacy? If
we are okay with that let's create an issue

maurolacy: Just a comment: I'm not sure if it's better to just move the references to the SDK, or if the SDK can directly define and provide a `Dockerfile` for building (or taking as a basis for building) a dockerised `bcd`. Let's investigate this as part of the issue.

SebastianElvis: Yes this is partially covered by https://github.com/babylonlabs-io/babylon/issues/62

PR #65, opened 2024-09-10
feat(ADR-25): Remove voting power from jailed finality providers

### LINKED ISSUE #64: Jail sluggish finality providers
[ref](https://github.com/babylonlabs-io/babylon/issues/8)

In the current implementation, we have identified sluggish finality providers. To complete the jailing mechanism, we need to remove the voting power of those finality providers to improve finality liveness. In particular, this needs the following changes:

**Babylon side**
- [ ] Jail sluggish finality providers by not assigning voting power for future blocks
- [ ] Allow unjailing of jailed finality providers after a certain duration

**Finality provider side**
- [ ] Stop the daemon if the finality provider is jailed
- [ ] CLI to support unjailing

gitferry:
This PR partially closes #64 by jailing sluggish finality providers by not assigning voting power to them in the voting power table. In particular, this PR made the following changes:
1. Change the `sluggish` label of a finality provider to `jailed`.
2. Implement `AfterSluggishFinalityProviderDetected` hook in the `x/btcstaking` module to set the label of a finality provider to `jailed`. 
3. Introduce `EventJailedFinalityProvider` event as `EventPowerDistUpdate`, which is emitted after a finality provider is jailed.
4. Introduce a new parameter `jail_duration` to `x/finality` (needs to handle upgrade in a separate PR).
5. Introduce a new field `jailed_until` to `FinalityProviderSigningInfo` in `x/finality`, which is calculated as the timestamp of the jailing block plus the `jail_duration` of the parameter.
6. Clear the signing info record of the jailed finality provider and set `jailed_until=jailingBlock.Time.Add(params.JailDuration)`.
7. Within the `BeginBlocker` of the next block, upon `EventJailedFinalityProvider`, record the finality provider public key and do not assign voting power to it.
8. Modify relevant tests


x/epoching/keeper/msg_server.go
```diff
@@ -242,7 +243,7 @@ func (ms msgServer) WrappedCancelUnbondingDelegation(goCtx context.Context, msg
 		)
 	}
 
-	blockHeight := uint64(ctx.BlockHeader().Height)
+	blockHeight := uint64(ctx.HeaderInfo().Height)
```
maurolacy, line 246: What is the difference here? Just curious.

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -150,15 +150,14 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 	ctx context.Context,
 	dc *types.VotingPowerDistCache,
 	events []*types.EventPowerDistUpdate,
-	maxActiveFps uint32,
```
maurolacy, line 153: Is this max being enforced somewhere else? Just curious.

x/finality/keeper/liveness.go
```diff
@@ -200,3 +182,21 @@ func (k Keeper) updateSigningInfo(
...
+	}
+
+	err = sdk.UnwrapSDKContext(ctx).EventManager().EmitTypedEvent(
+		finalitytypes.NewEventJailedFinalityProvider(fpBtcPk),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to emit sluggish finality provider detected event: %w", err)
+	}
+
+	finalitytypes.IncrementSluggishFinalityProviderCounter()
```
maurolacy, line 199: Checked and couldn't find the call to the corresponding 'Decrement` call. Perhaps just remove this counter?

x/finality/types/params.go
```diff
@@ -13,6 +14,7 @@ const (
 	DefaultSignedBlocksWindow = int64(100)
 	DefaultMinPubRand         = 100
 	DefaultFinalitySigTimeout = 3
+	DefaultJailDuration       = 60 * 10 * time.Second // 10 min
```
maurolacy, line 17: This looks like too little to me. 25 hours or so sounds reasonable. So that jailing:

- Is a punishment in itself for being offline.
- Gives some time to the operator to properly restore service / recover liveness.

maurolacy (approved):
LGTM. Just some questions / suggestions for default jailing duration.

Agree with removing the distinction between sluggish and jailed, as they are basically the same state (i.e. sluggish => jailed).

I've seen that in some impls jailed and slashed are also, not exactly the same state, but implied (slashed => jailed). This can be done as well in a follow-up, as it might simplify the code paths, checks, etc. The only requirement would be, slashed => jailed with jailing duration = forever, i.e. you cannot unjail after slashing.

x/epoching/keeper/msg_server.go
gitferry: `ctx.BlockHeader()` returns `ctx.header` which is deprecated. We should use `HeaderInfo` instead. See https://github.com/cosmos/cosmos-sdk/blob/698627016d64dbf33bea01837364c5a6e55a1ee3/types/context.go#L45

x/btcstaking/keeper/power_dist_change.go
gitferry: Yep, here https://github.com/babylonlabs-io/babylon/blob/0a90a5e404fecdd0650bc0042ffff6735e6f515b/x/btcstaking/keeper/power_dist_change.go#L75.

x/finality/types/metrics.go
```diff
@@ -63,10 +63,10 @@ func IncrementSluggishFinalityProviderCounter() {
 	)
 }
 
-// DecrementSluggishFinalityProviderCounter increments the counter for the sluggish
+// DecrementJailedFinalityProviderCounter increments the counter for the jailed
```
maurolacy, line 66:
```suggestion
// DecrementJailedFinalityProviderCounter decrements the counter for the jailed
```

x/finality/types/params.go
gitferry: This was copied from Cosmos SDK but yeah I agree that 1 day seems more reasonable. Fixed in [2caa4af](https://github.com/babylonlabs-io/babylon/pull/65/commits/2caa4af037b38ed3a30efbf8ae82a671ec6ddb75)

proto/babylon/finality/v1/finality.proto
```diff
@@ -64,4 +66,7 @@ message FinalityProviderSigningInfo {
     // missed_blocks_counter defines a counter to avoid unnecessary array reads.
     // Note that `Sum(MissedBlocksBitArray)` always equals `MissedBlocksCounter`.
     int64 missed_blocks_counter = 3;
+    // Timestamp until which the validator is jailed due to liveness downtime.
+    google.protobuf.Timestamp jailed_until = 4
+    [(gogoproto.stdtime) = true, (gogoproto.nullable) = false, (amino.dont_omitempty) = true];
```
RafilxTenfen, line 71: if this cannot be null, what time it will be when the FP is not jailed?

proto/babylon/btcstaking/v1/events.proto
```diff
@@ -40,11 +40,19 @@ message EventPowerDistUpdate {
     bytes pk = 1 [ (gogoproto.customtype) = "github.com/babylonlabs-io/babylon/types.BIP340PubKey" ];
   }
 
+  // EventJailedFinalityProvider defines an event that a finality provider
+  // is jailed
+  message EventJailedFinalityProvider {
```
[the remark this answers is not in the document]
RafilxTenfen: also, we should add the fp cosmos SDK address to this as well

RafilxTenfen (commented):
LGTM, only some suggestions

proto/babylon/btcstaking/v1/events.proto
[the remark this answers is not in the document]
RafilxTenfen: I was thinking for easily finding the event from query from a FP point of view

RafilxTenfen (approved):

proto/babylon/finality/v1/finality.proto
RafilxTenfen: got it, thanks!

Lazar955 (approved):
Nice work 🚀

proto/babylon/btcstaking/v1/events.proto
```diff
@@ -40,11 +40,19 @@ message EventPowerDistUpdate {
...
+    bytes pk = 1 [ (gogoproto.customtype) = "github.com/babylonlabs-io/babylon/types.BIP340PubKey" ];
+  }
+
   // ev is the event that affects voting power distribution
   oneof ev {
     // slashed_fp means a finality provider is slashed
     EventSlashedFinalityProvider slashed_fp = 1;
+    // jailed_fp means a finality provider is jailed
+    EventJailedFinalityProvider jailed_fp = 2;
```
SebastianElvis, line 54: We will have an unjailed event in subsequent PRs right?

x/btcstaking/keeper/hooks.go
```diff
@@ -22,18 +21,5 @@ func (k Keeper) Hooks() Hooks {
 
 // AfterSluggishFinalityProviderDetected updates the status of the given finality provider to `sluggish`
 func (h Hooks) AfterSluggishFinalityProviderDetected(ctx context.Context, fpPk *bbntypes.BIP340PubKey) error {
```
SebastianElvis, line 23: Do we want to change another word other than sluggish?

x/finality/keeper/liveness.go
```diff
@@ -200,3 +182,21 @@ func (k Keeper) updateSigningInfo(
 
 	return modifiedSignInfo, &signInfo, nil
 }
+
+func (k Keeper) jailSluggishFinalityProvider(ctx context.Context, fpBtcPk *types.BIP340PubKey) error {
+	err := k.hooks.AfterSluggishFinalityProviderDetected(ctx, fpBtcPk)
+	if err != nil {
+		return err
+	}
```
SebastianElvis, line 190: How about we call `k.BTCStakingKeeper.JailFinalityProvider` directly? From what I see the `AfterSluggishFinalityProviderDetected` hook is only used here and its only functionality is to call `JailFinalityProvider`. Removing this hook relationship makes the code easier to understand

proto/babylon/finality/v1/finality.proto
SebastianElvis: What do you think about making this nullable and we use a nil check to determine the jailing status of a FP?

x/finality/keeper/liveness.go
```diff
@@ -74,54 +82,28 @@ func (k Keeper) HandleFinalityProviderLiveness(ctx context.Context, fpPk *types.
...
-			return err
+		if err := k.jailSluggishFinalityProvider(ctx, fpPk); err != nil {
+			return fmt.Errorf("failed to jail sluggish finality provider %s: %w", fpPk.MarshalHex(), err)
 		}
 
-		if err := sdkCtx.EventManager().EmitTypedEvent(
-			finalitytypes.NewEventSluggishFinalityProviderDetected(fpPk),
-		); err != nil {
-			panic(fmt.Errorf("failed to emit sluggish finality provider detected event for height %d: %w", height, err))
+		signInfo.JailedUntil = sdkCtx.HeaderInfo().Time.Add(params.JailDuration)
```
SebastianElvis, line 94: Just to confirm, this value will be checked in `MsgUnjail` in subsequent PRs right?

SebastianElvis (commented):
Great work! Mostly questions

proto/babylon/btcstaking/v1/events.proto
gitferry: I don't think so because for unjailing, we just need to revert the `jailed` flag in the finality provider and voting power will be assigned. In [a1b253d](https://github.com/babylonlabs-io/babylon/pull/65/commits/a1b253d42261366bb623674f0b611d4cf49df885), we used the same approach as we did in the timestamping pub rand where `isJailed` is introduced to the fp dist info and if `isJailed` is true, the voting power will be hidden. In another word, `jailing` should not affect delegation accumulation.

x/btcstaking/keeper/hooks.go
gitferry: I'm not sure. The logic here is that `sluggish detected` -> `jail fp` -> `slash rewards`. Do you have something in mind?

x/finality/keeper/liveness.go
gitferry: We would have `slashing reward` logic in the future. So better to insert it here?

proto/babylon/finality/v1/finality.proto
gitferry: Hmmm the jailing status should be obtained from the fp proto other than the signing info but it can also be inferred by checking whether the value is zero

x/finality/keeper/liveness.go
gitferry: Exactly

proto/babylon/btcstaking/v1/events.proto
SebastianElvis: Under which condition will `IsJailed` be reverted? This is slightly different from `IsTimestamped` which won't be reverted once set `true`

x/finality/keeper/liveness.go
SebastianElvis: We can add the logic of slashing rewards into `JailFinalityProvider`, as removing FP voting power and slashing its reward always need to be done at the same time. No strong opinion on this though

proto/babylon/btcstaking/v1/events.proto
gitferry: we update `IsJailed` of each fp dist info of the cache within `recordVotingPowerAndCache` like we did for updating the timestamping status. See https://github.com/babylonlabs-io/babylon/blob/a1b253d42261366bb623674f0b611d4cf49df885/x/btcstaking/keeper/power_dist_change.go#L100

x/finality/keeper/liveness.go
gitferry: Agreed that jailing and slashing rewards should happen at the same time, so we can achieve this by having both modules implement the same hook. Some [discussion](https://github.com/babylonchain/babylon/pull/681#discussion_r1660620692) about why we decided to use a hook.

proto/babylon/btcstaking/v1/events.proto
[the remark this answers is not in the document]
gitferry: this event (also the event for slashing) is not emitted for query but consumed internally for updating the voting table

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -91,11 +91,17 @@ func (k Keeper) recordVotingPowerAndCache(ctx context.Context, prevDc, newDc *ty
...
+		// all the pub rand committed by the fpDistInfo, which might slow down
 		// the process, need optimization
-		fp.IsTimestamped = k.FinalityKeeper.HasTimestampedPubRand(ctx, fp.BtcPk, babylonTipHeight)
+		fpDistInfo.IsTimestamped = k.FinalityKeeper.HasTimestampedPubRand(ctx, fpDistInfo.BtcPk, babylonTipHeight)
+
+		fp, err := k.GetFinalityProvider(ctx, fpDistInfo.BtcPk.MustMarshal())
+		if err != nil {
+			panic(fmt.Errorf("the finality provider %s does not exist: %w", fp.BtcPk.MarshalHex(), err))
+		}
+		fpDistInfo.IsJailed = fp.IsJailed()
```
SebastianElvis, line 104: This will cause some overhead imo -- upon every new block computing the new voting power table will need to retrieve all FPs in the DB, leading to `O(n)` overhead in DB read where `n` is the number of FPs. This was not the case where the DB overhead is `O(# of BTC delegations with state updates)`.

Since we will have `MsgUnjail` anyway, how about we do the state update using the event-based approach similar to the existing events? That is,

- Add a field `IsJailed` to `FinalityProviderDistInfo`
- Upon jailing a FP, push a jailed event to the event queue
- Upon `MsgUnjail` is approved, push an unjailed event to the event queue
- `ProcessAllPowerDistUpdateEvents` processes all the corresponding jailed/unjailed events and sets `IsJailed` for corresponding `FinalityProviderDistInfo` objects

x/finality/keeper/liveness.go
SebastianElvis: Sure. In fact `x/finality` module has access to `x/incentive` so it can invoke the incentive module to slash the reward. But no problem if you think this is better done in a hook

x/btcstaking/keeper/power_dist_change.go
gitferry: Nice point! Chaned to setting `IsJailed` to `FinalityProviderDistInfo` upon event processing and will have `UnjailEvent` to revert the flag in the subsequent PR

x/btcstaking/types/incentive.go
```diff
@@ -56,6 +56,9 @@ func (dc *VotingPowerDistCache) ApplyActiveFinalityProviders(maxActiveFPs uint32
 		if !fp.IsTimestamped {
 			break
 		}
+		if fp.IsJailed {
+			break
+		}
```
SebastianElvis, line 61: Given all these new jailed/timestamped concepts, I started to feel `TotalVotingPower` is no longer an accurate name. How about we rename it to sth like `TotalBTCStake`? Not for this PR, just some thought

SebastianElvis (approved):
Great work!

x/btcstaking/types/incentive.go
gitferry: Good point. Tracked in https://github.com/babylonlabs-io/babylon/issues/72. Will fix it in a separate PR

x/btcstaking/types/btcstaking.go
```diff
@@ -42,18 +42,24 @@ func (fp *FinalityProvider) ValidateBasic() error {
...
 		}
 		if !fps[i].IsTimestamped && fps[j].IsTimestamped {
 			return false
 		}
+		if !fps[i].IsJailed && fps[j].IsJailed {
+			return true
+		}
+		if fps[i].IsJailed && !fps[j].IsJailed {
+			return false
+		}
```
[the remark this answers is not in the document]
gitferry: That would be cleaner. Fixed in [a871d44](https://github.com/babylonlabs-io/babylon/pull/65/commits/a871d44af7256d48ec854312fffe3bdcd4d719f4)

PR #69, opened 2024-09-12
feat: Babylon FP cascaded slashing

gusin13:
There are 2 user stories related to Cascaded Slashing 

1. When Babylon FP is slashed
- Find all delegations involving this slashed FP
- For each such delegation find all the restaked consumer ID's
- For each involved consumer, send IBC packet to notify this slashing event. 

2. When Consumer FP is slashed on consumer chain. 
- Consumer sends slashing evidence to Babylon through IBC
- After receiving such event, Babylon performs the same steps as in 1. 

This PR implements User story 1. 
Also includes e2e tests.


x/btcstaking/types/btc_staking_consumer_events.go
```diff
@@ -99,3 +99,13 @@ func CreateUnbondedBTCDelegationEvent(unbondedDel *BTCDelegation) (*BTCStakingCo
 
 	return event, nil
 }
+
+func CreateSlashedBTCDelegationEvent(slashedDel *BTCDelegation) *BTCStakingConsumerEvent {
+	return &BTCStakingConsumerEvent{
+		Event: &BTCStakingConsumerEvent_SlashedDel{
+			SlashedDel: &SlashedBTCDelegation{
+				StakingTxHash: slashedDel.MustGetStakingTxHash().String(),
```
SebastianElvis, line 107: The extracted SK is still necessary for this packet? as the consumer side needs to verify the FP is indeed slashed

x/btcstaking/keeper/finality_providers.go
```diff
@@ -67,6 +68,35 @@ func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error
...
+		// Get consumer IDs of non-Babylon finality providers
+		restakedFPConsumerIDs, err := k.restakedFPConsumerIDs(ctx, delegation.FpBtcPkList)
+		if err != nil {
+			return err
+		}
+
+		// Send event to each involved consumer chain
+		for _, consumerID := range restakedFPConsumerIDs {
+			// TODO: repeated set ops in KV store is inefficient; consider refactoring
+			if err := k.AddBTCStakingConsumerEvent(ctx, consumerID, consumerEvent); err != nil {
```
SebastianElvis, line 91: As the first step, we can refactor the function to take a list of `consumerEvent`s, append them to the `BTCStakingIBCPacket`, and set it back to DB

x/btcstaking/keeper/finality_providers.go
```diff
@@ -67,6 +68,35 @@ func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error
...
+func (k Keeper) PropagateFPSlashingToConsumers(ctx context.Context, fpBTCPK *bbn.BIP340PubKey) error {
+	// Get all delegations for this finality provider
+	delegations, err := k.getFPBTCDelegations(ctx, fpBTCPK)
+	if err != nil {
+		return err
+	}
+
+	for _, delegation := range delegations {
+		// Create SlashedBTCDelegation event
+		consumerEvent := types.CreateSlashedBTCDelegationEvent(delegation)
```
SebastianElvis, line 80: This function does not seem optimal indeed.

How about we refactor this parts as following:

- Have a function that given a slashed FP, outputs a map where key is consumer ID and value is a list of `SlashedBTCDelegation` events.
- Given this map, for each consumer ID, we append them to the IBC packet using `AddBTCStakingConsumerEvent`

Here the function can be implemented as follows: for each delegation, gets all involved FPs and then involved consumer IDs, and appends the events to involved consumer IDs in the map

x/btcstaking/types/btc_staking_consumer_events.go
SebastianElvis: > is it ok if we send slashing evidence instead of extracted SK and let the contract handle the extraction process (based on the full-validation feature flag), wdyt?

Yep sounds good to me, as this offsets computational cost on Babylon side to consumer side :+1:

x/btcstaking/keeper/btc_staking_consumer_events.go
```diff
@@ -40,6 +40,37 @@ func (k Keeper) AddBTCStakingConsumerEvent(ctx context.Context, consumerID strin
 	return nil
 }
 
+func (k Keeper) AddBTCStakingConsumerEvents(ctx context.Context, consumerID string, events []*types.BTCStakingConsumerEvent) error {
```
SebastianElvis, line 43: I'd say `AddBTCStakingConsumerEvent` can be removed or simply be a wrapper of this function

x/btcstaking/keeper/finality_providers.go
```diff
@@ -67,6 +68,94 @@ func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error
...
+		seenConsumers := make(map[string]bool)
+
+		for _, delegationFPBTCPK := range delegation.FpBtcPkList {
+			fpBTCPKHex := delegationFPBTCPK.MarshalHex()
+			consumerID, exists := fpToConsumerMap[fpBTCPKHex]
+			if !exists {
+				// If not in map, check if it's a Babylon FP or get its consumer
+				if _, err := k.GetFinalityProvider(ctx, delegationFPBTCPK); err == nil {
+					continue // It's a Babylon FP, skip
+				} else if consumerID, err = k.bscKeeper.GetConsumerOfFinalityProvider(ctx, &delegationFPBTCPK); err == nil {
```
SebastianElvis, line 140: Is calling `GetConsumerOfFinalityProvider` enough? as if a FP is a consumer FP then `GetFinalityProvider` has to return error

SebastianElvis (approved):
Lgtm :+1: great work

x/btcstaking/keeper/finality_providers.go
```diff
@@ -67,6 +68,94 @@ func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error
...
+					// Found consumer, add to map
+					fpToConsumerMap[fpBTCPKHex] = consumerID
+				} else {
+					return nil, types.ErrFpNotFound.Wrapf("finality provider pk %s is not found", fpBTCPKHex)
+				}
+			}
+
+			// Add event to the consumer's event list only if not seen for this delegation
+			if !seenConsumers[consumerID] {
+				consumerEvents[consumerID] = append(consumerEvents[consumerID], consumerEvent)
```
maurolacy, line 150: The ideal struct would be a consumer id map of a set (or boolean map) of delegations. So that you don't need the "seen" helper.

Nit: You can as well create a list of delegations (using their staking hash as id or so) here, and create the consumer events outside this function.

maurolacy (approved):
LGTM.

test/e2e/bcd_consumer_integration_test.go
```diff
@@ -221,16 +240,161 @@ func (s *BCDConsumerIntegrationTestSuite) Test5ActivateDelegation() {
...
+		}
+
+		// check if the consumerFp's BTC public key exists in the map
+		consumerFpBtcPkHex := consumerFp.BtcPk.MarshalHex()
+		fpInfo, exists := fpMap[consumerFpBtcPkHex]
+		if !exists {
+			return false
+		}
+
+		return fpInfo.BtcPkHex == consumerFp.BtcPk.MarshalHex() && fpInfo.Power == 0
```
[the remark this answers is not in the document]
maurolacy: Ah, yes. This FP is not slashed, but rather, its voting power is being discounted the slashed delegations of the slashed Babylon FP.

PR #74, opened 2024-09-16
feat(ADR-25): Unjail jailed finality providers

gitferry:
This PR partially closes https://github.com/babylonlabs-io/babylon/issues/64 by introducing `MsgUnjailFinalityProviders` to allow the fp operator to unjail a jailed finality provider after the jailing period is passed. In particular, this PR made the following changes:
- introduce `MsgUnjailFinalityProvider`, within the handler:
    - get signing info and check current block timestamp
    - call `UnjailFinalityProvider` from the btcstaking module, which
        - checks fp existence
        - checks if the fp is jailed
        - changes the IsJailed flag to false
        - creats and emits EventUnjailFinalityProvider, which will be consumed in the `BeginBlocker` of the next block
- within `BeginBlocker` of the btcstaking module, upon `EventUnjailFinalityProvider`
    - change the `IsJailed` flag of the finality provider dist info to `false` so that its accumulated delegated value will be transferred to actual voting power
- fuzz tests


x/finality/keeper/msg_server.go
```diff
@@ -237,6 +237,33 @@ func (ms msgServer) CommitPubRandList(goCtx context.Context, req *types.MsgCommi
 	return &types.MsgCommitPubRandListResponse{}, nil
 }
 
+func (k Keeper) UnjailFinalityProvider(ctx context.Context, req *types.MsgUnjailFinalityProvider) (*types.MsgUnjailFinalityProviderResponse, error) {
+	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), types.MetricsKeyUnjailFinalityProvider)
+
+	fpPk := req.FpBtcPk
+	info, err := k.FinalityProviderSigningTracker.Get(ctx, fpPk.MustMarshal())
```
RafilxTenfen, line 244: this fpPk comes from the request, should we validate it first, before call `MustMarshal`?

gitferry: fpPk is already a `BIP340PubKey` type. Why do we need to validate it?

proto/babylon/finality/v1/tx.proto
```diff
@@ -73,17 +76,30 @@ message MsgAddFinalitySigResponse{}
...
+
+// MsgUnjailFinalityProvider defines the Msg/UnjailFinalityProvider request type
+message MsgUnjailFinalityProvider {
+    option (gogoproto.goproto_getters) = false;
+    option (cosmos.msg.v1.signer) = "signer";
+
+    string signer = 1;
+    // fp_btc_pk is the BTC PK of the finality provider that commits the public randomness
+    bytes fp_btc_pk = 2 [ (gogoproto.customtype) = "github.com/babylonlabs-io/babylon/types.BIP340PubKey" ];
+}
```
SebastianElvis, line 102: It seems that any BBN account (not just the BBN account that creates the FP) can unjail an arbitrary FP. Is this expected or we want to enforce that only the BBN account that creates the FP can unjail itself?

x/finality/keeper/msg_server.go
```diff
@@ -237,6 +237,33 @@ func (ms msgServer) CommitPubRandList(goCtx context.Context, req *types.MsgCommi
 	return &types.MsgCommitPubRandListResponse{}, nil
 }
 
+func (k Keeper) UnjailFinalityProvider(ctx context.Context, req *types.MsgUnjailFinalityProvider) (*types.MsgUnjailFinalityProviderResponse, error) {
```
SebastianElvis, line 240:
```suggestion
func (ms msgServer) UnjailFinalityProvider(ctx context.Context, req *types.MsgUnjailFinalityProvider) (*types.MsgUnjailFinalityProviderResponse, error) {
```

x/finality/keeper/msg_server.go
```diff
@@ -237,6 +237,33 @@ func (ms msgServer) CommitPubRandList(goCtx context.Context, req *types.MsgCommi
...
+		return nil, fmt.Errorf("failed to get the signing info of finality provider %s: %w", fpPk.MarshalHex(), err)
+	}
+
+	// cannot be unjailed until jailing period is passed
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	curBlockTime := sdkCtx.HeaderInfo().Time
+	if curBlockTime.Before(info.JailedUntil) {
+		return nil, types.ErrJailingPeriodNotPassed.Wrapf(
+			fmt.Sprintf("current block time: %v, required %v", curBlockTime, info.JailedUntil))
+	}
```
SebastianElvis, line 255: What do you think about wrapping L252 into a function of `FinalityProviderSigningInfo`?

SebastianElvis (approved):
Lgtm, only small questions :+1:

proto/babylon/finality/v1/tx.proto
SebastianElvis: OK thinking this further I think some authentication is beneficial here. Otherwise there is an attack: let's say a FP is jailed and the FP operator is not aware of that. Then an adversary keeps unjailing it and the protocol keeps jailing it, such that all rewards accumulated by this FP are slashed, before the FP operator realises that its FP is down. If there is an authentication then this attack won't happen.

The authentication is quite simple: get FP from DB, and compare whether the `Addr` of FP is same as `req.Signer`

gitferry: Nice point! We should definitely have the authentication. Maybe not for the monetary reason but because unplanned unjailing will ruin the fp's reputation.

PR #82, opened 2024-09-19
feat: Update go workflow that includes gosec

maiquanghiep:


filippos47 (approved):
Thanks Hiep! Some things we've also discussed offline:
- As expected, all the service repos have a bulk of spotted gosec issues. If I'm not mistaken, babylond has the most reports (around 2k). It'd be useful if we can use the GH UI to visualize the babylond error report, if it's easy to enable.
- Due to the volume of the reported issues, we'll allow the gosec step to fail initially. The steps to making the gosec step compulsory would be the following:
  - The service owners identify the valid reported issues and resolve them
  - The service owners identify the non-valid reported issues and in conjuction with us silence them (this will be possible through `gosec` flags)
  - Once the above steps are applied and the gosec tool reports no issues, we remove the "allow-to-fail" CLI flag

Lazar955:
> Damm, it is a lot of warnings! - https://github.com/babylonlabs-io/babylon/actions/runs/10985752891/job/30498103625?pr=82 Thanks for this @maiquanghiep

We should prob ignore `.pb.go` and other generated files

maiquanghiep:
Indeed when adding `-exclude-generated` option that:
```
gosec can ignore generated go files with default generated code comment.
// Code generated by some generator DO NOT EDIT.
```
The issues reduce significantly to less than 200
https://github.com/babylonlabs-io/babylon/actions/runs/11043951607

maiquanghiep:
Staled PR

PR #85, opened 2024-09-19
feat: Consumer FP cascaded slashing

gusin13:
There are 2 user stories related to Cascaded Slashing 

1. When Babylon FP is slashed
- Find all delegations involving this slashed FP
- For each such delegation find all the restaked consumer ID's
- For each involved consumer, send IBC packet to notify this slashing event. 

2. When Consumer FP is slashed on consumer chain. 
- Consumer sends slashing evidence to Babylon through IBC
- Babylon receives the evidence, extracts and verifies the SK 
- Adjusts/discounts the voting power distribution table for all involved Babylon FPs 
- Propagates the slashing to all other involved consumers 

This PR implements User story 2 ( Us1 is already merged https://github.com/babylonlabs-io/babylon/pull/69) 
Also includes e2e tests.

### DIFF
--- a/x/zoneconcierge/module_ibc.go
+++ b/x/zoneconcierge/module_ibc.go
@@ -176,7 +176,6 @@ func (im IBCModule) OnRecvPacket(
 	// Add other packet types here if needed
 	default:
 		errMsg := fmt.Sprintf("unrecognized %s packet type: %T", types.ModuleName, packet)
-		im.keeper.Logger(ctx).Error("Unrecognized packet type", "type", fmt.Sprintf("%T", packet))
 		return channeltypes.NewErrorAcknowledgement(errorsmod.Wrap(sdkerrors.ErrUnknownRequest, errMsg))
 	}
 }


contrib/images/Makefile
```diff
@@ -8,7 +8,7 @@ babylond:
 	docker build --tag babylonlabs-io/babylond -f babylond/Dockerfile ${BABYLON_FULL_PATH}
 
 babylond-e2e:
-	docker build --tag babylonlabs-io/babylond -f babylond/Dockerfile ${BABYLON_FULL_PATH} \
+	docker build --no-cache --tag babylonlabs-io/babylond -f babylond/Dockerfile ${BABYLON_FULL_PATH} \
```
SebastianElvis, line 11: probably this is a leftover?

contrib/images/Makefile
```diff
@@ -41,7 +41,7 @@ cosmos-relayer-rmi:
 	docker rmi babylonlabs-io/cosmos-relayer 2>/dev/null; true
 
 build-ibcsim-bcd:
-	docker build --tag babylonlabs-io/ibcsim-bcd -f ibcsim-bcd/Dockerfile ${BABYLON_FULL_PATH}
+	docker build --no-cache --tag babylonlabs-io/ibcsim-bcd -f ibcsim-bcd/Dockerfile ${BABYLON_FULL_PATH}
```
SebastianElvis, line 44: same here

proto/babylon/btcstaking/v1/events.proto
```diff
@@ -37,14 +39,25 @@ message EventPowerDistUpdate {
...
   // TODO: unify with existing slashing events
   message EventSlashedFinalityProvider {
-    bytes pk = 1 [ (gogoproto.customtype) = "github.com/babylonlabs-io/babylon/types.BIP340PubKey" ];
+    bytes pk = 1 [(gogoproto.customtype) = "github.com/babylonlabs-io/babylon/types.BIP340PubKey"];
+  }
+
+  // EventSlashedBTCDelegation defines an event that occurs when a consumer FP is slashed.
+  // This event represents the impact on a specific BTC delegation containing the slashed
+  // consumer FP. Babylon FPs in this delegation will have their voting power discounted
+  // as a result.
```
SebastianElvis, line 48:
```suggestion
  // This event is emitted upon a finality provider that this BTC delegation restakes to is slashed. Babylon FPs in this delegation will have their voting power discounted
  // as a result.
```

x/zoneconcierge/keeper/ibc_packet_btc_staking_consumer_event.go
```diff
@@ -72,3 +74,61 @@ func (k Keeper) HandleConsumerRegistration(
...
+) error {
+	clientID, _, err := k.channelKeeper.GetChannelClientState(ctx, destinationPort, destinationChannel)
+	if err != nil {
+		return fmt.Errorf("failed to get client state: %w", err)
+	}
+
+	evidence := consumerSlashing.Evidence
+	if evidence == nil {
+		return fmt.Errorf("consumer slashing evidence is nil")
+	}
```
SebastianElvis, line 92: This can be due to a misbehaviour of the consumer side, rather than an issue in Babylon side. So we will need to return an IBC acknowledgement with some error, rather than returning error here.

Similar comments apply to other error cases in this function

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -181,6 +187,16 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
+				if _, slashed := slashedBTCDels[btcDel.StakingTxHash]; slashed {
+					// If the BTC delegation is slashed, set its voting power to 0.
+					// This ensures that for this specific finality provider,
+					// the voting power of this delegation does not contribute
+					// to the total voting power. By modifying btcDel directly,
+					// we ensure that in subsequent heights, when the old cache
+					// is fetched, this delegation remains discounted for this FP.
+					btcDel.VotingPower = 0
+				}
 				fp.AddBTCDelDistInfo(&btcDel)
```
SebastianElvis, line 200: - How about unrolling these two if statements? e.g., 
  - if unbonded then `continue`, 
  - if slashed then `continue`,
  - otherwise, execute `fp.AddBTCDelDistInfo(&btcDel)`

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -189,7 +205,13 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
 			// handle new BTC delegations for this finality provider
 			for _, d := range fpActiveBTCDels {
-				fp.AddBTCDel(d)
+				del := *d
+				if _, slashed := slashedBTCDels[d.MustGetStakingTxHash().String()]; slashed {
+					// If the new BTC delegation is slashed, set its voting power to 0
+					del.TotalSat = 0
+				}
+
+				fp.AddBTCDel(&del)
```
SebastianElvis, line 214: Instead of zero-ising `TotalSat` and add this BTC delegation, how aboure we skip adding it if slashed? Basically a slashed BTC delegation shall never appear in the voting power cache

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -230,7 +252,12 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 		// add each BTC delegation
 		fpActiveBTCDels := activeBTCDels[fpBTCPKHex]
 		for _, d := range fpActiveBTCDels {
-			fpDistInfo.AddBTCDel(d)
+			del := *d
+			if _, slashed := slashedBTCDels[d.MustGetStakingTxHash().String()]; slashed {
+				// If the BTC delegation is slashed, set its voting power to 0.
+				del.TotalSat = 0
+			}
+			fpDistInfo.AddBTCDel(&del)
```
SebastianElvis, line 260: Same here, how about we avoid adding this slashed btc del?

x/btcstaking/keeper/btc_delegators.go
```diff
@@ -60,8 +60,31 @@ func (k Keeper) getBTCDelegatorDelegations(ctx context.Context, fpBTCPK *bbn.BIP
...
+	if k.HasFinalityProvider(ctx, *fpBTCPK) {
+		// Babylon finality provider
+		store = k.btcDelegatorFpStore(ctx, fpBTCPK)
+	} else if k.bscKeeper.HasConsumerFinalityProvider(ctx, fpBTCPK) {
+		// Consumer finality provider
+		store = k.btcConsumerDelegatorStore(ctx, fpBTCPK)
+	} else {
+		// if not found in either store, return error
+		return nil, types.ErrFpNotFound
+	}
```
SebastianElvis, line 86: For now, how about we distinguish between functions for Babylon FPs and consumer FPs and split them into two functions?

x/btcstaking/keeper/finality_providers.go
```diff
@@ -68,6 +68,46 @@ func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error
 	return nil
 }
 
+// SlashConsumerFinalityProvider slashes a consumer finality provider with the given PK
+func (k Keeper) SlashConsumerFinalityProvider(ctx context.Context, consumerID string, fpBTCPK *bbn.BIP340PubKey) error {
```
SebastianElvis, line 72: Do we need to emit an event so that the vigilante gets notified and slash?

SebastianElvis: Also we might need to be careful for error handling here. Sometimes this might not be the error in Babylon, but consumer side reports false alarms. For example, it's possible that consumer side reports a slashed FP twice and L81 is triggered, making Babylon panick

SebastianElvis (commented):
Almost complete the picture, amazing! Some initial comments

proto/babylon/btcstaking/v1/events.proto
SebastianElvis: sg

x/zoneconcierge/keeper/ibc_packet_btc_staking_consumer_event.go
SebastianElvis: Ah I see, all good then

x/btcstaking/keeper/btc_delegators.go
SebastianElvis: Hmm indeed, okay then

x/btcstaking/keeper/finality_providers.go
SebastianElvis: > hmm good point, i think for now its not needed as vigilante/btc-slasher is not subscribing to consumer tx events, can add a TODO for now but later would vigilante also monitor consumer chains?

How about we do this in this PR? This is a small change and vigilante in fact does not distinguish between Babylon/consumer tx events. So as long as we emit using code [here](https://github.com/babylonlabs-io/babylon/blob/1d97ebd1b94409083ae06da65e2289506c770e1f/x/finality/keeper/msg_server.go#L247-L251) vigilante shall work

test/e2e/bytecode/btc_staking.wasm
maurolacy, line 1: Update `version.txt`?

maurolacy (commented):
Main logic LGTM. Will check e2e test next.

test/e2e/bcd_consumer_integration_test.go
```diff
@@ -339,22 +326,25 @@ func (s *BCDConsumerIntegrationTestSuite) Test6BabylonFPCascadedSlashing() {
...
+		fp, err := s.babylonController.QueryFinalityProvider(babylonFpBIP340PKHex)
 		if err != nil {
-			s.T().Logf("Error querying finality provider slashed status: %v", err)
+			s.T().Logf("Error querying finality provider: %v", err)
 			return false
 		}
-		return slashed
+		return fp != nil &&
+			fp.FinalityProvider.SlashedBtcHeight > 0 &&
+			fp.FinalityProvider.VotingPower == 0
```
maurolacy, line 347: 👍🏼

test/e2e/bcd_consumer_integration_test.go
```diff
@@ -367,34 +357,156 @@ func (s *BCDConsumerIntegrationTestSuite) Test6BabylonFPCascadedSlashing() {
...
 			return false
 		}
-		if fpsByPower == nil || len(fpsByPower.Fps) == 0 {
+
+		return fpInfo != nil && fpInfo.Power == 0 && fpInfo.BtcPkHex == consumerFp.BtcPk.MarshalHex()
+	}, time.Minute, time.Second*5)
+}
+
+func (s *BCDConsumerIntegrationTestSuite) Test7ConsumerFPCascadedSlashing() {
+	consumerId := "07-tendermint-0"
```
maurolacy, line 371: Nit: Read / set from context.

test/e2e/bcd_consumer_integration_test.go
```diff
@@ -367,34 +357,156 @@ func (s *BCDConsumerIntegrationTestSuite) Test6BabylonFPCascadedSlashing() {
...
+	babylonFpBIP340PKHex := bbntypes.NewBIP340PubKeyFromBTCPK(babylonFpBTCPK2).MarshalHex()
+	s.Eventually(func() bool {
+		fp, err := s.babylonController.QueryFinalityProvider(babylonFpBIP340PKHex)
+		if err != nil {
+			s.T().Logf("Error querying finality provider: %v", err)
 			return false
 		}
+		return fp != nil &&
+			fp.FinalityProvider.VotingPower == 0 && // as there is only 1 delegation which got slashed
+			fp.FinalityProvider.SlashedBtcHeight == 0 // should not be slashed
```
maurolacy, line 494: 👍🏼 Only missing piece here that would be a nice to have, is to have more than one delegation. So that we can check the voting power discount is only for the slashed dels.

test/e2e/bcd_consumer_integration_test.go
```diff
@@ -367,34 +357,156 @@ func (s *BCDConsumerIntegrationTestSuite) Test6BabylonFPCascadedSlashing() {
...
+	// check consumer FP record in Babylon is updated
+	consumerFpBIP340PKHex := consumerFp.BtcPk.MarshalHex()
+	s.Eventually(func() bool {
+		fp, err := s.babylonController.QueryFinalityProvider(consumerFpBIP340PKHex)
+		if err != nil {
+			s.T().Logf("Error querying finality provider: %v", err)
+			return false
+		}
+		return fp != nil &&
+			fp.FinalityProvider.SlashedBtcHeight > 0 // should be recorded slashed
```
maurolacy, line 506: 👍🏼 Nice.

test/e2e/bcd_consumer_integration_test.go
```diff
@@ -777,7 +882,17 @@ func (s *BCDConsumerIntegrationTestSuite) initBabylonController() error {
...
+	// Get the current working directory
+	currentDir, err := os.Getwd()
+	if err != nil {
+		s.T().Fatalf("Failed to get current working directory: %v", err)
+	}
+
 	cfg.BtcStakingContractAddress = "bbnc1nc5tatafv6eyq7llkr2gv50ff9e22mnf70qgjlv737ktmt4eswrqgn0kq0"
+	cfg.ChainID = "bcd-test"
+	cfg.KeyDirectory = filepath.Join(currentDir, "../../contrib/images/ibcsim-bcd/.testnets/bcd/bcd-test")
+	cfg.AccountPrefix = "bbnc"
```
maurolacy, line 895: Not ideal, but good to go.

maurolacy (approved):
LGTM. Great e2e tests.

x/btcstaking/keeper/finality_providers.go
SebastianElvis: Ah I see, all good. Somehow I missed it, sorry :sweat_smile:

SebastianElvis (approved):
Great work :+1:

test/e2e/bcd_consumer_integration/clientcontroller/cosmwasm/msg.go
```diff
@@ -202,6 +207,14 @@ type QueryMsgFinalityProviders struct {
...
+	FinalityProvider FinalityProviderQuery `json:"finality_provider"`
+}
+
+type FinalityProviderQuery struct {
+	BtcPkHex string `json:"btc_pk_hex"`
+}
+
 type QueryMsgDelegations struct {
 	Delegations struct{} `json:"delegations"`
 }
```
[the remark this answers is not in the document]
maurolacy: Yes, this is how I usually saw these structs. The main benefit is that are clearly defined and contained from a single place. Another is that follow the Rust side structs more closely.

test/e2e/bytecode/btc_staking.wasm
maurolacy: Then please just use the contract from main (or version tag, which is better) next time. After all the debugging, you switch to the release version, and confirm that everything works with that version.

maurolacy: Unresolving for visibility.

PR #88, opened 2024-09-20
chore: fork x/staking to handle queue

RafilxTenfen:
Main points:
- Handles if it should call the cosmos-sdk msg handler by the key in context
- Fork the minimum needed for `x/staking` to queue it to epoching if the ctx key is not present


RafilxTenfen:
Not needed, wallets implemented support for wrapped msgs

PR #91, opened 2024-09-24
fix: testnet command minor issues

RafilxTenfen:
- Add default gen state to `ibc` module
- Add `jq` to dockerfile, needed to local deployments
- Specify min gas prices as default `1ubbn` in a single place


RafilxTenfen:
After this is merged to main, I will cherry pick to v0.10.x

gitferry (approved):
Thanks for the fix!

KonradStaniec (approved):

PR #93, opened 2024-09-26
fix: Epoch initial height

gitferry:
When we initialize epoch, we should also initialize epoch 1. Otherwise, epoch 0 will be accessed

### DIFF
--- a/x/epoching/keeper/epochs.go
+++ b/x/epoching/keeper/epochs.go
@@ -37,10 +37,13 @@ func (k Keeper) InitEpoch(ctx context.Context) *types.Epoch {
 	if header.Height > 0 {
 		panic("InitEpoch can be invoked only at genesis")
 	}
+	genesisEpoch := types.NewEpoch(0, 1, 0, &header.Time)
+	k.setEpochInfo(ctx, 0, &genesisEpoch)
 	epochInterval := k.GetParams(ctx).EpochInterval
-	epoch := types.NewEpoch(0, epochInterval, 0, &header.Time)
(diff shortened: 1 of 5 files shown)


Lazar955 (approved):
lgtm

x/epoching/keeper/epochs.go
```diff
@@ -37,10 +37,13 @@ func (k Keeper) InitEpoch(ctx context.Context) *types.Epoch {
...
 		panic("InitEpoch can be invoked only at genesis")
 	}
+	genesisEpoch := types.NewEpoch(0, 1, 0, &header.Time)
+	k.setEpochInfo(ctx, 0, &genesisEpoch)
 	epochInterval := k.GetParams(ctx).EpochInterval
-	epoch := types.NewEpoch(0, epochInterval, 0, &header.Time)
-	k.setEpochInfo(ctx, 0, &epoch)
-	return &epoch
+	firstEpoch := types.NewEpoch(1, epochInterval, 1, &header.Time)
+	k.setEpochInfo(ctx, 1, &firstEpoch)
```
SebastianElvis, line 44: What's the problem that this tries to address? Upon BeginBlock of block 1 the epoch 1 metadata will be generated, so it seems not necessary to initialise it upon InitGenesis (block 0)

SebastianElvis: Sounds good, so how about we keep L40-41 and remove L42-44?

SebastianElvis (approved):

PR #97, opened 2024-09-26
chore: add changelog for pr 91

RafilxTenfen:
Forgot to add changelog in #91


SebastianElvis (approved):
Probably it's a good idea to add CI check for changelog, or have a PR template with checklist?

KonradStaniec:
osmosis have workflow to remaind about changelog - https://github.com/osmosis-labs/osmosis/blob/main/.github/workflows/changelog-entry-reminder.yml, unless pr have special label ` A:no-changelog`. Maybe we should copy it ?

PR #103, opened 2024-09-27
chore: upgrade token distribution

RafilxTenfen:
- Update upgrade name from signet-launch to v1
- A new file with data for token distribution to run during the upgrade. Expected template: 

```json
{
  "token_distribution": [
    {
      "address_sender": "bbn14d97wthm9fqvvdd96ax8lnfppwknndxztevs7k",
      "address_receiver": "bbn13t5cnqj6t0p4xa40cwhmgv4wju0zl6g8slk8rz",
      "amount": 100000
    },
    ...
  ]
}
```

- New function from test suite to run before the upgrade `type PreUpgradeFunc func([]*chain.Config)`


app/upgrades/signetlaunch/upgrades.go
```diff
@@ -131,32 +117,96 @@ func upgradeFinalityParameters(
...
+		}
+
+		sender, err := sdk.AccAddressFromBech32(td.AddressSender)
+		if err != nil {
+			return err
+		}
+
+		amount := sdk.NewCoin(appparams.BaseCoinUnit, sdkmath.NewInt(td.Amount))
+		if err := bankK.SendCoins(ctx, sender, receiver, sdk.NewCoins(amount)); err != nil {
+			return err
```
KonradStaniec, line 161: Just few quesation for better understanding:
- what happens if this returns and error ? does upgrade stop ?
- I imagine `bankK.SendCoins`  does all necessary validation if sender have enough coins etc?

I am just thinking of the scenario if by mistake we have some invalid account as sender and what will happen then. I imagine upgrade will happen on chain which is priavate yet and it will fails, so we will be able easily to re-do it.

app/upgrades/signetlaunch/upgrades_test.go
```diff
@@ -7,14 +7,18 @@ import (
...
+	appparams "github.com/babylonlabs-io/babylon/app/params"
+	"github.com/babylonlabs-io/babylon/test/e2e/util"
 	tmproto "github.com/cometbft/cometbft/proto/tendermint/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	minttypes "github.com/cosmos/cosmos-sdk/x/mint/types"
 	"github.com/stretchr/testify/suite"
 
 	"github.com/babylonlabs-io/babylon/app"
-	v1 "github.com/babylonlabs-io/babylon/app/upgrades/signetlaunch"
+	"github.com/babylonlabs-io/babylon/app/upgrades/signetlaunch"
```
KonradStaniec, line 21: maybe it is time to re-name `signetlaunch` to standard `v1` naming as in other chains ? (i.e https://github.com/osmosis-labs/osmosis/tree/main/app/upgrades) 

I imagine phase-2 upgrade will be v1, phase-3 will be v2, ... etc. ?

wdyt ? (it can be done in separate pr though)

app/upgrades/signetlaunch/upgrades.go
RafilxTenfen: yes, if it fails it is going to panic and stop the upgrade 
The upgrade is called during `PreBlocker` I expect that the context is handled with commit and rollback in case of failures, but in our case, since it is all private we will be able to run it again

app/upgrades/signetlaunch/upgrades_test.go
RafilxTenfen: sounds good, I will rename the package

RafilxTenfen: renamed to v1 at https://github.com/babylonlabs-io/babylon/pull/103/commits/b63978f3a8e45a997c779b3d8d58753e6fa19790

KonradStaniec (approved):

PR #107, opened 2024-09-30
feat: in-protocol minimum gas price

SebastianElvis:
Resolves https://github.com/babylonlabs-io/pm/issues/49

This PR introduces the in-protocol minimum gas price mechanism, in which the consensus (more specifically the AnteHandler) enforces every tx has to set a gas price at least 0.002 ubbn. The PR also provides relevant tests. The impl closely follows https://github.com/celestiaorg/celestia-app/pull/2985.

In addition, this PR creates a new package `app/ante` to abstract out the construction of the AnteHandler for Babylon, following the practice at Celestia.

- RFC: https://github.com/babylonlabs-io/pm/pull/56
- ADR: https://github.com/babylonlabs-io/pm/pull/61


app/ante/ante.go
```diff
@@ -0,0 +1,91 @@
...
+		wasmapp.HandlerOptions{
+			HandlerOptions: authante.HandlerOptions{
+				AccountKeeper:   accountKeeper,
+				BankKeeper:      bankKeeper,
+				SignModeHandler: signModeHandler,
+				FeegrantKeeper:  feegrantKeeper,
+				SigGasConsumer:  authante.DefaultSigVerificationGasConsumer,
+				// CheckTxFeeWithGlobalMinGasPrices will enforce the global minimum
+				// gas price for all transactions.
+				TxFeeChecker: CheckTxFeeWithGlobalMinGasPrices,
```
SebastianElvis, line 51: NOTE: this is the place where min gas price is enforced

Lazar955 (approved):
Nice work, lgtm

app/params/config.go
```diff
@@ -20,6 +20,18 @@ const (
 	Bech32PrefixAccAddr = "bbn"
 )
 
+// taken from https://github.com/celestiaorg/celestia-app/pull/2985
+const (
+	// DefaultMinGasPrice is the default min gas price that gets set in the app.toml file.
+	// The min gas price acts as a filter. Transactions below that limit will not pass
+	// a nodes `CheckTx` and thus not be proposed by that node.
+	DefaultMinGasPrice = 0.002
```
RafilxTenfen, line 28: The gas price is already in microbbn `1ubbn` == 1^10-6 bbn
setting the minGasPrice as 1ubbn is too much?

SebastianElvis: Imo 1ubbn is too much. Typically a tx costs >100,000 gas (sometimes it can be 500,000 - 1,000,000 ubbn). With 1ubbn gas price it means that a tx costs at least 0.1bbn. This is a bit too much for an appchain.

Probably that's why Celestia uses 0.002, which is the value here

RafilxTenfen (approved):
:100:

RafilxTenfen:
We should probably update https://github.com/babylonlabs-io/babylon/blob/83fb0e789f8ab18e3a4be17d0f363c3068ecf391/cmd/babylond/cmd/custom_babylon_config.go#L46 
To use the min global default

PR #111, opened 2024-10-01
feat(adr-26): Separate inclusion proof from staking tx info

gitferry:

This PR implements the  separates the handling inclusion proof from btc delegation. In particular, this PR:
1. introduces a new parameter `delegation_creation_base_gas_fee_sat` which will be used to consume more gas when the inclusion proof is not included in the delegation
2. separated the handling of inclusion proof into a different method
3. do not set start height and end height if the inclusion proof is not included
4. do not emit voting power change even if the inclusion proof is not included

A subsequent PR will introduce a new message to handle the inclusion proof


proto/babylon/btcstaking/v1/btcstaking.proto
```diff
@@ -195,4 +196,11 @@ message SelectiveSlashingEvidence {
     // the covenant adaptor/Schnorr signature pair. It is the consequence
     // of selective slashing.
     bytes recovered_fp_btc_sk = 3;
-  }
+}
+
+message InclusionProof {
```
KonradStaniec, line 201: Message needs doc comments as otherwise `make proto-lint` won't pass.

A bit tangent, I am not sure how `make proto-lint` got out of CI, most probably during migration to re-usable pipelines.

proto/babylon/btcstaking/v1/params.proto
```diff
@@ -55,6 +55,8 @@ message Params {
   ];
   // max_active_finality_providers is the maximum number of active finality providers in the BTC staking protocol
   uint32 max_active_finality_providers = 13;
+  // base gas fee for delegation creation in Satoshi
+  int64 delegation_creation_base_gas_fee_sat = 14;
```
KonradStaniec, line 59: imo this should be an `uint64` as negative value do not make sense here and we also avoid type conversion in `ConsumeGas`

x/btcstaking/keeper/msg_server.go
```diff
@@ -173,43 +171,24 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
-		return nil, types.ErrInvalidStakingTx.Wrapf("not k-deep: k=%d; depth=%d", btccParams.BtcConfirmationDepth, stakingTxDepth)
-	}
-	// ensure staking tx's timelock has more than w BTC blocks left
-	if btcTip.Height+btccParams.CheckpointFinalizationTimeout >= endHeight {
-		return nil, types.ErrInvalidStakingTx.Wrapf("staking tx's timelock has no more than w(=%d) blocks left", btccParams.CheckpointFinalizationTimeout)
+		startHeight = inclusionHeight
+		endHeight = startHeight + uint64(parsedMsg.StakingTime)
+	} else {
+		// consume base gas fee to compensate future submission of inclusion proof
+		ctx.GasMeter().ConsumeGas(uint64(ms.GetParams(ctx).DelegationCreationBaseGasFeeSat), "delegation creation fee")
```
KonradStaniec, line 191: params are already retrieved in line:
```
// 5. Validate parsed message against parameters
vp := ms.GetParamsWithVersion(ctx)
```
so lets avoid making another roundtrip to database

x/btcstaking/types/create_delegation_parser.go
```diff
@@ -104,20 +104,20 @@ type ParsedProofOfInclusion struct {
 }
 
 func NewParsedProofOfInclusion(
-	info *btcckpttypes.TransactionInfo,
+	ip *InclusionProof,
 ) (*ParsedProofOfInclusion, error) {
-	if info == nil {
-		return nil, fmt.Errorf("cannot parse nil *btcckpttypes.TransactionInfo")
+	if ip == nil {
```
KonradStaniec, line 109: I think we need to allow `nil` inclusion proof as  now this proof is optional in `MsgCreateDelegation`

Also in the struct `ParsedCreateDelegationMessage` I would leave doc comment for `StakingTxProofOfInclusion` field that, so something along the lines:
```
// StakingTxInclusionProof is optional is and it is up to the caller to verify
// whether it is present or not
StakingTxProofOfInclusion  *ParsedProofOfInclusion
```

x/btcstaking/client/cli/tx.go
```diff
@@ -215,67 +214,72 @@ func NewCreateBTCDelegationCmd() *cobra.Command {
...
 
-			// get staking tx info
-			stakingTxInfo, err := btcctypes.NewTransactionInfoFromHex(args[2])
+			// get staking tx bytes
+			stakingTx, err := hex.DecodeString(args[2])
+			if err != nil {
+				return err
+			}
+
+			inclusionProof, err := types.NewInclusionProofFromHex(args[3])
```
KonradStaniec, line 223: shouldn't we call the `types.NewInclusionProofFromHex(` only if `len(args[3]) > 0`  i.e if user provides empty string as inclusion proof we create message where inclusion proof is `nil` ?

x/btcstaking/keeper/msg_server.go
```diff
@@ -173,43 +171,24 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
-	if stakingTxDepth < btccParams.BtcConfirmationDepth {
-		return nil, types.ErrInvalidStakingTx.Wrapf("not k-deep: k=%d; depth=%d", btccParams.BtcConfirmationDepth, stakingTxDepth)
-	}
-	// ensure staking tx's timelock has more than w BTC blocks left
-	if btcTip.Height+btccParams.CheckpointFinalizationTimeout >= endHeight {
-		return nil, types.ErrInvalidStakingTx.Wrapf("staking tx's timelock has no more than w(=%d) blocks left", btccParams.CheckpointFinalizationTimeout)
+		startHeight = inclusionHeight
+		endHeight = startHeight + uint64(parsedMsg.StakingTime)
+	} else {
+		// consume base gas fee to compensate future submission of inclusion proof
```
KonradStaniec, line 190: hmm this is a bit misleading, I would say that we consume more gas to protect babylon chain and covenant members against spamming i.e creating delegation that will never reach btc chain

proto/babylon/btcstaking/v1/params.proto
gitferry: I thought we should keep consistent with other parameters e.g., `min_slashing_tx_fee_sat` and `unbonding_fee_sat`

KonradStaniec: ahh I see confusion!

This should be `delegation_creation_base_gas_fee` , not `delegation_creation_base_gas_fee_sat` i.e the additional gas fee is not in satoshis but in cosmos gas unit. 

`min_slashing_tx_fee_sat` and `unbonding_fee_sat` are used as `int64` becouse thats what `btcd` library uses, so in those cases minimize amount of type conversions/

KonradStaniec (approved):
LGTM!

PR #113, opened 2024-10-01
chore: upgrade multibuild

RafilxTenfen:


app/upgrades/v1/testnet/btc_headers.go
RafilxTenfen, line 1: At the moment, data files from the mainnet and testnet are the same but just to simulate/test, when we get close to launch, the testnet and mainnet will have different data files

app/upgrades/v1/data_params_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/babylonlabs-io/babylon/app"
+	v1 "github.com/babylonlabs-io/babylon/app/upgrades/v1"
+)
+
+func TestHardCodedBtcStakingParamsAreValid(t *testing.T) {
+	bbnApp := app.NewTmpBabylonApp()
+	for _, upgradeData := range UpgradeV1Data {
```
RafilxTenfen, line 14: each unit test now check both testnet and mainnet data files

app/upgrades/v1/upgrades.go
```diff
@@ -34,70 +34,66 @@ import (
...
-	SignedTxsFP []any `json:"signed_txs_create_fp"`
-}
-
-type DataTokenDistribution struct {
-	TokenDistribution []struct {
-		AddressSender   string `json:"address_sender"`
-		AddressReceiver string `json:"address_receiver"`
-		Amount          int64  `json:"amount"`
-	} `json:"token_distribution"`
+func CreateUpgrade(upgradeDataStr UpgradeDataString) upgrades.Upgrade {
```
RafilxTenfen, line 40: avoid having a var for each upgrade, by accepting the upgrade data as parameter

contrib/images/Makefile
```diff
@@ -9,7 +9,7 @@ babylond: babylond-rmi
 
 babylond-e2e:
 	docker build --tag babylonlabs-io/babylond -f babylond/Dockerfile ${BABYLON_FULL_PATH} \
-		--build-arg BUILD_TAGS="e2e"
+		--build-arg BABYLON_BUILD_OPTIONS="testnet"
```
RafilxTenfen, line 12: Fixed to use correct build arg, `BUILD_TAGS` should only be set by the makefile

app/include_upgrade_testnet.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/babylonlabs-io/babylon/app/upgrades"
+	v1 "github.com/babylonlabs-io/babylon/app/upgrades/v1"
+	"github.com/babylonlabs-io/babylon/app/upgrades/v1/testnet"
+)
+
+// init is used to include v1 upgrade testnet data
+// it is also used for e2e testing
+func init() {
+	Upgrades = []upgrades.Upgrade{v1.CreateUpgrade(v1.UpgradeDataString{
```
RafilxTenfen, line 14: only place beside tests that imports testnet and is included only if there is a `testnet` in build tags

app/include_upgrade_mainnet.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"github.com/babylonlabs-io/babylon/app/upgrades"
+	v1 "github.com/babylonlabs-io/babylon/app/upgrades/v1"
+	"github.com/babylonlabs-io/babylon/app/upgrades/v1/mainnet"
+)
+
+// init is used to include v1 upgrade for mainnet data
+func init() {
+	Upgrades = []upgrades.Upgrade{v1.CreateUpgrade(v1.UpgradeDataString{
```
RafilxTenfen, line 13: only place beside tests that imports mainnet and is included only if there is a `mainnet` in build tags
> Makefile sets by default this build tag

app/include_upgrade_mainnet.go
```diff
@@ -0,0 +1,20 @@
+//go:build mainnet
```
SebastianElvis, line 1: I remember Go allows to build with multiple tags e.g., `-tags "mainnet,testnet"`. Wondering what will happen in this case :thinking:

app/upgrades/types.go
```diff
@@ -26,12 +26,15 @@ type Upgrade struct {
 	UpgradeName string
 
 	// CreateUpgradeHandler defines the function that creates an upgrade handler
-	CreateUpgradeHandler func(*module.Manager, module.Configurator, *keepers.AppKeepers) upgradetypes.UpgradeHandler
+	CreateUpgradeHandler CreateUpgradeHandler
```
SebastianElvis, line 29: No strong opinion but I'm a bit hesitant on using the same name for variable name and type

SebastianElvis (approved):
I have less context here, so only some (non-blocking) general questions

app/include_upgrade_mainnet.go
RafilxTenfen: good question, in both cases I am replacing the upgrades var from app.go with a new upgrade 
So, the question is which one ? :sweat_smile: 

One other point is that in the makefile or it includes mainnet or testnet to the build tags 
https://github.com/babylonlabs-io/babylon/blob/aea173fe43423b42918d8f827976856c69179239/Makefile#L82-L87

app/upgrades/types.go
RafilxTenfen: update type to UpgradeHandlerCreator in https://github.com/babylonlabs-io/babylon/pull/113/commits/42b62c2c0dc8be6637c90164955f0b4031e4baf9

KonradStaniec (approved):
Looks great!

PR #115, opened 2024-10-02
feat(adr-26): Add `MsgAddBTCDelegationInclusionProof`

gitferry:

This PR follows #111 by introducing a new message `MsgAddBTCDelegationInclusionProof` and modifying other message handlers. In particular, this PR:
* introduced `MsgAddBTCDelegationInclusionProof` and the relevant handler
* introduced `staking_time` to `btc_delegation` as db schema change. This is needed to calculate end height of the delegation when the inclusion proof is received
* modified handlers `AddCovenantSigs` and `BTCUndelegate` to emit voting power event only if the inclusion proof is received


KonradStaniec:
closed in favour of - https://github.com/babylonlabs-io/babylon/pull/118

PR #125, opened 2024-10-03
feat(ADR-028): refunding transaction fee for certain txs

### LINKED ISSUE #124: transaction fee refunding mechanism
Messages to refund

- [ ] MsgInsertHeaders
- [x] MsgInsertBTCSpvProof
- [x] MsgAddCovenantSigs
- [x] MsgUndelegate
- [x] MsgActivateBTCDelegation
- [x] MsgSelectiveSlashingEvidence
- [x] MsgAddFinalitySig

SebastianElvis:
Resolves https://github.com/babylonlabs-io/pm/issues/64

This PR introduces the functionality of refunding fee for certain txs, including BTC headers/checkpoints in BTC timestamping protocol, and finality signature, BTC delegation inclusion proof, covenant signatures, undelegation, selective slashing evidence in BTC staking protocol.

The implementation leverages a new key-only KV store in the incentive module for storing refundable messages, and a new PostHandler for refunding a tx if all msgs in it are refundable.

Along the way, this PR also adds dependency from BTC light client / BTC staking modules to the incentive module, and adds relevant e2e tests to ensure that the tx fee refunding indeed works.

- RFC: https://github.com/babylonlabs-io/pm/blob/main/rfc/rfc-010-transaction-fee-refund-protocol.md
- ADR: https://github.com/babylonlabs-io/pm/blob/main/adr/adr-028-transaction-fee-refund-protocol.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### State Machine Breaking
 
+* [#125](https://github.com/babylonlabs-io/babylon/pull/125) Implement ADR-028 and
+refund transaction fee for certain transactions from protocol stakeholders
 * [#107](https://github.com/babylonlabs-io/babylon/pull/107) Implement ADR-027 and
 enable in-protocol minimum gas price
 * [#103](https://github.com/babylonlabs-io/babylon/pull/103) Add token distribution
--- a/test/e2e/configurer/chain/commands.go
+++ b/test/e2e/configurer/chain/commands.go
@@ -235,10 +235,10 @@ func (n *NodeConfig) FinalizeSealedEpochs(startEpoch uint64, lastEpoch uint64) {
 		n.SubmitRefundableTxWithAssertion(func() {
 			n.InsertHeader(&opReturn1.HeaderBytes)
 			n.InsertHeader(&opReturn2.HeaderBytes)
-		})
+		}, true)
 		n.SubmitRefundableTxWithAssertion(func() {
 			n.InsertProofs(opReturn1.SpvProof, opReturn2.SpvProof)
-		})
+		}, true)
 
 		n.WaitForCondition(func() bool {
 			ckpt, err := n.QueryRawCheckpoint(checkpoint.Ckpt.EpochNum)
@@ -453,6 +453,7 @@ func (n *NodeConfig) TxGovVote(from string, propID int, option govv1.VoteOption,
 // and asserts that the tx fee is refunded
 func (n *NodeConfig) SubmitRefundableTxWithAssertion(
 	f func(),
+	shouldBeRefunded bool,
 ) {
 	// balance before submitting the refundable tx
 	submitterBalanceBefore, err := n.QueryBalances(n.PublicAddress)
@@ -464,5 +465,9 @@ func (n *NodeConfig) SubmitRefundableTxWithAssertion(
 	// ensure the tx fee is refunded and the balance is not changed
 	submitterBalanceAfter, err := n.QueryBalances(n.PublicAddress)
 	require.NoError(n.t, err)
-	require.Equal(n.t, submitterBalanceBefore, submitterBalanceAfter)
+	if shouldBeRefunded {
+		require.Equal(n.t, submitterBalanceBefore, submitterBalanceAfter)
+	} else {
+		require.True(n.t, submitterBalanceBefore.IsAllGT(submitterBalanceAfter))
+	}
 }
--- a/test/e2e/configurer/chain/commands.go
+++ b/test/e2e/configurer/chain/commands.go
@@ -232,9 +232,10 @@ func (n *NodeConfig) FinalizeSealedEpochs(startEpoch uint64, lastEpoch uint64) {
 		tx2 := datagen.CreatOpReturnTransaction(r, p2)
 		opReturn2 := datagen.CreateBlockWithTransaction(r, opReturn1.HeaderBytes.ToBlockHeader(), tx2)
 
-		n.InsertHeader(&opReturn1.HeaderBytes)
-		n.InsertHeader(&opReturn2.HeaderBytes)
-
+		n.SubmitRefundableTxWithAssertion(func() {
+			n.InsertHeader(&opReturn1.HeaderBytes)
+			n.InsertHeader(&opReturn2.HeaderBytes)
+		})
 		n.SubmitRefundableTxWithAssertion(func() {
 			n.InsertProofs(opReturn1.SpvProof, opReturn2.SpvProof)
 		})
--- a/test/e2e/configurer/chain/commands.go
+++ b/test/e2e/configurer/chain/commands.go
@@ -234,7 +234,10 @@ func (n *NodeConfig) FinalizeSealedEpochs(startEpoch uint64, lastEpoch uint64) {
 
 		n.InsertHeader(&opReturn1.HeaderBytes)
 		n.InsertHeader(&opReturn2.HeaderBytes)
-		n.InsertProofs(opReturn1.SpvProof, opReturn2.SpvProof)
+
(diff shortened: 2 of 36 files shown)


test/e2e/configurer/chain/commands.go
```diff
@@ -444,3 +448,26 @@ func (n *NodeConfig) TxGovVote(from string, propID int, option govv1.VoteOption,
...
+func (n *NodeConfig) SubmitRefundableTxWithAssertion(
+	f func(),
+	shouldBeRefunded bool,
+) {
+	// balance before submitting the refundable tx
+	submitterBalanceBefore, err := n.QueryBalances(n.PublicAddress)
+	require.NoError(n.t, err)
+
+	// submit refundable tx
+	f()
```
RafilxTenfen, line 463: should we wait for a block to know if the tx was included in the block? and then check new balances

x/incentive/keeper/refund_tx_decorator.go
```diff
@@ -0,0 +1,64 @@
...
+		msgHash := types.HashMsg(msg)
+		if d.k.HasRefundableMsg(ctx, msgHash) {
+			refundableMsgHashList = append(refundableMsgHashList, msgHash)
+		}
+	}
+
+	// if all messages in the tx are refundable, refund the tx
+	if len(refundableMsgHashList) == len(tx.GetMsgs()) {
+		err := d.k.RefundTx(ctx, feeTx)
+		if err != nil {
```
RafilxTenfen, line 49: why not return the error if it fails to refund?

RafilxTenfen (approved):
LGTM, nice PR! 
Only one question

test/e2e/configurer/chain/commands.go
SebastianElvis: Currently I leave this as responsibility for the function caller. Somestimes there is more sophisticated check, e.g., query the inclusion of a finality signature. Wdyt?

x/incentive/keeper/refund_tx_decorator.go
SebastianElvis: If this returns error then the tx will be rejected and all resulting state transition will be reverted. It's a bit unexpected to fail the tx due to fee refund, but tbh not sure if it's a good thing to do. If `RefundTx` fails it might mean something is really wrong. Will think more about it tmr

CHANGELOG.md
```diff
@@ -41,6 +41,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### State Machine Breaking
 
+* [#125](https://github.com/babylonlabs-io/babylon/pull/125) Implement ADR-028 and
```
KonradStaniec, line 44: this should go to unreleased

x/btccheckpoint/keeper/msg_server.go
```diff
@@ -92,6 +92,11 @@ func (ms msgServer) InsertBTCSpvProof(ctx context.Context, req *types.MsgInsertB
 		return nil, err
 	}
 
+	// At this point, the BTC checkpoint is considered the first valid submission
```
KonradStaniec, line 95: this is not exactly true as it may be as well not first submission. If the intention is to refund only first one, the place to do so would be elsewere.

On the other hand, I thought we want to refund all valid submissions ?

x/btcstaking/keeper/msg_server.go
```diff
@@ -466,6 +470,14 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
 		params,
 	)
 
+	// at this point, the covenant signatures are verified and are not duplicated.
+	// Thus, we can safely consider this message as refundable
+	// NOTE: currently we refund tx fee for covenant signatures even if the BTC
```
KonradStaniec, line 475: hmm this is not entirely true tbh. We have check for quorum super early in the handler:
```
	if btcDel.HasCovenantQuorums(params.CovenantQuorum) {
		ms.Logger(ctx).Debug("Received covenant signature after achieving quorum", "covenant pk", req.Pk.MarshalHex())
		return &types.MsgAddCovenantSigsResponse{}, nil
	}
```

so this won't be refunded. And refunding it would require a bit of refactor.

Either way I think we cannot refund such case, as then covenant committe member could send his signature forever and bloating chain with it.

x/incentive/keeper/refund_tx_decorator.go
```diff
@@ -0,0 +1,64 @@
...
+	if !ok {
+		return next(ctx, tx, simulate, success)
+	}
+
+	refundableMsgHashList := make([][]byte, 0)
+
+	// iterate over all messages in the tx, and record whether they are refundable
+	for _, msg := range tx.GetMsgs() {
+		msgHash := types.HashMsg(msg)
+		if d.k.HasRefundableMsg(ctx, msgHash) {
```
KonradStaniec, line 41: I think we need to filter our the duplicated form the message list or do refunds only in case when there is 1 message in transaction. 

Consider case, if finality provider woud create transactions with finality votes:
1. First vote would be valid finality vote which would be indexable as to rewards
2. All other votes would be duplicates

Transaction would be successful as duplicated vote is not an error:
```
	if err == nil && existingSig.Equals(req.FinalitySig) {
		ms.Logger(ctx).Debug("Received duplicated finiality vote", "block height", req.BlockHeight, "finality provider", req.FpBtcPk)
		// exactly same vote already exists, return success to the provider
		return &types.MsgAddFinalitySigResponse{}, nil
	}
```
Not finality provider can bloat the ledgers for free i.e he can send transactions with 100 duplicated votes and not pay anything.

SebastianElvis: In this example, upon duplicated finality signature the function will return directly without calling `IndexRefundableMsg`, so [this condition](https://github.com/babylonlabs-io/babylon/blob/6e5210d0eacca6570860152d92ec58f4da7c8cf3/x/incentive/keeper/refund_tx_decorator.go#L47-L48) won't be met and the tx won't get refunded. Or did I miss anything?

x/btcstaking/keeper/msg_server.go
SebastianElvis: Hmm I see. So this won't get refunded in the current impl

I actually think we should refund when covenant quorum is reached. Otherwise for each BTC delegation the covenant member has a rough probability of `(covenant size - covenant quorum) / covenant size` to pay some tx fee, so running a covenant emulator won't be free. But the goal of refunding is to ensure running covenant emulator does not incur tx fee.

> Either way I think we cannot refund such case, as then covenant committe member could send his signature forever and bloating chain with it.

Does this attack vector exist? for each BTC delegation a covenant member can only send 1 valid covenant signature, and subsequent ones will be rejected and won't get refunded ([ref](https://github.com/babylonlabs-io/babylon/blob/6e5210d0eacca6570860152d92ec58f4da7c8cf3/x/btcstaking/keeper/msg_server.go#L354-L358)).

x/incentive/keeper/refund_tx_decorator.go
SebastianElvis: Per offline discussion, this is an issue: the problem is that these messages share the same hash and then all of them are considered refundable in the posthandler. Two ways of addressing this issue:
deduplicate hashes in posthandler before comparing the length
return error upon duplicated finality signature

x/btccheckpoint/keeper/msg_server.go
SebastianElvis: Yeah this is a typo we want to refund all valid submissions (more specifically, every unique pair of BTC txs). Will clarify in a follow-up

x/btcstaking/keeper/msg_server.go
KonradStaniec: oh you are right. 

So if we reject duplicates early and not refund them this is fine.

Though handler need refactor either way, as we should refund signatures after quorum only if they are valid and pass all checks.

PR #128, opened 2024-10-03
f/Better IBC transfer tests

maurolacy:
Improve IBC coin transfer tests.

### DIFF
--- a/test/e2e/ibc_transfer_e2e_test.go
+++ b/test/e2e/ibc_transfer_e2e_test.go
@@ -46,9 +46,9 @@ func (s *IBCTransferTestSuite) Test1IBCTransfer() {
 	s.NoError(err)
 
 	sender := initialization.ValidatorWalletName
-	babylonNode.SendIBCTransfer(sender, sender, "", sdk.NewInt64Coin("ubbn", 1000))
+	babylonNode.SendIBCTransfer(sender, sender, "", sdk.NewInt64Coin("ubbn", 1_000_000))
 
-	time.Sleep(1 * time.Minute)
+	time.Sleep(2 * time.Minute)
 
 	// TODO: check the transfer is successful. Right now this is done by manually looking at the log
 }
--- a/test/e2e/ibc_transfer_e2e_test.go
+++ b/test/e2e/ibc_transfer_e2e_test.go
@@ -124,9 +124,7 @@ func (s *IBCTransferTestSuite) Test2IBCTransferBack() {
 	s.Require().Len(balanceB, 2)
 	// Look for the ugly IBC one
 	denom := getFirstIBCDenom(balanceB)
-	// Confirm val on B has enough funds
 	amount := balanceB.AmountOf(denom).Int64() - int64(delta) // have to pay gas fees
-	s.Assert().GreaterOrEqual(balanceB.AmountOf(denom).Int64(), amount)
 
 	transferCoin := sdk.NewInt64Coin(denom, amount)
 
--- a/test/e2e/ibc_transfer_e2e_test.go
+++ b/test/e2e/ibc_transfer_e2e_test.go
@@ -40,12 +40,25 @@ func (s *IBCTransferTestSuite) TearDownSuite() {
 	}
 }
 
+func getFirstIBCDenom(balance sdk.Coins) string {
+	// Look up the ugly IBC denom
+	denoms := balance.Denoms()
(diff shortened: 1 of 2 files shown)


test/e2e/ibc_transfer_e2e_test.go
```diff
@@ -39,16 +40,114 @@ func (s *IBCTransferTestSuite) TearDownSuite() {
...
+	// Send transfer from val in chain-A (Node 3) to val in chain-B (Node 3)
+	babylonNodeA.SendIBCTransfer(val, addrB, "transfer", transferCoin)
+
+	time.Sleep(10 * time.Second)
+
+	// Check that the transfer is successful.
+	// Amounts have been discounted from val in chain-A and added (as a wrapped denom) to val in chain-B
+	balanceA2, err := babylonNodeA.QueryBalances(addrA)
+	s.Require().NoError(err)
+	s.Assert().InDelta(balanceA.Sub(transferCoin).AmountOf(denom).Int64(), balanceA2.AmountOf(denom).Int64(), delta)
```
Lazar955, line 95: small nit, can we use `require.eventually` instead of timeout

test/e2e/ibc_transfer_e2e_test.go
```diff
@@ -39,16 +40,114 @@ func (s *IBCTransferTestSuite) TearDownSuite() {
...
+	// Send transfer from val in chain-B (Node 3) to val in chain-A (Node 1)
+	addrA := babylonNodeA.GetWallet(val)
+	balanceA, err := babylonNodeA.QueryBalances(addrA)
+	s.Require().NoError(err)
 
-	time.Sleep(1 * time.Minute)
+	babylonNodeB.SendIBCTransfer(val, addrA, "transfer back", transferCoin)
 
-	// TODO: check the transfer is successful. Right now this is done by manually looking at the log
+	time.Sleep(10 * time.Second)
```
Lazar955, line 138: same here

RafilxTenfen:
One tip is that osmosis made some IBC testing https://github.com/osmosis-labs/osmosis/blob/main/tests/e2e/e2e_test.go#L154-L172
and we copied most of the e2e testing code from them, so it might make sense to take a look there on what they are testing

test/e2e/ibc_transfer_e2e_test.go
maurolacy: Yes, can do, but not sure what to wait for. Will take a look. 10 seconds is not so bad... it was 30 before. :-)

maurolacy:
> )smosis made some IBC testing https://github.com/osmosis-labs/osmosis/blob/main/tests/e2e/e2e_test.go#L154-L172 and we copied most of the e2e testing code from them, so it might make sense to take a look there on what they are testing

Will take a look, thanks.

test/e2e/ibc_transfer_e2e_test.go
maurolacy: Ah, `eventually` is just the condition and a timeout, good.

RafilxTenfen (approved):
LGTM, this could be also added to main

Lazar955 (approved):

test/e2e/ibc_transfer_e2e_test.go
```diff
@@ -39,16 +41,132 @@ func (s *IBCTransferTestSuite) TearDownSuite() {
...
+			return false
+		}
+		// Check that there are now two denoms in B
+		if len(balanceB2) != 2 {
+			return false
+		}
+		denomB := getFirstIBCDenom(balanceB2)
+		// Check the balance of the IBC denom
+		return math.Abs(float64(balanceB2.AmountOf(denomB).Int64()-
+			transferCoin.Amount.Int64())) < delta
```
SebastianElvis, line 113: Does it also makes sense to ensure the suffix of `denomB` is same as `denom` in chainA?

SebastianElvis (approved):
Thanks for confirming this!

test/e2e/ibc_transfer_e2e_test.go
SebastianElvis: Checked IBC doc and it seems no :( the denom will be `ibc/XXX` where `XXX` is the has of the path but not the denom in A. Nvm

maurolacy:
> this could be also added to main

Will cherry-pick to `main` after merging.

PR #130, opened 2024-10-04
chore: tx fee refunding follow-up

SebastianElvis:
This follow-up PR fixes the comments in https://github.com/babylonlabs-io/babylon/pull/125, including

- changelog and typo comments in BTCCheckpoint module
- allow covenant signatures after covenant quorum are reached, and return error if the covenant signature is duplicated. This is to ensure covenants won't have operational cost when submitting covenant sig late, and avoid refunding to duplicated covenant signatures
- reject duplicated finality signature. This is to avoid refunding to duplicated covenant signatures
- add a dedup check in the PostHandler, to ensure one won't exploit the refunding by having many duplicated messages in a single tx. Also added a fuzz test for this.


x/finality/keeper/msg_server.go
```diff
@@ -97,9 +97,10 @@ func (ms msgServer) AddFinalitySig(goCtx context.Context, req *types.MsgAddFinal
 	}
 	existingSig, err := ms.GetSig(ctx, req.BlockHeight, fpPK)
 	if err == nil && existingSig.Equals(req.FinalitySig) {
-		ms.Logger(ctx).Debug("Received duplicated finiality vote", "block height", req.BlockHeight, "finality provider", req.FpBtcPk)
-		// exactly same vote already exists, return success to the provider
-		return &types.MsgAddFinalitySigResponse{}, nil
+		ms.Logger(ctx).Debug("Received duplicated finality vote", "block height", req.BlockHeight, "finality provider", req.FpBtcPk)
+		// exactly same vote already exists, return error
+		// this is to secure the tx refunding against duplicated messages
+		return nil, types.ErrDuplicatedFinalitySig
```
KonradStaniec, line 103: I would wait for @gitferry review here as he may have some insight how this will influence finality provider side

KonradStaniec (approved):
Looks good!

x/finality/keeper/msg_server.go
SebastianElvis: I think we will need to add this error as expected error [here](https://github.com/babylonlabs-io/finality-provider/blob/74baa4c0cb1e802608a3f721064cab1dfbb99267/clientcontroller/babylon.go#L178-L259). After that it should be good

RafilxTenfen (approved):
LGTM :tada:

x/btcstaking/keeper/msg_server.go
```diff
@@ -353,21 +353,18 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
...
-
-	// ensure BTC delegation is still pending, i.e., not expired
+	// ensure BTC delegation is still pending, i.e., not unbonded
 	btcTipHeight := ms.btclcKeeper.GetTipInfo(ctx).Height
 	wValue := ms.btccKeeper.GetParams(ctx).CheckpointFinalizationTimeout
 	status := btcDel.GetStatus(btcTipHeight, wValue, params.CovenantQuorum)
-	if status != types.BTCDelegationStatus_PENDING {
-		ms.Logger(ctx).Debug("Received covenant signature after the BTC delegation is already expired", "covenant pk", req.Pk.MarshalHex())
-		return &types.MsgAddCovenantSigsResponse{}, nil
+	if status == types.BTCDelegationStatus_UNBONDED {
```
KonradStaniec, line 365: Just note here: I guess this change allow to accept more signatures than the quorum. I imagine slasher or some other tool may relay of there being only quorum of signatures when building witness for sending slashing transactions. This may require some updates there. ( I definitely remember staker, assumes there is only quorum of sigs on Babylon chain)

Lazar955 (approved):
nice work!

x/btcstaking/keeper/msg_server.go
SebastianElvis: Hmm will check when bumping peripheral repositories. In which situation BTC staker will need to rely on this behaviour? As far as I know in other repositories (eg vigilante) it does not even rely on how many covenant signatures are in a BTC delegation, as slashing only uses one of all covenant signatures. Did I miss anything?

x/finality/keeper/msg_server.go
gitferry: Thanks for the waiting. I agree with @SebastianElvis that we can add the duplicate error as expected in the finality provider. I guess this applies in covenant emulator sending duplicate sigs as well. Added issue https://github.com/babylonlabs-io/finality-provider/issues/84 and https://github.com/babylonlabs-io/covenant-emulator/issues/17 to track this

gitferry (approved):
Thanks for the waiting! Great work!

x/btcstaking/keeper/msg_server.go
KonradStaniec: What do you mean by `as slashing only uses one of all covenant signatures` ?

The problem I have in mind is for example here: https://github.com/babylonlabs-io/vigilante/blob/main/btcstaking-tracker/btcslasher/slasher_utils.go#L344. 

We build the wittness from all covenant signatures in the delegation. Up until now it was fine, because we accepted only quorum of signatures in the delegation. 

After this change this is not fine now we can have more that quorum signatures in delegation, but BTC script expects the witness to have exactly quorum of signatures.

PR #131, opened 2024-10-04
New external staking events

KonradStaniec:
Implements events generation as described in https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/31195227/API+asks+for+BBN+node

Few caveats:
- for in delegation creation event we have mulitple finality providers keys to avoid breaking changes for phase-3
- generation of event responsible for expiry happens in `BeginBlocker` so that unbonding early and unbonding through expiry have are retrieved in different way

This pr handles only staking events, finality provider events will be added in separate pr.


proto/babylon/btcstaking/v1/events.proto
SebastianElvis, line 1: Events will go through consensus and occupy block space, so in general it's recommended to keep events as small as possible, and let the consumer (in this context backend) responsible for performing an extra query to get the full data. With big events emitted the chain might have performance issue. I remember this principle is mentioned in https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/31195227/API+asks+for+BBN+node as well

> The events will only contain minimal data such as staking tx hash, the API is responsible to decode the transaction data and get the remaining fields.

Did I miss any context on why the [existing events](https://github.com/babylonlabs-io/babylon/blob/d8862162536acc177feedcd82b74ffc60ed30897/proto/babylon/btcstaking/v1/events.proto#L12-L24) do not suffice to serve the backend?

KonradStaniec: So current events are not enough to drive the api with minimal amount of queries. 

About space, I think this is thread in which came to conclusion that events proposed in the page are non-issue - https://babylonlabsworkspace.slack.com/archives/G07DYV8MA1M/p1728020691866979

x/btcstaking/keeper/btc_delegations.go
```diff
@@ -57,11 +57,28 @@ func (k Keeper) AddBTCDelegation(ctx sdk.Context, btcDel *types.BTCDelegation) e
 		panic(fmt.Errorf("failed to emit EventBTCDelegationStateUpdate for the new pending BTC delegation: %w", err))
 	}
 
+	if err := ctx.EventManager().EmitTypedEvent(types.NewBtcDelCreationEvent(
```
SebastianElvis, line 60: Can L51-58 be removed?

x/btcstaking/keeper/btc_delegations.go
```diff
@@ -95,6 +112,14 @@ func (k Keeper) addCovenantSigsToBTCDelegation(
 
 	k.setBTCDelegation(ctx, btcDel)
 
+	if err := ctx.EventManager().EmitTypedEvent(types.NewCovenantSignatureReceivedEvent(
```
SebastianElvis, line 115: Not 100% sure about this -- do we need an event upon receiving each cov signature or upon receiving a covenant quorum?

x/btcstaking/keeper/btc_delegations.go
```diff
@@ -111,6 +136,15 @@ func (k Keeper) addCovenantSigsToBTCDelegation(
 		activeEvent := types.NewEventPowerDistUpdateWithBTCDel(event)
 		btcTip := k.btclcKeeper.GetTipInfo(ctx)
 		k.addPowerDistUpdateEvent(ctx, btcTip.Height, activeEvent)
+	} else if btcDel.HasCovenantQuorums(params.CovenantQuorum) && !btcDel.HasInclusionProof() {
```
SebastianElvis, line 139: If conditions in L125 and L139 could be consolidated somehow, e.g., 

```go
if btcDel.HasCovenantQuorums(params.CovenantQuorum) {
    if btcDel.HasInclusionProof() {

    } else {
        
    }
}
```

x/btcstaking/keeper/msg_server.go
```diff
@@ -294,6 +294,17 @@ func (ms msgServer) AddBTCDelegationInclusionProof(
 	if err := ctx.EventManager().EmitTypedEvent(event); err != nil {
 		panic(fmt.Errorf("failed to emit EventBTCDelegationStateUpdate for the new active BTC delegation: %w", err))
 	}
+
+	stakingTxHash := btcDel.MustGetStakingTxHash()
+
+	if err := ctx.EventManager().EmitTypedEvent(types.NewInclusionProofEvent(
```
SebastianElvis, line 300: can L290-296 be removed?

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -170,19 +170,28 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
 			} else if delEvent.NewState == types.BTCDelegationStatus_UNBONDED {
-				// add the expired BTC delegation to the map
+				sdkCtx := sdk.UnwrapSDKContext(ctx)
+				// delegation expired and become unbonded emit block event about this
+				// information
+				if !btcDel.IsUnbondedEarly() {
+					if err := sdkCtx.EventManager().EmitTypedEvent(delEvent); err != nil {
+						panic(fmt.Errorf("failed to emit EventBTCDelegationStateUpdate for the new unbonded BTC delegation: %w", err))
+					}
+				}
```
SebastianElvis, line 193: Can we emit this inside `BTCUndelegate`? A slight concern is that this couples emitting events with the main responsibility of this function, which is computing voting power table.

Later on if voting power table is moved to `x/finality` then we could not emit staking events here

CHANGELOG.md
```diff
@@ -37,6 +37,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+### Improvements
+
+* #[131](https://github.com/babylonlabs-io/babylon/pull/131) Add new staking
+events
```
SebastianElvis, line 43: This is state machine breaking as it affects some Merkle root (IIRC the data root) of the block ([ref](https://github.com/tendermint/tendermint/issues/7471))

x/btcstaking/keeper/btc_delegations.go
KonradStaniec: hmm we should emit both events, but it can be simplified by using ` EmitTypedEvents(tevs ...proto.Message)` . Will do that.

SebastianElvis: Why is it the case? in my mind the former event only contains a subset of the latter event and they share the same type of consumers

x/btcstaking/keeper/btc_delegations.go
KonradStaniec: hmm this is probably matter of preference but I try do avoid deeply nested `if/else` blocks.

Though I will experiment with this a bit to check the readability.

x/btcstaking/keeper/power_dist_change.go
SebastianElvis: Hmm I just realised that this is the event of timelock unbonding but not early unbonding :sweat_smile: sorry for the confusion. Then yeah this is the best place to emit this event imo

x/btcstaking/keeper/btc_delegations.go
KonradStaniec: there are at least two advantages:
1.  consumers that do not want to process the "fat" event can only follow the state state updates events
2. consumers know the initial state of created delegation i.e that is is `PENDING`. Without it they would need to either assume some state or we would need to add state field to `EventBTCDelegationCreated`

x/btcstaking/keeper/btc_delegations.go
KonradStaniec: hmm it looks good. I can leave with one level of nesting in this case 😅

x/btcstaking/keeper/btc_delegations.go
SebastianElvis: Hmm I'm not sure if there is a use case for the "slim" event, as the main consumer of this API is backend and other components do not seem to be interested in a newly pending delegation. Also `EventBTCDelegationCreated` seems to imply this BTC delegation is pending, and if not subsequent events will be emitted. That's why I would still prefer to keep each state transition to have a unique type of event.

Another thing is that the `EventBTCDelegationStateUpdate` event now has two use cases, one for voting power update and the other for serving API consumers. This is not a perfect idea as `EventBTCDelegationStateUpdate` is in fact a "DB" object. Now it seems to be a good time to fix this

KonradStaniec: > Another thing is that the EventBTCDelegationStateUpdate event now has two use cases, one for voting power update and the other for serving API consumers. This is not a perfect idea as EventBTCDelegationStateUpdate is in fact a "DB" object. Now it seems to be a good time to fix this

This is good point, we should not expose db objects  for external consumers. Either way, I would not tangle this into this pr which is about new events for api but create separate one for this purpose

Being totally honest I do not have strong preference here and imo both approaches have its merits. @jrwbabylonlab what would api prefer:
1. Some special events, like `delegation_creation`  +  state transition events (i.e one event showing what is current state of delegation)
2. Or set of different events like: `delegation_creation`, `proof_of_inclusion_provided`, `unbonded_early`, `expired` and `covenant_quorum_reached` ?

CHANGELOG.md
SebastianElvis: Hmm have checked the source code of CometBFT, and it seems that `LastResultsHash` is the [hash](https://github.com/cometbft/cometbft/blob/bbb86ced558f05a60c2402283e0cb13d08dc57a2/state/execution.go#L697-L698) of all responses of `DeliverTx`. So if event schmas are changed this will affect the `LastResultsHash`. So this seems to mean it's consensus breaking?

KonradStaniec: hmm this `TxResultsHash` goes to - https://github.com/cometbft/cometbft/blob/v0.38.x/state/store.go#L412
which create some special `types.NewResults` struct, which only cares about 4 fields from whole response: https://github.com/cometbft/cometbft/blob/v0.38.x/types/results.go#L47 (events are ommitted)

gitferry (approved):
Great work!

CHANGELOG.md
SebastianElvis: Right, good find! Events are not deterministic and this won't be consensus breaking then :+1:

x/btcstaking/keeper/btc_delegations.go
KonradStaniec: I would avoid the flags as much as possible. 

We could attach new state to every specific event if this would help the api ?

proto/babylon/btcstaking/v1/events.proto
```diff
@@ -64,3 +64,58 @@ message EventPowerDistUpdate {
...
+message EventCovenantSignatureRecevied{
+  // staking_tx_hash is the hash of the staking identifing the BTC delegation
+  // that this covenant signature is for
+  string staking_tx_hash = 1;
+  // covenant_btc_pk_hex is the hex str of Bitcoin secp256k1 PK of the
+  // covnenat committee that send the signature
+  string covenant_btc_pk_hex = 2;
+  // covenant_unbonding_signature_hex is the hex str of the BIP340 Schnorr
+  // signature of the covenant committee on the unbonding tx
+  string covenant_unbonding_signature_hex = 3;
```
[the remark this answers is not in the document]
KonradStaniec: I imagine api will process those events, retrieve signatures from them and then staker will be able to unbond without consulting Babylon anymore

proto/babylon/btcstaking/v1/events.proto
```diff
@@ -64,3 +64,58 @@ message EventPowerDistUpdate {
...
+  uint32 params_version = 2;
+  // finality_provider_btc_pks_hex is the list of hex str of Bitcoin secp256k1 PK of
+  // the finality providers that this BTC delegation delegates to
+  // the PK follows encoding in BIP-340 spec
+  repeated string finality_provider_btc_pks_hex = 3;
+  // staker_btc_pk_hex is the hex str of Bitcoin secp256k1 PK of the staker that
+  // creates this BTC delegation the PK follows encoding in BIP-340 spec
+  string staker_btc_pk_hex = 4;
+  // timelock is the timelock of the staking tx specified in the BTC script
+  uint32 timelock = 5;
```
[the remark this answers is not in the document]
KonradStaniec: good point lets be consistent then across the code base.

proto/babylon/btcstaking/v1/events.proto
```diff
@@ -64,3 +64,58 @@ message EventPowerDistUpdate {
...
+message EventBTCDelegationInclusionProofReceived {
+  // staking_tx_hash is the hash of the staking tx.
+  // It uniquely identifies a BTC delegation
+  string staking_tx_hash = 1;
+  // start_height is the start BTC height of the BTC delegation
+  // it is the start BTC height of the timelock
+  uint64 start_height = 2;
+  // end_height is the end height of the BTC delegation
+  // it is calculated by end_height = start_height + staking_time
+  uint64 end_height = 3;
```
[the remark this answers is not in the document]
KonradStaniec: probably, but tbh 4bytes is small price to pay and we already have all the information on hand when generating this event.

KonradStaniec:
@SebastianElvis I have applied all the comment so I will merge it now. If there will be some missing things I will improve them in follow up.

PR #132, opened 2024-10-04
chore: Upgrade CosmWasm Parameters 

RafilxTenfen:
- Update cosmwasm params during upgrade
- Mainnet only gov account can store new contracts
- Testnet everybody can store new contracts


app/include_upgrade_mainnet.go
```diff
@@ -13,6 +13,7 @@ func init() {
 	Upgrades = []upgrades.Upgrade{v1.CreateUpgrade(v1.UpgradeDataString{
 		BtcStakingParamStr:    mainnet.BtcStakingParamStr,
 		FinalityParamStr:      mainnet.FinalityParamStr,
+		CosmWasmParamStr:      mainnet.CosmWasmParamStr,
```
SebastianElvis, line 16: Is there a reason why parameters are specified by using strings rather than the parameter structs? These parameters are not like BTC headers that are difficult to be parsed into objects

KonradStaniec: tbh tough call, I like the consistency of current way (everything is a string), but given that we will have separate parameters for mainnet and testnet, maybe it makes sense to use structs for simple cases (like modules parameters)? 

One question is how then will be define those on the fly parameters for devnet ? It will be much tougher to manipulate go struct, that json string.

RafilxTenfen: > It will be much tougher to manipulate go struct, that json string.
yeap, that is true

KonradStaniec: another options is to have structs for mainnet/testnet, and json strings for devnet, but this sounds terrible.

I am leaning to leaving everything as json strings for compatibility/simplicity sake. 

Do you agree @SebastianElvis or do you have something else in mind?

On note @RafilxTenfen, maybe we can leave some comment in upgrade readme or in code, why do we use json string. (i.e easy devnet testing)

RafilxTenfen: I will add

SebastianElvis: I see the rationale here. Json strings seem to be a simpler choice then

SebastianElvis (approved):

KonradStaniec (approved):

Lazar955 (approved):

PR #134, opened 2024-10-06
chore: Remove nolint:errcheck and instead handle errors

vitsalis:
As we are adding static analyzers in the codebase, many of them will point out unhandled errors. If we add annotations to our codebase for each of them, this might bloat the codebase. Another approach is to only allow a single static analyzer fail about unhandled errors, while ignoring the rest.

In this PR, I opted to just handle the errors in a vanilla way to remove the `nolint:errcheck` from the annotations. The way the errors are handled for `types.RegisterQueryHandlerClient` was inspired by [osmosis](https://github.com/osmosis-labs/osmosis/blob/10c8eab5d0d362aed5ad435a0ad2b1e644854b1a/x/mint/module.go#L80). Note though that Osmosis uses `nolint:errcheck` as well for the same [function](https://github.com/osmosis-labs/osmosis/blob/10c8eab5d0d362aed5ad435a0ad2b1e644854b1a/x/txfees/module.go#L92) in other places.


crypto/schnorr-adaptor-signature/keys.go
```diff
@@ -43,7 +43,8 @@ func NewDecyptionKeyFromBytes(decKeyBytes []byte) (*DecryptionKey, error) {
 	}
 
 	var decKeyScalar btcec.ModNScalar
-	decKeyScalar.SetByteSlice(decKeyBytes) //nolint:errcheck
+	// TODO: should we check the return value to identify whether overflow happened?
```
gitferry, line 46: Seems not needed here but it would be good for @SebastianElvis to verify

x/btccheckpoint/module.go
```diff
@@ -80,9 +80,8 @@ func (AppModuleBasic) RegisterRESTRoutes(clientCtx client.Context, rtr *mux.Rout
 
 // RegisterGRPCGatewayRoutes registers the gRPC Gateway routes for the module.
 func (AppModuleBasic) RegisterGRPCGatewayRoutes(clientCtx client.Context, mux *runtime.ServeMux) {
-	err := types.RegisterQueryHandlerClient(context.Background(), mux, types.NewQueryClient(clientCtx))
-	if err != nil {
-		panic(err)
+	if err := types.RegisterQueryHandlerClient(context.Background(), mux, types.NewQueryClient(clientCtx)); err != nil {
```
gitferry, line 83: Seems we can panic here?

gitferry (approved):

crypto/schnorr-adaptor-signature/keys.go
SebastianElvis: Yep this is not needed. If we really want to check, then we can panick upon error (which means something is really wrong e.g., in the btcd lib)

x/btccheckpoint/module.go
vitsalis: thought to keep it consistent with the other ones, this was the only instance that previously panicked. Should we panic on every one of them or just return? Osmosis seems to do the latter, so I followed their approach, but happy to change it if people think the other way is better.

Lazar955 (approved):

crypto/schnorr-adaptor-signature/keys.go
vitsalis: ok removed the todo

RafilxTenfen (commented):
LGTM, missing changelogs

RafilxTenfen (approved):

x/btccheckpoint/module.go
gitferry: Maybe we can panic on every one of them as the errors indicate programming errors

PR #141, opened 2024-10-07
fix handling covent message when quorum is already reached

KonradStaniec:
Pr: https://github.com/babylonlabs-io/babylon/pull/130 allowed for accepting covenant signatures after quorum is reached.

This change introduces regression i.e covenant signatures accepted after quorum is reached also generated voting power events, this could lead to weird results in processing voting power events.

This pr fixes that by making sure that events of any kind are generated only if this is first time quorum is reached.

### DIFF
--- a/x/btcstaking/keeper/btc_delegations.go
+++ b/x/btcstaking/keeper/btc_delegations.go
@@ -94,6 +94,7 @@ func (k Keeper) addCovenantSigsToBTCDelegation(
 	parsedUnbondingSlashingAdaptorSignatures []asig.AdaptorSignature,
 	params *types.Params,
 ) {
+	hadQuorum := btcDel.HasCovenantQuorums(params.CovenantQuorum)
 
 	// All is fine add received signatures to the BTC delegation and BtcUndelegation
 	btcDel.AddCovenantSigs(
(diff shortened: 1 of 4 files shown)


RafilxTenfen (approved):
LGTM :tada:

x/btcstaking/keeper/btc_delegations.go
```diff
@@ -115,7 +116,9 @@ func (k Keeper) addCovenantSigsToBTCDelegation(
 
 	// If reaching the covenant quorum after this msg, the BTC delegation becomes
 	// active. Then, record and emit this event
-	if btcDel.HasCovenantQuorums(params.CovenantQuorum) {
+	// We only emit power distribution events, and external quorum events if it
+	// is the first time the quorum is reached
+	if !hadQuorum && btcDel.HasCovenantQuorums(params.CovenantQuorum) {
```
SebastianElvis, line 121: Oh that's a nice catch :+1: thanks

SebastianElvis (approved):

SebastianElvis:
Btw, changing `CreateCovenantSigs` to iterate over all covenant members (as opposed to a quorum number of covenant members) can reveal this bug. Probably it's worth doing as we no longer expect Babylon to handle only a quorum number of covenant signatures?

KonradStaniec:
> Btw, changing CreateCovenantSigs to iterate over all covenant members (as opposed to a quorum number of covenant members) can reveal this bug. Probably it's worth doing as we no longer expect Babylon to handle only a quorum number of covenant signatures?

good point 👍  added in last commit

Issue #72, opened 2024-09-16
TotalVotingPower is not accurate in FinalityProviderDistInfo


gitferry:
`TotalVotingPower` is not accurate in `FinalityProviderDistInfo` anymore because in cases where the finality providers do not have timestamped pub rand or they are jailed, they will be assigned zero voting power but can continue to accept delegations. Therefore, `TotalVotingPower` should be renamed to `TotalDelegationValue` or something.

[context](https://github.com/babylonlabs-io/babylon/pull/65#discussion_r1760451612)

PR #144, opened 2024-10-08
feat: Add new external events related to finality provider status change

gitferry:
This is a follow-up PR of #131. This PR implements events generation related to finality providers as described in https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/31195227/API+asks+for+BBN+node. In particular, this PR

1. adds events for fp creation and editing, which are emitted within corresponding message handlers
2. adds events for fp status update, i.e., active, inactive, jailed, slashed, which are emitted during voting power table update per BeginBlock


x/btcstaking/keeper/power_dist_change.go
```diff
@@ -238,6 +261,15 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
 		// assigning delegation to it
 		if _, ok := slashedFPs[fpBTCPKHex]; ok {
+			fp.IsSlashed = true
+
+			statusChangeEvent := types.NewFinalityProviderStatusChangeEvent(fp.BtcPk, types.FinalityProviderStatus_STATUS_SLASHED)
+			if err := sdkCtx.EventManager().EmitTypedEvent(statusChangeEvent); err != nil {
+				panic(fmt.Errorf(
+					"failed to emit FinalityProviderStatusChangeEvent with status %s: %w",
+					types.FinalityProviderStatus_STATUS_SLASHED.String(), err))
+			}
```
SebastianElvis, line 271: Can this be done in the message handler of finality signatures? Doing it here might add more repsonsibilities to this function, which was supposed to focus on computing the voting power table

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -246,6 +278,13 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 		// but won't be assigned with voting power
 		if _, ok := jailedFPs[fpBTCPKHex]; ok {
 			fp.IsJailed = true
+
+			statusChangeEvent := types.NewFinalityProviderStatusChangeEvent(fp.BtcPk, types.FinalityProviderStatus_STATUS_JAILED)
+			if err := sdkCtx.EventManager().EmitTypedEvent(statusChangeEvent); err != nil {
+				panic(fmt.Errorf(
+					"failed to emit FinalityProviderStatusChangeEvent with status %s: %w",
+					types.FinalityProviderStatus_STATUS_JAILED.String(), err))
+			}
```
SebastianElvis, line 287: Can this be done in `HandleLiveness`?

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -106,21 +107,43 @@ func (k Keeper) recordVotingPowerAndCache(ctx context.Context, prevDc, newDc *ty
...
+
+		statusChangeEvent := types.NewFinalityProviderStatusChangeEvent(fp.BtcPk, types.FinalityProviderStatus_STATUS_ACTIVE)
+		if err := sdkCtx.EventManager().EmitTypedEvent(statusChangeEvent); err != nil {
+			panic(fmt.Errorf(
+				"failed to emit FinalityProviderStatusChangeEvent with status %s: %w",
+				types.FinalityProviderStatus_STATUS_ACTIVE.String(), err))
+		}
+
+		k.Logger(sdkCtx).Info("a new finality provider becomes active", "pk", fp.BtcPk.MarshalHex())
+	}
```
SebastianElvis, line 129: What do you think about moving the logic of emitting events out of `recordVotingPowerCache` as a separate function, and call this function in `UpdatePowerDist`? Suggesting this as `recordVotingPowerCache` is already a bit convoluted and has its own responsibility

proto/babylon/btcstaking/v1/incentive.proto
```diff
@@ -38,10 +38,14 @@ message FinalityProviderDistInfo {
...
     // is_timestamped indicates whether the finality provider
     // has timestamped public randomness committed
+    // if no, it should not be assigned voting power
     bool is_timestamped = 6;
     // is_jailed indicates whether the finality provider
     // is jailed, if so, it should not be assigned voting power
     bool is_jailed = 7;
+    // is_slashed indicates whether the finality provider
+    // is slashed, if so, it should not be assigned voting power
+    bool is_slashed = 8;
```
SebastianElvis, line 48: Imo this adds some coupling between the external services and the consensus protocol itself. If possible, we could emit slashing events upon slashing a FP in `MsgAddFinalitySig`. Wdyt?

SebastianElvis (dismissed):

x/btcstaking/keeper/power_dist_change.go
gitferry: We already have delegation status change event emitted here https://github.com/babylonlabs-io/babylon/blob/c33f66f0a235c759db907ac16f8f8cc1466129cb/x/btcstaking/keeper/power_dist_change.go#L213. I thought it would be good to put these types of events in one place

x/btcstaking/keeper/power_dist_change.go
gitferry: If it's outside of `recordVotingPowerCache`, we would need to do an extra iteration of getting active/inactive fps and getting new active/inactive ones, which slightly adds overhead. Do we want to do that?

x/btcstaking/keeper/power_dist_change.go
KonradStaniec: I think I agree with this i.e in case of `FinalityProviderStatusChangeEvent` the way to go is to keep it as on type of even i.e if we can generate all state transitions in transaction context we should do it. If we can't , then it must be a block event.  

Having that in mind, imo we need to generate all `FinalityProviderStatusChangeEvent` in block context as some state transitions do not happen due to transactions like for example, becoming jailed.

gitferry: Yep, imo `slashing` and status change are different types of events that may interest to different types of subscribers. For subscribers (this PR targets on) that are interested in status change, they should be able to get all the status change event within `block_events`

x/btcstaking/keeper/power_dist_change.go
KonradStaniec: tbh I agree with Runchao that `recordVotingPowerCache` is already convoluted enough, but on the other hand I would not add additional extra iteration just for events sake. 

I am slightly leadning to:
1. merging this pr at is is to not block the api work and have all events in next version `v0.12.0`
2. schedule refactoring of this `power_dist_change.go` file and mechanism of updates for the next devnet3. It would also be great to tackle this `todo` while doing it -  https://github.com/babylonlabs-io/babylon/blob/v0.11.0/x/btcstaking/keeper/power_dist_change.go#L205

what do you guys think ?

gitferry: I also agree that we should separate `recordVotingPowerCache` into at least two functions: (1) transforming delegations into voting power with consideration of timestamping, jailing, etc. (2) emitting events and saving the cache.
Another refactoring point to consider: https://github.com/babylonlabs-io/babylon/issues/72

KonradStaniec: >Another refactoring point to consider: https://github.com/babylonlabs-io/babylon/issues/72

uh this seems important 👍

CHANGELOG.md
```diff
@@ -48,7 +48,8 @@ messages inside `authz.MsgExec`
 
 ### Improvements
 
-* #[131](https://github.com/babylonlabs-io/babylon/pull/131) Add new staking
+* [#131](https://github.com/babylonlabs-io/babylon/pull/131) and
+  [#144](https://github.com/babylonlabs-io/babylon/pull/144) Add new staking
```
KonradStaniec, line 52: I would separte change log entries for:
- new staking events
- new finality provider events

KonradStaniec (approved):

x/btcstaking/keeper/power_dist_change.go
KonradStaniec: task created to track it https://github.com/babylonlabs-io/babylon/issues/145 (scheduled for devnet3)

SebastianElvis (approved):
Thanks for the discussion :+1:

PR #154, opened 2024-10-09
fix(btcstaking/client): edit fp arg

Lazar955:
We're accessing arg at index 1, but should be at 0

### DIFF
--- a/x/btcstaking/client/cli/tx.go
+++ b/x/btcstaking/client/cli/tx.go
@@ -159,7 +159,7 @@ func NewEditFinalityProviderCmd() *cobra.Command {
 			}
 
 			// get BTC PK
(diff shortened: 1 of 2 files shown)


x/btcstaking/client/cli/tx.go
```diff
@@ -159,7 +159,7 @@ func NewEditFinalityProviderCmd() *cobra.Command {
 			}
 
 			// get BTC PK
-			btcPK, err := hex.DecodeString(args[1])
+			btcPK, err := hex.DecodeString(args[0])
```
RafilxTenfen, line 162: good catch!

RafilxTenfen (approved):
miss changelogs ^^

SebastianElvis (approved):

gitferry (approved):

PR #170, opened 2024-10-11
ci: add goreleaser setup and move out changelog reminder

SebastianElvis:
This PR 

- sets up Go releaser for Babylon. Currently it only supports linux amd64. Other targets will be supported in the future. To test locally, run `make release-snapshot`
- moves changelog reminder out from the CI flow

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,94 @@
+project_name: babylon
+
+builds:
+  - id: babylond-darwin-amd64
+    main: ./cmd/babylond/main.go
+    binary: babylond
+    hooks:
+      pre:
+        - wget https://github.com/CosmWasm/wasmvm/releases/download/{{ .Env.COSMWASM_VERSION }}/libwasmvmstatic_darwin.a -O /lib/libwasmvmstatic_darwin.a
+    env:
+      - CC=o64-clang
+      - CGO_LDFLAGS=-L/lib
+    goos:
+      - darwin
+    goarch:
+      - amd64
+    flags:
+      - -mod=readonly
+      - -trimpath
+    ldflags:
+      - -X github.com/cosmos/cosmos-sdk/version.Name=osmosis
+      - -X github.com/cosmos/cosmos-sdk/version.AppName=babylond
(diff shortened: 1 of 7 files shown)


SebastianElvis:
Hmm any idea why changelog reminder is not happy with the error `not a PR` :sweat_smile:

KonradStaniec (approved):
LGTM!

Maybe we could update `RELEASE_PROCESS.md`  file to indicate when binaries will be build ? I imagine it would be when tagging release branch ?

SebastianElvis:
> LGTM!
> 
> Maybe we could update `RELEASE_PROCESS.md` file to indicate when binaries will be build ? I imagine it would be when tagging release branch ?

Yes this is already [mentioned](https://github.com/babylonlabs-io/babylon/blob/88e92469d75001c6cf73a3042e32693ba0370fb2/RELEASE_PROCESS.md#L120-L127). And you are right this is done upon tagging

KonradStaniec:
>Yes this is already [mentioned](https://github.com/babylonlabs-io/babylon/blob/88e92469d75001c6cf73a3042e32693ba0370fb2/RELEASE_PROCESS.md#L120-L127).

oh I do not see any mentions there, that upon pushing a tag new binaries will be build and published. There is only mention abut backporitng label

SebastianElvis:
> oh I do not see any mentions there, that upon pushing a tag new binaries will be build and published. There is only mention abut backporitng label

Ah you are right! Added

.goreleaser.yml
```diff
@@ -0,0 +1,62 @@
...
+    env:
+      - GO111MODULE=on
+    flags:
+      - -mod=readonly
+      - -trimpath
+    ldflags:
+      - -X github.com/cosmos/cosmos-sdk/version.Name=babylon
+      - -X github.com/cosmos/cosmos-sdk/version.AppName=babylond
+      - -X github.com/cosmos/cosmos-sdk/version.Version={{ .Version }}
+      - -X github.com/cosmos/cosmos-sdk/version.Commit={{ .Commit }}
```
Lazar955, line 23: 👏 💯

Lazar955 (approved):
Nice work

PR #180, opened 2024-10-11
fix: Non-determinism in sorting finality providers in the voting power table

gitferry:
Non-determinism is caused due to not considering cases where both two items are jailed or non-timestamped, and when voting power is equal. This PR addressed this by determining the order in those cases by the increasing order of btc pk hex.


Lazar955 (approved):
Nice work!

KonradStaniec (approved):

PR #181, opened 2024-10-11
chore: Bitcoin heights and depths made uint32

### LINKED ISSUE #179: Use uint32 for BTC header heights and depths
The `uint32` type is enough to cover both BTC heights and BTC chain depth values. Note that `int32` is used by popular libraries such as lightning and btcd.

vitsalis:
Resolves #179 

This is an impactful change that brings the BTC height types that we use more in sync with the expected types for BTC block heights, similar to those that are used by other libraries. It is part of the code improvement efforts that we are doing in preparation of the next testnet.

Given that it touches many files, I believe that this change should happen sooner than later so that it can be appropriately tested.


CHANGELOG.md
```diff
@@ -44,6 +44,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Improvements
 
 * [#148](https://github.com/babylonlabs-io/babylon/pull/148) Add block results query
+* [#181](https://github.com/babylonlabs-io/babylon/pull/181) Modify BTC heights
```
KonradStaniec, line 47: This should be in `### State Machine Breaking`

KonradStaniec (approved):
looks good 👍

Lazar955 (approved):

gitferry (approved):
LGTM! Do we want to make the same change to auxiliary programs?

vitsalis:
> LGTM! Do we want to make the same change to auxiliary programs?

yep, I expect this can be handled when auxiliary programs are updated to the latest Babylon.

PR #183, opened 2024-10-11
backport: Fix non-determinism in sorting finality providers in the voting power table (#180)

gitferry:
This is a manual backport as the automatic process failed due to conflicts in changelog


KonradStaniec:
hmm if the jailing already happened on devnet2, this will require devnet2 restart (cc: @filippos47)

KonradStaniec (approved):

Issue #174, opened 2024-10-11
Passing nil proof in MsgAddFinality vote leads to Panic


KonradStaniec:
Context: https://github.com/babylonchain/babylon/issues/668

PR #186, opened 2024-10-11
Do not panic on nil proof when handling votes

KonradStaniec:
If not would receive finality vote with `nil` Proof, it would panic. 

Pr:
- fixes that
- add test case 

Fixes: https://github.com/babylonlabs-io/babylon/issues/174

### DIFF
--- a/x/finality/types/msg.go
+++ b/x/finality/types/msg.go
@@ -4,6 +4,7 @@ import (
 	fmt "fmt"
 
 	"github.com/babylonlabs-io/babylon/crypto/eots"
+	bbn "github.com/babylonlabs-io/babylon/types"
 	"github.com/cometbft/cometbft/crypto/merkle"
 	"github.com/cometbft/cometbft/crypto/tmhash"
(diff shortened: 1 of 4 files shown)


RafilxTenfen (approved):
:rocket:

Lazar955 (approved):
lgtm

x/finality/types/msg.go
```diff
@@ -20,6 +20,30 @@ func (m *MsgAddFinalitySig) MsgToSign() []byte {
 	return msgToSignForVote(m.BlockHeight, m.BlockAppHash)
 }
 
+func (m *MsgAddFinalitySig) ValidateBasic() error {
```
SebastianElvis, line 23: We could check lengths of `fpBtcPk`, `PubRand`, and `FinalitySig`

SebastianElvis (approved):

x/finality/types/msg.go
KonradStaniec: good point! 👍

PR #193, opened 2024-10-14
fix: fixing witness construction of slashing tx

SebastianElvis:
Also adapts all tests to ensure the witness construction works even if all covenant members provide signatures.

### DIFF
--- a/x/btcstaking/types/btc_slashing_tx.go
+++ b/x/btcstaking/types/btc_slashing_tx.go
@@ -274,6 +274,7 @@ func (tx *BTCSlashingTx) BuildSlashingTxWithWitness(
 	outputIdx uint32,
 	delegatorSig *bbn.BIP340Signature,
 	covenantSigs []*asig.AdaptorSignature,
+	covenantQuorum uint32,
 	slashingPathSpendInfo *btcstaking.SpendInfo,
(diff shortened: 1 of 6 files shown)


x/btcstaking/types/btc_slashing_tx.go
```diff
@@ -274,6 +274,7 @@ func (tx *BTCSlashingTx) BuildSlashingTxWithWitness(
 	outputIdx uint32,
 	delegatorSig *bbn.BIP340Signature,
 	covenantSigs []*asig.AdaptorSignature,
+	covenantQuorum uint32,
```
KonradStaniec, line 277: maybe we can return error early from this function i.e:
```
if len(covenantSigs) < covenantQuorum {
  return nil, fmt.Errorf("not enough signatures")
}
```

SebastianElvis: Good point! As `covenantSigs` can contain `nil` (in order to have the length of covenant committee size), I added the below logic below L303

```go
	// ensure the number of covenant signatures is at least the quorum number
	if numSigs < covenantQuorum {
		return nil, fmt.Errorf("not enough covenant signatures to reach quorum")
	}
```

does this make sense?

KonradStaniec (approved):

PR #194, opened 2024-10-14
chore: Handle `FinalityProviderSigningInfo` object

samricotta:
## Summary

Currently some of the objects in the babylon apis are exposed and need to be handled correctly seen in [here](https://github.com/babylonlabs-io/pm/issues/72)

This PR handles the `FinalityProviderSigningInfo` in https://github.com/babylonlabs-io/babylon/blob/v0.11.0/x/finality/keeper/grpc_query.go#L230

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
-### Security
+### API Breaking
 
 * [#194](https://github.com/babylonlabs-io/babylon/pull/194) Adjusted handling of `FinalityProviderSigningInfo` in finality keeper queries to improve API security
   * Modified `QuerySigningInfosResponse` to remove direct exposure of sensitive fields
--- a/proto/babylon/finality/v1/query.proto
+++ b/proto/babylon/finality/v1/query.proto
@@ -234,9 +234,14 @@ message QuerySigningInfoRequest {
 // SigningInfoResponse defines the API response containing a finality provider's signing info
 // for monitoring their liveness activity.
 message SigningInfoResponse {
+  // fp_btc_pk is the BTC PK of the finality provider that casts this vote
   string fp_btc_pk_hex = 1;
+  // start_height is the block height at which finality provider become active
   int64 start_height = 2;
+  // missed_blocks_counter defines a counter to avoid unnecessary array reads.
+  // Note that `Sum(MissedBlocksBitArray)` always equals `MissedBlocksCounter`.
   int64 missed_blocks_counter = 3;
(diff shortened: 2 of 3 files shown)


x/finality/keeper/grpc_query.go
```diff
@@ -271,5 +271,22 @@ func (k Keeper) SigningInfos(ctx context.Context, req *types.QuerySigningInfosRe
 	if err != nil {
 		return nil, err
 	}
-	return &types.QuerySigningInfosResponse{FpSigningInfos: signInfos, Pagination: pageRes}, nil
+	return &types.QuerySigningInfosResponse{SigningInfos: k.convertToSigningInfosResponse(signInfos), Pagination: pageRes}, nil
+}
+
+func (k Keeper) convertToSigningInfoResponse(info types.FinalityProviderSigningInfo) types.SigningInfoResponse {
```
RafilxTenfen, line 277: suggestion to not attach functions to keeper if they do not use anything from the keeper 
```suggestion
func convertToSigningInfoResponse(info types.FinalityProviderSigningInfo) types.SigningInfoResponse {
```

RafilxTenfen, line 277: suggestion to not attach functions to keeper if they do not use anything from the keeper 
```suggestion
func convertToSigningInfoResponse(info types.FinalityProviderSigningInfo) types.SigningInfoResponse {
```

RafilxTenfen (commented):
LGTM, only one suggestion

RafilxTenfen (approved):
LGTM, only one suggestion

CHANGELOG.md
```diff
@@ -37,6 +37,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+### Security
```
KonradStaniec, line 40: I think I would put this in `API Breaking` category

proto/babylon/finality/v1/query.proto
```diff
@@ -229,11 +231,19 @@ message QuerySigningInfoRequest {
   string fp_btc_pk_hex = 1;
 }
 
+// SigningInfoResponse defines the API response containing a finality provider's signing info
+// for monitoring their liveness activity.
+message SigningInfoResponse {
+  string fp_btc_pk_hex = 1;
```
KonradStaniec, line 237: every field need to have a comment otherwise `make proto-lint` won't pass (it was part of our CI but somewhere during migration we lost it 😅 )

x/finality/keeper/grpc_query.go
KonradStaniec: agree with @RafilxTenfen on that lets make this normal function

KonradStaniec (approved):
Some minor comments in general lgtm!

PR #197, opened 2024-10-15
fix: `BTCDelgationResponse` missing `staking_time`

gitferry:
In pre-approval flow, the delegation has zero value for both `startHeight` and `endHeight`. Therefore, the staking time should not be inferred from `endHeight - startHeight`. Instead, it should be stored seprately. This PR adds `stakingTime` to `BTCDelegationResponse`


KonradStaniec (approved):

RafilxTenfen:
This will need a new release at btc-staker, because it broke the proto for query `QueryBTCDelegationResponse`

gitferry:
> This will need a new release at btc-staker, because it broke the proto for query QueryBTCDelegationResponse

Yep, this will go to the next devnet release

proto/babylon/btcstaking/v1/query.proto
```diff
@@ -251,40 +251,42 @@ message BTCDelegationResponse {
   // fp_btc_pk_list is the list of BIP-340 PKs of the finality providers that
   // this BTC delegation delegates to
   repeated bytes fp_btc_pk_list = 3 [ (gogoproto.customtype) = "github.com/babylonlabs-io/babylon/types.BIP340PubKey" ];
+  // staking_time is the number of blocks for which the delegation is locked on BTC chain
+  uint32 staking_time = 4;
```
maurolacy, line 255: Nit: To try to be more backwards compatible, you can use a new label here, i.e.:
`uint32 staking_time = 17;`

That way the older versions will be able to handle all the other fields.

PR #203, opened 2024-10-15
chore: update exposed db object `RewardGuages` in `x/incentive` module

samricotta:

Currently some of the objects in the babylon apis are exposed and need to be handled correctly seen in https://github.com/babylonlabs-io/pm/issues/72

This PR handles the exposed objects in `x/incentive`

Specifically the `RewardGuages` object.


RafilxTenfen (approved):
LGTM :rocket:

gitferry (approved):

proto/babylon/incentive/query.proto
```diff
@@ -44,11 +44,26 @@ message QueryRewardGaugesRequest {
...
+    // Can have multiple coin denoms
+    repeated cosmos.base.v1beta1.Coin coins = 1 [
+        (gogoproto.nullable) = false,
+        (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+    ];
+    // withdrawn_coins are coins that have been withdrawn by the stakeholder already
+    repeated cosmos.base.v1beta1.Coin withdrawn_coins = 2 [
+        (gogoproto.nullable) = false,
+        (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+    ];
```
RafilxTenfen, line 59: This Response type is exactly like the `RewardGauge` should we have two exactly structures just to split from the db used structure? @babylonlabs-io/core-dev

SebastianElvis: I think this is the case, as the two structs might diverge in the future. If we reuse the same struct now, then it's made difficult if we want to make modifications on API or DB.

RafilxTenfen (approved):
LGTM :100:

gitferry (approved):
Great work!

PR #204, opened 2024-10-15
chore: add check in finality for block height finality parameter to accept finality vote/pubrand

RafilxTenfen:
Add finality param to minimum block height to allow receiving Finality and Pub Rand Commit of `x/finality` 

This value is a module parameter and it can be updated by governance proposal


x/btcstaking/keeper/keeper.go
```diff
@@ -82,10 +82,38 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 // the voting power distribution cache used for computing voting power table
 // and distributing rewards once the block is finalised by finality providers.
 func (k Keeper) BeginBlocker(ctx context.Context) error {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	activationHeight := k.GetActivationBlockHeightVotingPower()
+	currHeight := sdkCtx.HeaderInfo().Height
+	if currHeight < activationHeight {
+		// TODO: remove it after Phase-2 launch
```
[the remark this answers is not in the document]
RafilxTenfen: in a future upgrade we would be able to remove it

x/btcstaking/keeper/keeper.go
```diff
@@ -82,10 +82,38 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
+func (k Keeper) GetActivationBlockHeightVotingPower() int64 {
+	switch k.btcNet.Name {
+	case chaincfg.MainNetParams.Name:
+		return 300
+	case chaincfg.TestNet3Params.Name:
+		return 50
+	case chaincfg.RegressionNetParams.Name:
+		return 40
+	default:
+		return 0
```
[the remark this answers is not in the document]
RafilxTenfen: by default I was thinking to continue activating at height 0 to not affect any tests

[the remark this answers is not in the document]
RafilxTenfen: by default I was thinking to continue activating at height 0 to not affect any tests

x/btcstaking/keeper/keeper.go
```diff
@@ -82,10 +82,38 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
+// voting power distribution events.
+func (k Keeper) GetActivationBlockHeightVotingPower() int64 {
+	switch k.btcNet.Name {
+	case chaincfg.MainNetParams.Name:
+		return 300
+	case chaincfg.TestNet3Params.Name:
+		return 50
+	case chaincfg.RegressionNetParams.Name:
+		return 40
+	default:
```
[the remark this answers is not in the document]
RafilxTenfen: Add signet instead of testnet

x/btcstaking/keeper/keeper.go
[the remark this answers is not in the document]
RafilxTenfen: add disclaimer that should be removed in a future coordinated upgrade

x/btcstaking/keeper/keeper.go
```diff
@@ -82,10 +82,38 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
 	return nil
 }
+
+// GetActivationBlockHeightVotingPower returns the minimum block
+// height that the babylon should have to start processing
+// voting power distribution events.
+func (k Keeper) GetActivationBlockHeightVotingPower() int64 {
+	switch k.btcNet.Name {
+	case chaincfg.MainNetParams.Name:
+		return 300
```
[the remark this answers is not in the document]
RafilxTenfen: add a quick math to mainnet block activation at 17500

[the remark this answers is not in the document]
RafilxTenfen: Add a comment that considering a 10s block time and that the upgrade will happen at block 220, the activation height should be 17500 if we want 48Hrs for FPs to register

RafilxTenfen:
Will be modified to block finality vote and pub rand commit before activation height

types/utils.go
```diff
@@ -29,3 +31,34 @@ func CheckForDuplicatesAndEmptyStrings(input []string) error {
 
 	return nil
 }
+
+// GetActivationHeight returns the minimum block height from which the BTC
+// staking protocol starts updating the voting power table.
```
gitferry, line 36:
```suggestion
// GetActivationHeight returns the minimum block height from which the BTC
// staking protocol starts processing finality messages
```

types/utils.go
```diff
@@ -29,3 +31,34 @@ func CheckForDuplicatesAndEmptyStrings(input []string) error {
...
+	// with the test that will be done and the block time
+	// that the validator sets in config.
+	case chaincfg.MainNetParams.Name:
+		// For mainnet considering we want 48 hours
+		// at a block time of 10s that would be 17280 blocks
+		// considering the upgrade for Phase-2 will happen at block
+		// 220, the mainnet activation height for btcstaking should
+		// be 17280 + 220.
+		return 17500
+	case chaincfg.SigNetParams.Name:
```
gitferry, line 49: In testnet, we would use Signet for btc but to simulate the exact behavior as the mainnet. What activation height do we want for that?

CHANGELOG.md
```diff
@@ -58,6 +58,8 @@ to total bonded sat
 
 * [#181](https://github.com/babylonlabs-io/babylon/pull/181) Modify BTC heights
   and depths to be of uint32 type instead of uint64.
+* [#204](https://github.com/babylonlabs-io/babylon/pull/204) Add minimum babylon
+  block height to start process BTC staking voting power distribution events.
```
gitferry, line 62:
```suggestion
* [#204](https://github.com/babylonlabs-io/babylon/pull/204) Add minimum babylon
  block height to start processing finality messages.
```

gitferry (commented):
I was thinking it might be better to make the activation height a parameter other than hardcoding. Major benefit is that we don't need to care about what btc network we are running but just parameterize it in prepare-genesis. Wdyt?

RafilxTenfen:
> I was thinking it might be better to make the activation height a parameter other than hardcoding. Major benefit is that we don't need to care about what btc network we are running but just parameterize it in prepare-genesis. Wdyt?

We could add it as a parameter, but after the activation height is reached it could be removed and if we have in proto it would be a deprecated value forever there until we have a new proto version which is pretty difficult 
So, in my mind we have a few possibilities to change the activation height based on the BTC network

types/utils.go
RafilxTenfen: To simulate the exact behavior would need to be the same height as mainnet 17500

KonradStaniec (commented):
In finality endblocker there is this bit:
https://github.com/babylonlabs-io/babylon/blob/v0.13.0/x/finality/abci.go#L25

I wonder whether we should somehow tie this activation condition with new parameter ?

proto/babylon/finality/v1/params.proto
```diff
@@ -30,4 +30,7 @@ message Params {
   // jail_duration is the minimum period of time that a finality provider remains jailed
   google.protobuf.Duration jail_duration = 5
   [(gogoproto.nullable) = false, (amino.dont_omitempty) = true, (gogoproto.stdduration) = true];
+  // activation_block_height is the babylon block which the finality module will start to
+  // accept public randomness commit and finality voting.
+  uint64 activation_block_height = 6;
```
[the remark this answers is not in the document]
SebastianElvis: Hmm wouldn't this be better suited as a constant rather than a parameter? Since

- This value will only be used until a given height
- It might be possible that a gov prop to update this value to a very big one, to disable finality. This sounds like an attack vector to me.

x/finality/keeper/msg_server.go
```diff
@@ -303,3 +309,39 @@ func (k Keeper) slashFinalityProvider(ctx context.Context, fpBtcPk *bbn.BIP340Pu
...
+			"finality block height: %d is lower than activation height %d",
+			msg.BlockHeight, activationHeight,
+		)
+	}
+	return nil
+}
+
+// validateActivationHeightCommitPubRand returns error if the msg commit pub rand list
+// start height is lower than the activation height
+func (ms msgServer) validateActivationHeightCommitPubRand(ctx sdk.Context, msg *types.MsgCommitPubRandList) error {
```
SebastianElvis, line 333: If we disable pub rand commit before the finality activation height, wouldn't it be possible that upon the finality actviation height, none of the FPs will have timestamped pub rand, such that no FP will be voting. Only after ~10 hours some FPs start having timestamped pub rand and will start voting.

Wdyt about allowing FPs to commit pub rand anytime, and just disabling finality signature until fiunality activation height?

proto/babylon/finality/v1/params.proto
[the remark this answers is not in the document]
RafilxTenfen: There was a conversation during the friday meeting that this value should be configurable that might change depending on the chain finality registration and amount of btc transitioned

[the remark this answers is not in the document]
RafilxTenfen: @KonradStaniec yes, we could add to the upgrade handler, mainnet I will set to 48hrs after upgrade ~17500 blocks and for testnet set to a low number to not hold the e2e test too much, also will add a check for it in the e2e testing

Thanks for the remainder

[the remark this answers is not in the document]
KonradStaniec: on last cross-team meeting (btw. do you have invite to this one ? it is the same time as core meeing but on Friday) we decided to make this a parameter as it may happen that we won't gather enough stake before the activation height we hard code, which would be critical scenario. When it is parameter, it would be just a question of gov proposal to postpone finality activation a bit.

> It might be possible that a gov prop to update this value to a very big one, to disable finality. This sounds like an attack vector to me.

wouldn't that require >66% of the native stake being malicious ? i.e if that happens we have other problems 😅 and at the beginning of the Babylon chain, it will be Babylon foundation who controls most stake.

cmd/babylond/cmd/flags.go
```diff
@@ -19,76 +19,78 @@ import (
...
+	flagMaxStakingTimeBlocks          = "max-staking-time-blocks"
+	flagMaxActiveFinalityProviders    = "max-active-finality-providers"
+	flagMinUnbondingTime              = "min-unbonding-time"
+	flagUnbondingFeeSat               = "unbonding-fee-sat"
+	flagSlashingPkScript              = "slashing-pk-script"
+	flagMinSlashingFee                = "min-slashing-fee-sat"
+	flagSlashingRate                  = "slashing-rate"
+	flagMinCommissionRate             = "min-commission-rate"
+	flagSignedBlocksWindow            = "signed-blocks-window"
+	flagFinalityActivationBlockHeight = "finality-activation-block-height"
```
gitferry, line 52:
```suggestion
	flagActivationHeight = "activation-height"
```

try to be as much succinct as possible for parameters

x/finality/keeper/keeper.go
```diff
@@ -95,3 +95,9 @@ func (k Keeper) GetCurrentEpoch(ctx context.Context) uint64 {
 
 	return currentEpoch.EpochNumber
 }
+
+// GetActivationHeight returns the activation height based
+// on the btc network config.
+func (k Keeper) GetActivationHeight(ctx sdk.Context) uint64 {
+	return k.GetParams(ctx).ActivationBlockHeight
+}
```
gitferry, line 103: Not sure this is still needed as this can always be accessed via `k.GetParams(ctx).ActivationBlockHeight`

x/finality/keeper/msg_server.go
```diff
@@ -303,3 +309,39 @@ func (k Keeper) slashFinalityProvider(ctx context.Context, fpBtcPk *bbn.BIP340Pu
 		panic(fmt.Errorf("failed to emit EventSlashedFinalityProvider event: %w", err))
 	}
 }
+
+// validateActivationHeightAddFinalitySig returns error if the msg add finality
+// block height is lower than the activation height
+func (ms msgServer) validateActivationHeightAddFinalitySig(ctx sdk.Context, msg *types.MsgAddFinalitySig) error {
```
gitferry, line 315:
```suggestion
func (ms msgServer) validateActivationHeight(ctx sdk.Context, height uint32) error {
```
think maybe we could make it general by just passing in the height. logs can be wrapped by the called for more context

x/finality/keeper/msg_server.go
gitferry: I think the point here is that fps won't waste any pub rand as the first one is not lower than the activation height. Fps can adjust the logic for the first commit starting from higher than the activation height but still can submit the commit early

x/finality/keeper/msg_server.go
```diff
@@ -53,9 +53,12 @@ func (ms msgServer) AddFinalitySig(goCtx context.Context, req *types.MsgAddFinal
 		return nil, err
 	}
 
-	fpPK := req.FpBtcPk
-
 	ctx := sdk.UnwrapSDKContext(goCtx)
+	if err := ms.validateActivationHeightAddFinalitySig(ctx, req); err != nil {
```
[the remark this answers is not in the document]
gitferry: I think we should return here https://github.com/babylonlabs-io/babylon/blob/dd108f4013d9de0e684336134f49c14bd8eac2d1/x/btcstaking/keeper/power_dist_change.go#L52 if the finality is not activated. This way, voting power is not assigned after delegation events are processed so that we won't start tracking voting history of a newly active fp

x/finality/keeper/msg_server.go
RafilxTenfen: finality providers can commit the pub rand for the `start height = finality activation height` at any time and have it timestamped 

> Wdyt about allowing FPs to commit pub rand anytime, and just disabling finality signature until fiunality activation height?

they would commit some pub rand that is not going to be used, not a big problem to be honest 
What do you all think? @gitferry @KonradStaniec

RafilxTenfen: just saw @gitferry opinion https://github.com/babylonlabs-io/babylon/pull/204#discussion_r1808533906

KonradStaniec: tbh I would keep it as it is i.e:
- fp can commit randomness at any time (i.e send transaction with commitment any time)
- fp can commit randomness only starting from activation height (i.e lowest commit must have at least start height >= activation height)
so this is what also @gitferry mentioned

SebastianElvis: I see, the logic here is that the start height should be at least activation height, rather than that FP can commit pub rand only after the activation height. That makes sense :+1: sorry for the confusion

proto/babylon/finality/v1/params.proto
SebastianElvis, line 35:
```suggestion
  uint64 finality_activation_height = 6;
```

This is because the BTC staking activation height will be moved to the Finality module ([ref](https://github.com/babylonlabs-io/babylon/pull/216)), the BTC staking module will fall back to a storage for BTC delegations and FPs (for Babylon and all consumer chains). Better to distinguish between BTC staking activation v.s. finality activation. Same comments for other "activation" usage

SebastianElvis (approved):
Pretty solid work :clap:

test/e2e/configurer/chain/commands_btcstaking.go
```diff
@@ -188,7 +196,15 @@ func (n *NodeConfig) CommitPubRandList(fpBTCPK *bbn.BIP340PubKey, startHeight ui
 	// gas
 	cmd = append(cmd, "--gas=500000")
 
-	_, _, err := n.containerManager.ExecTxCmd(n.t, n.chainId, n.Name, cmd)
+	allTxArgs := []string{fmt.Sprintf("--chain-id=%s", n.chainId), "--gas-prices=0.002ubbn", "-b=sync", "--yes", "--keyring-backend=test", "--log_format=json", "--home=/home/babylon/babylondata"}
+	txCommand := append(cmd, allTxArgs...)
+	return n.containerManager.ExecCmd(n.t, n.Name, txCommand, success)
+
+	// return n.containerManager.ExecTxCmd(n.t, n.chainId, n.Name, cmd)
```
KonradStaniec, line 203: some leftover ?

KonradStaniec (approved):

x/finality/abci.go
```diff
@@ -20,6 +20,11 @@ func BeginBlocker(ctx context.Context, k keeper.Keeper) error {
...
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyEndBlocker)
 
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	if uint64(sdkCtx.HeaderInfo().Height) < k.GetParams(ctx).FinalityActivationHeight {
+		return []abci.ValidatorUpdate{}, nil
+	}
+
 	// if the BTC staking protocol is activated, i.e., there exists a height where a finality provider
 	// has voting power, start indexing and tallying blocks
 	if _, err := k.BTCStakingKeeper.GetBTCStakingActivatedHeight(ctx); err == nil {
```
gitferry, line 30: nit: I think we can combine the two checks into one validation like `validateActivation` indication that we only handle finality when (1) the height higher than the activation height and (2) at least one fp has voting power at this height

gitferry (approved):
Great work! After some offline discussion, I think adding the activation height in the abci of finality module is a good way to go 👍

test/e2e/configurer/chain/commands_btcstaking.go
RafilxTenfen: e2e is not going nice, still need to fix the e2e test check before merge

Lazar955 (approved):
🙌

test/e2e/configurer/chain/commands_btcstaking.go
RafilxTenfen: finally https://github.com/babylonlabs-io/babylon/pull/204/commits/a5bc7f3bf7827865a4e93f71ded856e128b9026e

test/e2e/containers/containers.go
```diff
@@ -132,6 +132,9 @@ func (m *Manager) ExecCmd(t *testing.T, fullContainerName string, command []stri
 			// Note that this does not match all errors.
 			// This only works if CLI outputs "Error" or "error"
 			// to stderr.
+			fmt.Printf("\n Debug: errOut %s", errBufString)
+			fmt.Printf("\n Debug: output %s", outBuf.String())
+
```
maurolacy, line 137: Nit: Remove debug statements.

x/finality/keeper/msg_server.go
```diff
@@ -53,9 +53,13 @@ func (ms msgServer) AddFinalitySig(goCtx context.Context, req *types.MsgAddFinal
 		return nil, err
 	}
 
-	fpPK := req.FpBtcPk
-
 	ctx := sdk.UnwrapSDKContext(goCtx)
+	activationHeight, errMod := ms.validateActivationHeight(ctx, req.BlockHeight)
+	if errMod != nil {
+		return nil, errMod.Wrapf("finality block height: %d is lower than activation height %d", req.BlockHeight, activationHeight)
```
maurolacy, line 59: Nit: The error message should be returned from inside the validation function, for generality.

x/finality/keeper/msg_server.go
```diff
@@ -188,6 +192,13 @@ func (ms msgServer) CommitPubRandList(goCtx context.Context, req *types.MsgCommi
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), types.MetricsKeyCommitPubRandList)
 
 	ctx := sdk.UnwrapSDKContext(goCtx)
+	activationHeight, errMod := ms.validateActivationHeight(ctx, req.StartHeight)
+	if errMod != nil {
+		return nil, types.ErrFinalityNotActivated.Wrapf(
+			"public rand commit start block height: %d is lower than activation height %d",
```
maurolacy, line 198: Same here.

PR #208, opened 2024-10-17
chore: update exposed db object `Evidence` in `x/finality` module

samricotta:

Currently some of the objects in the babylon apis are exposed and need to be handled correctly seen in https://github.com/babylonlabs-io/pm/issues/72

This PR handles the exposed objects in `x/finality`

Specifically the `Evidence` object.


RafilxTenfen (approved):
:rocket:

PR #211, opened 2024-10-18
mint: Babylon inflation module

SebastianElvis:
Resolves https://github.com/babylonlabs-io/pm/issues/92

This PR provides the implementation of the inflation module. 

- The module shares the same interfaces as `x/mint`, and serves as a drop-in replacement of Cosmos SDK's `x/mint`.
- The module implements a different inflation function. Currently, it employs Celestia's [algorithm](https://docs.celestia.org/learn/staking-governance-supply) where the inflation rate decreases gradually w.r.t. the block timestamp. It contains all necessary dependencies for implementing any inflation function.
- The implementation is adapted from [Celestia](https://github.com/celestiaorg/celestia-app/tree/main/x/mint) and [Cosmos SDK](https://github.com/cosmos/cosmos-sdk/tree/main/x/mint).

Steps for implementation:

- [x] copy https://github.com/celestiaorg/celestia-app/tree/main/x/mint to `x/mint`
- [x] fix all dependencies
- [x] bump module to Cosmos SDK v0.50
- [x] replace Cosmos x/mint with the new x/mint
- [x] fix unit tests
- [x] fix e2e tests
- [x] fix software upgrade e2e
- [x] changelog
- [x] write ADR
- [x] implement the design
- [x] properly initialise the new mint module in upgrade handler
- [x] fix doc


app/upgrades/v1/upgrades.go
```diff
@@ -60,16 +64,43 @@ func CreateUpgradeHandler(upgradeDataStr UpgradeDataString) upgrades.UpgradeHand
 				return nil, err
 			}
 
+			// Re-initialise the mint module as we have replaced Cosmos SDK's
+			// mint module with our own one.
+			// TODO: properly initialise the mint module in the upgrade handler
```
KonradStaniec, line 69: if we are adding new module and deleting new modues, then `StoreUpgarades` part of the upgrade must be updated i.e the part that now contains:
```
			Deleted: []string{ZoneConciergeStoreKey},
		},
```

x/mint/README.md
```diff
@@ -0,0 +1,128 @@
+# `x/mint`
+
+Adapted from <https://github.com/celestiaorg/celestia-app/tree/main/x/mint>
+TODO: fix the doc
+
+---
+
+## Abstract
+
+celestia-app's `x/mint` is a fork of the Cosmos SDK [`x/mint`](https://github.com/cosmos/cosmos-sdk/tree/5cd0b2316a7103468af38eab5d886f9f069c9cd7/x/mint) module that makes some changes to the inflation mechanism. The changes were motivated by a desire for Celestia to have a pre-determined inflation schedule. See [ADR-019](../../docs/architecture/adr-019-strict-inflation-schedule.md) for more details.
```
KonradStaniec, line 10: probably all references to celestia would need to be deleted here

x/mint/abci.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/cosmos/cosmos-sdk/telemetry"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// BeginBlocker updates the inflation rate, annual provisions, and then mints
+// the block provision for the current block.
+func BeginBlocker(ctx context.Context, k keeper.Keeper) {
+	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyBeginBlocker)
+
+	maybeUpdateMinter(ctx, k)
```
KonradStaniec, line 18: just to confirm most of this logic is copy paste from cosmos-sdk mint ?

app/upgrades/v1/upgrades.go
SebastianElvis: We are not adding new module. It's more like that we replace Cosmos SDK's `x/mint` with our own `x/mint`, with the same DB schema and DB key space. That's how the software upgrade passed

x/mint/README.md
SebastianElvis: Yep this will be done after we have the tokenomics and make this PR ready for review

x/mint/abci.go
SebastianElvis: Not really, it's entirely from Celestia https://github.com/celestiaorg/celestia-app/blob/main/x/mint/abci.go

app/upgrades/v1/upgrades.go
KonradStaniec: So if I understand correctly, whatever was stored under `cosmos/x/mint` module keys, will be overwritten by the new values from `babylon/x/mint` but still be stored under old keys ?

SebastianElvis: Yes. And note that the key prefix for this module is same as Cosmos SDK x/mint, i.e., it is `"mint"`.

proto/babylon/mint/v1/genesis.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+package babylon.mint.v1;
+
+option go_package = "github.com/babylonlabs-io/babylon/x/mint/types";
+
+// GenesisState defines the mint module's genesis state.
+message GenesisState {
+  reserved 1; // 1 was previously used for the `Minter` field.
```
RafilxTenfen, line 8: why not keep the minter? https://github.com/cosmos/cosmos-sdk/blob/b49f948b36bc991db5be431607b475633aed697e/proto/cosmos/mint/v1beta1/genesis.proto#L11

I think this is good information

```
  // minter is a space for holding current inflation information.
  Minter minter = 1 [(gogoproto.nullable) = false, (amino.dont_omitempty) = true];

```

and we could migrate the info during the upgrade

proto/babylon/mint/v1/mint.proto
```diff
@@ -0,0 +1,40 @@
...
+package babylon.mint.v1;
+
+option go_package = "github.com/babylonlabs-io/babylon/x/mint/types";
+
+import "gogoproto/gogo.proto";
+import "cosmos_proto/cosmos.proto";
+import "google/protobuf/timestamp.proto";
+
+// Minter represents the mint state.
+message Minter {
```
RafilxTenfen, line 11: we don't need params?

x/mint/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Implementation
+
+See [x/mint](../../x/mint) for the implementation of this module.
+
+## References
+
+1. [ADR-031](https://github.com/babylonlabs-io/pm/blob/main/adr/adr-031-mint-module.md)
+2. [Celestia's
+   ADR-019](../../docs/architecture/adr-019-strict-inflation-schedule.md)
```
RafilxTenfen, line 152: did not found inflation-schedule.md

x/mint/keeper/keeper.go
```diff
@@ -0,0 +1,118 @@
...
+// Keeper of the mint store
+type Keeper struct {
+	cdc              codec.BinaryCodec
+	storeService     store.KVStoreService
+	stakingKeeper    types.StakingKeeper
+	bankKeeper       types.BankKeeper
+	feeCollectorName string
+	// the address capable of executing a MsgUpdateParams message. Typically, this
+	// should be the x/gov module account.
+	authority string
```
RafilxTenfen, line 25: so we are going to have params?

x/mint/keeper/genesis.go
```diff
@@ -0,0 +1,33 @@
...
+	gt := types.GenesisTime{
+		GenesisTime: &blockTime,
+	}
+	if err := k.SetGenesisTime(ctx, gt); err != nil {
+		panic(err)
+	}
+	// Although ak.GetModuleAccount appears to be a no-op, it actually creates a
+	// new module account in the x/auth account store if it doesn't exist. See
+	// the x/auth keeper for more details.
+	ak.GetModuleAccount(ctx, types.ModuleName)
```
RafilxTenfen, line 26: > ak.GetModuleAccount(ctx, types.ModuleName)

We should not do this during initGenesis, if there is a need to create a new module account should be done in app.go or during upgrade

x/mint/types/expected_keepers.go
```diff
@@ -0,0 +1,26 @@
+package types // noalias
```
RafilxTenfen, line 1: what is the noalias ?

x/mint/types/keys.go
```diff
@@ -0,0 +1,23 @@
+package types
+
+// KeyMinter is the key to use for the Minter in the mint store.
+var KeyMinter = []byte("Minter")
```
RafilxTenfen, line 4: the store keys prefixes seems different from the sdk, is this expected?

https://github.com/cosmos/cosmos-sdk/blob/0e34478eb7420b69869ed50f129fc274a97a9b06/x/mint/types/keys.go#L5
```go 

var (
	// MinterKey is the key to use for the keeper store.
	MinterKey = collections.NewPrefix(0)
	ParamsKey = collections.NewPrefix(1)
)

const (
	// module name
	ModuleName = "mint"

	// StoreKey is the default store key for mint
	StoreKey = ModuleName
)

```

proto/babylon/mint/v1/genesis.proto
SebastianElvis: I think this is because in this inflation protocol, the minter object does not depend on the minter object in the previous block, so the protocol does not need to remember the minter in genesis at all. This is different from Cosmos SDK's design where the new inflation depends on the inflation at the previous block.

proto/babylon/mint/v1/mint.proto
SebastianElvis: I think having params in the inflation protocol is deliberately avoided because we don't want gov prop that updates the inflation, at least in the very beginning of the protocol. Apart from typical reasons, one more reason for Babylon is that we are not fully sure whether BTC stakers can get voting power in governance or not.

For now wdyt about adding a TODO there?

x/mint/keeper/keeper.go
SebastianElvis: This was from Celestia code base iirc. I guess they put it here for being future proof 😅

x/mint/types/expected_keepers.go
SebastianElvis: no idea 😩 removed then

x/mint/types/keys.go
SebastianElvis: Good point! No they should be same and let me fix it

x/mint/keeper/genesis.go
SebastianElvis: Tried, and found that we cannot remove it, otherwise will fail.

Checked Cosmos SDK code base and auth module is doing same https://github.com/cosmos/cosmos-sdk/blob/b49f948b36bc991db5be431607b475633aed697e/x/auth/keeper/genesis.go#L28-L29. Probably this is a bug of Cosmos SDK and Celestia people learned it in a hard way?

Wdyt about keeping it?

KonradStaniec (approved):

proto/babylon/mint/v1/genesis.proto
RafilxTenfen: but what about the minted values during the previous version?

proto/babylon/mint/v1/mint.proto
RafilxTenfen: sounds good to me

app/upgrades/v1/upgrades.go
RafilxTenfen, line 69: rmv todo

RafilxTenfen (approved):
LGTM

PR #215, opened 2024-10-21
feat: Implement ADR-29 - generalized unbonding

KonradStaniec:
Highlights:
- Now `MsgBTCUndelegate` contains stake spending transaction as well as inclusion proof
- `BTCDelegation` contains news field `DelegatorUnbondingInfo` instead of old staker signature field
- new field `DelegatorUnbondingInfo` contains information about in which header unbonding happened. Also, if `StakeSpendingTx` in the message was different that registered `UnbondingTx`, `DelegatorUnbondingInfo` contains whole stake spending tx.
- New event is emitted if unbonding happens through different transaction the the pre-registered one.


KonradStaniec:
added more reviewers than normal as this is pretty substantial change to the protocol

proto/babylon/btcstaking/v1/btcstaking.proto
```diff
@@ -109,6 +109,24 @@ message BTCDelegation {
...
+message DelegatorUnbondingInfo {
+    // spend_stake_tx is the transaction which spent the staking output. It is
+    // filled only if spend_stake_tx is different than unbonding_tx registered
+    // on the Babylon chain.
+    bytes spend_stake_tx = 1;
+    // spend_stake_tx_inclusion_block_hash is the block hash of the block in which
+    // spend_stake_tx was included
+    bytes spend_stake_tx_inclusion_block_hash = 2;
+    // spend_stake_tx_inclusion_index is the index of spend_stake_tx in the block
+    uint32 spend_stake_tx_inclusion_index = 3;
```
SebastianElvis, line 127: Do we need to keep those in the KV store? It seems not as they are only be used upon verification but nothing else

KonradStaniec: tbh my main use for those fields was to make it possible to easily check this transaction on Bitcoin node i.e when you have blockhash and transaction index you can retrieve tx from BTC node even if it does not have `--txindex` flag. 

So later it is easier to check unbondings (normal or not) against BTC network , and the cost is only additional 40bytes of data.  Do you think it is useful or overkill from node perspective ?

SebastianElvis: It's a bit overkill imo -- since we have millions of BTC delegations these extra fields might still incur some cost in the states. Also since we know the tx hash, we can find relevant block that contains this tx using Bitcoind APIs. Wdyt?

KonradStaniec: >It's a bit overkill imo -- since we have millions of BTC delegations these extra fields might still incur some cost in the states

For 5m delegation it will be around 50MB of data which seems negligible, but I understand your point: 50MB here, 50MB there and it adds up.

In https://github.com/babylonlabs-io/babylon/pull/215/commits/74db17e7881cc0580cfa334d347cf5afc66f0d8b:
- I have removed it from the state
- though I left it in the event, as this is not part of the state and this piece of data maybe important for somebody

proto/babylon/btcstaking/v1/btcstaking.proto
```diff
@@ -109,6 +109,19 @@ message BTCDelegation {
...
+// - spend_stake_tx: the transaction which spent the staking output
+// - spend_stake_tx_inclusion_block_hash: the block hash of the block in which
+// spend_stake_tx was included
+// - spend_stake_tx_sig_inclusion_index: the index of spend_stake_tx in the block
+message DelegatorUnbondingInfo {
+    // spend_stake_tx is the transaction which spent the staking output. It is
+    // filled only if spend_stake_tx is different than unbonding_tx registered
+    // on the Babylon chain.
+    bytes spend_stake_tx = 1;
+}
```
SebastianElvis, line 123: Do we still need this struct given that it only contains 1 field?

SebastianElvis (approved):

proto/babylon/btcstaking/v1/btcstaking.proto
KonradStaniec: yup, this signals whether unbonding actually happened i.e
```
if del.BtcUndelegation.DelegatorUnbondingInfo != nil
```
delegation is considered unbonded.

And if 

```
len(del.BtcUndelegation.DelegatorUnbondingInfo.SpendStakeTx) > 0 
```

then unbonding happened through unexpected unbonding transaction i.e the transaction different that `del.BtcUndelegation.UnbondingTx`

Assuming most of unbondings will happen through unbonding transaction committed to Babylon, this modelling enables us to save space i.e we do not store stake spending transaction if it happens to be already known unbonding transaction.

SebastianElvis: OK sounds good. Probably we can wrap some functions for `BTCDelegation` to signal these in the future :+1:

proto/babylon/btcstaking/v1/btcstaking.proto
```diff
@@ -109,6 +109,19 @@ message BTCDelegation {
     uint32 params_version = 16;
 }
 
+// DelegatorUnbondingInfo contains the information about transaction which spent
+// the staking output. It contains:
+// - spend_stake_tx: the transaction which spent the staking output
+// - spend_stake_tx_inclusion_block_hash: the block hash of the block in which
+// spend_stake_tx was included
+// - spend_stake_tx_sig_inclusion_index: the index of spend_stake_tx in the block
```
maurolacy, line 117: Nit: remove outdated documentation.

PR #222, opened 2024-10-23
finality: refactor voting power table and APIs

SebastianElvis:
Second half of #24 

This PR moves the voting power table/cache and relevant APIs to `x/finality`. This includes the following

- [x] move voting power table KV store to `x/finality`
- [x] move voting power dist cache to `x/finality`
- [x] move 4 APIs `ActiveFinalityProvidersAtHeight`, `FinalityProviderPowerAtHeight`, `FinalityProviderCurrentPower`, `ActivatedHeight` to `x/finality`
- [x] fix fuzz tests
- [x] fix e2e tests
- [x] fix doc
- [x] fix changelog

Note that this affects many LoCs since it touches a bunch of proto files. The actual modification is much smaller than it looks like.


x/btcstaking/types/errors.go
```diff
@@ -11,26 +11,24 @@ var (
...
+	ErrReusedStakingTx              = errorsmod.Register(ModuleName, 1106, "the BTC staking tx is already used")
+	ErrInvalidCovenantPK            = errorsmod.Register(ModuleName, 1107, "the BTC staking tx specifies a wrong covenant PK")
+	ErrInvalidStakingTx             = errorsmod.Register(ModuleName, 1108, "the BTC staking tx is not valid")
+	ErrInvalidSlashingTx            = errorsmod.Register(ModuleName, 1109, "the BTC slashing tx is not valid")
+	ErrInvalidCovenantSig           = errorsmod.Register(ModuleName, 1110, "the covenant signature is not valid")
+	ErrCommissionLTMinRate          = errorsmod.Register(ModuleName, 1111, "commission cannot be less than min rate")
+	ErrCommissionGTMaxRate          = errorsmod.Register(ModuleName, 1112, "commission cannot be more than one")
+	ErrInvalidDelegationState       = errorsmod.Register(ModuleName, 1113, "Unexpected delegation state")
+	ErrInvalidUnbondingTx           = errorsmod.Register(ModuleName, 1114, "the BTC unbonding tx is not valid")
+	ErrRewardDistCacheNotFound      = errorsmod.Register(ModuleName, 1115, "the reward distribution cache is not found")
```
gitferry, line 24: this is probably not needed

x/btcstaking/types/errors.go
```diff
@@ -11,26 +11,24 @@ var (
...
+	ErrEmptyFpList                  = errorsmod.Register(ModuleName, 1116, "the finality provider list is empty")
+	ErrInvalidProofOfPossession     = errorsmod.Register(ModuleName, 1117, "the proof of possession is not valid")
+	ErrDuplicatedFp                 = errorsmod.Register(ModuleName, 1118, "the staking request contains duplicated finality provider public key")
+	ErrInvalidBTCUndelegateReq      = errorsmod.Register(ModuleName, 1119, "invalid undelegation request")
+	ErrVotingPowerDistCacheNotFound = errorsmod.Register(ModuleName, 1120, "the voting power distribution cache is not found")
```
gitferry, line 29: ditto

x/finality/keeper/grpc_query.go
```diff
@@ -13,11 +13,111 @@ import (
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return &types.QueryActiveFinalityProvidersAtHeightResponse{FinalityProviders: convertToActiveFinalityProvidersAtHeightResponse(finalityProvidersWithMeta), Pagination: pageRes}, nil
+}
+
+// ActivatedHeight returns the Babylon height in which the BTC Staking protocol was enabled
+// TODO: Requires investigation on whether we can enable the BTC staking protocol at genesis
```
gitferry, line 107: seems this TODO is a leftover?

x/finality/keeper/grpc_query.go
```diff
@@ -13,11 +13,111 @@ import (
...
+	activatedHeight, err := k.GetBTCStakingActivatedHeight(sdkCtx)
+	if err != nil {
+		return nil, err
+	}
+	return &types.QueryActivatedHeightResponse{Height: activatedHeight}, nil
+}
+
 // ListPublicRandomness returns a list of public randomness committed by a given
 // finality provider
 // TODO: remove public randomness storage?
```
gitferry, line 123: can we remove this TODO?

x/finality/keeper/power_table.go
```diff
@@ -57,6 +26,14 @@ func (k Keeper) GetVotingPower(ctx context.Context, fpBTCPK []byte, height uint6
 	return sdk.BigEndianToUint64(powerBytes)
 }
 
+// GetVotingPower gets the voting power of a given finality provider at a given Babylon height
+func (k Keeper) GetVotingPower(ctx context.Context, fpBTCPK []byte, height uint64) uint64 {
+	if !k.BTCStakingKeeper.HasFinalityProvider(ctx, fpBTCPK) {
+		return 0
+	}
```
gitferry, line 33: Is this check necessary? `getVotingPower` will also return `0` if the fp does not exist in db

proto/babylon/finality/v1/query.proto
```diff
@@ -18,6 +18,27 @@ service Query {
...
+  // FinalityProviderCurrentPower queries the voting power of a finality provider at the current height
+  rpc FinalityProviderCurrentPower(QueryFinalityProviderCurrentPowerRequest) returns (QueryFinalityProviderCurrentPowerResponse) {
+    option (google.api.http).get = "/babylon/finality/v1/finality_providers/{fp_btc_pk_hex}/power";
+  }
+
+  // ActivatedHeight queries the height when BTC staking protocol is activated, i.e., the first height when
+  // there exists 1 finality provider with voting power
+  rpc ActivatedHeight(QueryActivatedHeightRequest) returns (QueryActivatedHeightResponse) {
+    option (google.api.http).get = "/babylon/finality/v1/activated_height";
+  }
```
gitferry, line 40: Is this query still useful given that we have activation height in the parameter?

x/finality/keeper/tallying.go
```diff
@@ -19,7 +19,7 @@ import (
 // but without block that has finality providers set AND does not receive QC
 func (k Keeper) TallyBlocks(ctx context.Context) {
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
-	activatedHeight, err := k.BTCStakingKeeper.GetBTCStakingActivatedHeight(ctx)
+	activatedHeight, err := k.GetBTCStakingActivatedHeight(ctx)
```
gitferry, line 22: Maybe we should call `k.IsFinalityActive()` here?

gitferry (approved):
Thanks for the heavy lifting! Great work! No blocker from my side. My major question is given that we have already introduce activation height into the parameter, not sure the old activated height (first height with at least 1 fp with positive voting power) is till useful.

x/finality/keeper/tallying.go
gitferry: Seems already checked at the caller's side

SebastianElvis:
> My major question is given that we have already introduce activation height into the parameter, not sure the old activated height (first height with at least 1 fp with positive voting power) is till useful.

I was thinking about this as well, and my answer is that the BTC staking activation height is still needed. There is actually no guarantee when the first BTC delegation will come to the network, so it's possible that it comes before or after the finality activation height.

Probably due to this reason, the `IsFinalityActive` is defined as finality activation height is reached AND BTC staking activation height is set ([ref](https://github.com/babylonlabs-io/babylon/blob/73585306bf1cebff7ba312565870b15c55103c05/x/finality/keeper/keeper.go#L93-L103)), and only when `IsFinalityActive` function returns true the finality module [will](https://github.com/babylonlabs-io/babylon/blob/73585306bf1cebff7ba312565870b15c55103c05/x/finality/abci.go#L23-L43) index/tally blocks and handle jailing. @RafilxTenfen correct me if I'm wrong

x/finality/keeper/tallying.go
SebastianElvis: Yep this function (together with `IndexBlock` and `HandleLiveness`) will be invoked only when `IsFinalityActive` returns true

proto/babylon/finality/v1/query.proto
SebastianElvis: I think we still need it, although we might want to change its definition now. At this moment the activated height should be something like `max(BTC activation height, finality activation height)`, rather than just the BTC activation height. Will create a Github issue for this

proto/babylon/finality/v1/genesis.proto
```diff
@@ -27,6 +27,11 @@ message GenesisState {
   // missed_blocks represents a map between finality provider public key and their
   // missed blocks.
   repeated FinalityProviderMissedBlocks missed_blocks = 8 [(gogoproto.nullable) = false];
+  // voting_powers the voting power of every finality provider at every block height.
+  repeated VotingPowerFP voting_powers = 9;
+  // vp_dst_cache is the table of all providers voting power with the total at one specific block.
+  // TODO: remove this after not storing in the keeper store it anymore.
```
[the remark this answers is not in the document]
SebastianElvis: I don't know what is this about actually :sweat_smile: But probably we will need to have `vp_dst_cache` in DB for calculating the reward, so :+1: on removing it

gitferry:
> I was thinking about this as well, and my answer is that the BTC staking activation height is still needed. There is actually no guarantee when the first BTC delegation will come to the network, so it's possible that it comes before or after the finality activation height.
> Probably due to this reason, the IsFinalityActive is defined as finality activation height is reached AND BTC staking activation height is set ([ref](https://github.com/babylonlabs-io/babylon/blob/73585306bf1cebff7ba312565870b15c55103c05/x/finality/keeper/keeper.go#L93-L103)), and only when IsFinalityActive function returns true the finality module [will](https://github.com/babylonlabs-io/babylon/blob/73585306bf1cebff7ba312565870b15c55103c05/x/finality/abci.go#L23-L43) index/tally blocks and handle jailing. @RafilxTenfen correct me if I'm wrong

I think our expectation of introducing this parameter is that after activation, fps can submit votes immediately. Even if there are no delegations or no timestamped pub rand after activation, we should still enable finality-related functionalities.

Maybe it's better to move to https://github.com/babylonlabs-io/pm/pull/85 for discussion as this is out of the scope of this PR. Sorry about that 😅

proto/babylon/finality/v1/query.proto
SebastianElvis: https://github.com/babylonlabs-io/babylon/issues/231

SebastianElvis:
Merging this PR now. Any further review could go to https://github.com/babylonlabs-io/babylon/issues/24

PR #224, opened 2024-10-23
chore: Make injected checkpoint a standard tx

### LINKED ISSUE #129: making InjectedCheckpoint a standard tx
Context: https://github.com/babylonchain/babylon/issues/555

gitferry:
Closes #129 by introducing the encoder and decoder into the proposal handler


x/checkpointing/proposal.go
```diff
@@ -30,13 +34,19 @@ func NewProposalHandler(
...
+	ckpttypes.RegisterInterfaces(encCfg.InterfaceRegistry)
+
 	return &ProposalHandler{
 		logger:                        logger,
 		ckptKeeper:                    ckptKeeper,
-		txVerifier:                    txVerifier,
+		bApp:                          bApp,
+		txEncoder:                     encCfg.TxConfig.TxEncoder(),
+		txDecoder:                     encCfg.TxConfig.TxDecoder(),
+		txFactory:                     tx.Factory{}.WithChainID(bApp.ChainID()).WithTxConfig(encCfg.TxConfig),
```
KonradStaniec, line 49: wdyt about using:
```
encCfg.TxConfig.NewTxBuilder()
```
directly to build transactions instead of ` tx.Factory{}` ?

My main fear is that `txFactory.BuildUnsignedTx` have a lot of internal logic itself, and this logic may change in time (with patch or minor version) and we do not catch this change and hit some consensus error. Ideally we would build this transaction with least amout of logic possible i.e just wrap our message with transaction proto.

x/checkpointing/proposal.go
```diff
@@ -30,13 +37,19 @@ func NewProposalHandler(
...
+	ckpttypes.RegisterInterfaces(encCfg.InterfaceRegistry)
+
 	return &ProposalHandler{
 		logger:                        logger,
 		ckptKeeper:                    ckptKeeper,
-		txVerifier:                    txVerifier,
+		bApp:                          bApp,
+		txEncoder:                     encCfg.TxConfig.TxEncoder(),
+		txDecoder:                     encCfg.TxConfig.TxDecoder(),
+		txBuilder:                     encCfg.TxConfig.NewTxBuilder(),
```
KonradStaniec, line 52: doesn't we need new `encCfg.TxConfig.NewTxBuilder()` per transaction ?  i.e every time we build checkpoint we call:
```
buiilder :=  encCfg.TxConfig.NewTxBuilder()

builder.SetMsgs(checkpoint)

builder.GetTx()
```

gitferry: I think it should be fine as every time `SetMsgs` is called, the `tx.Body` will be reset https://github.com/cosmos/cosmos-sdk/blob/62212dfc04b3790f3cc052cdc2fb757b32ca97d0/x/auth/tx/builder.go#L266

KonradStaniec: ah ok! thanks for double checking, then it is fine 👍  (it would be not fine if there would be some kind of concurrent access to prepare proposal, but I do not belive it is the case)

Lazar955 (approved):
Nice work!

PR #228, opened 2024-10-24
chore: Add inclusion height to unbonding

### LINKED ISSUE #225: Insert unbonding height to early unbonding event
[context](https://github.com/babylonlabs-io/babylon-staking-indexer/issues/28)

gitferry:
Closes #225.

* added inclusion height to delegator unbonding info in db
* added unbonding start height to early unbonding event and delegation query


proto/babylon/btcstaking/v1/btcstaking.proto
```diff
@@ -112,14 +112,14 @@ message BTCDelegation {
...
-// spend_stake_tx was included
-// - spend_stake_tx_sig_inclusion_index: the index of spend_stake_tx in the block
+// - inclusion_height: the height of the block in which the spend_staking_tx is included
 message DelegatorUnbondingInfo {
     // spend_stake_tx is the transaction which spent the staking output. It is
     // filled only if spend_stake_tx is different than unbonding_tx registered
     // on the Babylon chain.
     bytes spend_stake_tx = 1;
+    // inclusion_height is the height of the block in which the spend_staking_tx is included
+    uint32 inclusion_height = 2;
```
KonradStaniec, line 122: So this makes me a bit nervous, as due to re-orgs is is possible that this inclusion height can change. So we can be left with data is database that do not match reality.

In other words, I am fine with putting this info in event which is kind of ephemeral thing and we can treat it as `inclusion height at the time of receiving unbonding transaction on BTC` but I am not too sure about storing this data in database.

gitferry: I agree that re-orgs can cause mismatch issues. The tricky thing is that in our current flow, we deal with undelegate in message handler but deal with the relevant event in `BeginBlocker`. This means that the inclusion height is passed via db write and read. If we don't want to store the height in db, we would have to emit the early unbonding event in the message handler directly. So, two options:
1. leave it what it is with a note that the inclusion height is subjective to re-orgs, maybe rename it to `reported_inclusion_height` as it is only used for emitting events for now
2. do not store the inclusion height but emit the early unbonding event in the message handler. This will change the logic of handling delegation events a bit and make the early unbonding a tx event

I'll try with option 2 to see if it could work without significant change

gitferry: it can be replayed just like other events

gitferry: Thanks @jrwbabylonlab and @KonradStaniec for the discussion. I have changed it to (2) in [59bfc27](https://github.com/babylonlabs-io/babylon/pull/228/commits/59bfc27b60e0f6ad655d04b372b263c8d85a2eba).

@jrwbabylonlab please keep in mind that early unbonding will be a tx event while expired unbonding is still a block event.

Btw, do we want to backport it to `release/v0.14.x`?

KonradStaniec (approved):
Looks good 👍 

I think we can backport it and release 0.14.1, as it is not consensus breaking

PR #230, opened 2024-10-25
backport: Insert inclusion height to early unbonding event

gitferry:


proto/babylon/btcstaking/v1/events.proto
```diff
@@ -215,8 +215,10 @@ message EventBTCDelgationUnbondedEarly {
   // staking_tx_hash is the hash of the staking tx.
   // It uniquely identifies a BTC delegation
   string staking_tx_hash = 1 [(amino.dont_omitempty) = true];
+  // start_height is the start BTC height of the early unbonding
+  string start_height = 2 [(amino.dont_omitempty) = true];
```
maurolacy, line 219: Why is this not a uint32 or uint64? Just curious.

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -188,8 +188,10 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 					activeBTCDels[fpBTCPKHex] = append(activeBTCDels[fpBTCPKHex], btcDel)
 				}
 			} else if delEvent.NewState == types.BTCDelegationStatus_UNBONDED {
-				// emit event about this unbonded BTC delegation
-				types.EmitUnbondedBTCDelEvent(sdkCtx, delEvent.StakingTxHash, btcDel.IsUnbondedEarly())
+				// emit expired event if it is not early unbonding
+				if !btcDel.IsUnbondedEarly() {
+					types.EmitExpiredDelegationEvent(sdkCtx, delEvent.StakingTxHash)
+				}
```
maurolacy, line 194: What if it's an early unbond? No event emitted?

proto/babylon/btcstaking/v1/events.proto
gitferry: This is for easier parsing for subscribers of the event

x/btcstaking/keeper/power_dist_change.go
gitferry: It is now emitted during processing undelegate message https://github.com/babylonlabs-io/babylon/pull/228/files#diff-0490899e13cee0223ce9b01147d9ad2753112c4e8118443e1b54684738fc4969R585. The reason we changed the emitting there is because we don't want to save the inclusion height in db considering re-org could happen on Bitcoin. [Discussions](https://github.com/babylonlabs-io/babylon/pull/228#discussion_r1814288603)

PR #234, opened 2024-10-29
chore: add happy case implementation of btc distribution

RafilxTenfen:


RafilxTenfen:
closing this PR as other distribution will be implemented, probably following F1

PR #242, opened 2024-10-31
feat(adr-032): Add ResumeFinalityProposal and handler

gitferry:
This PR introduces the `ResumeFinalityProposal` and implements the handler, which is part of [ADR-32](https://github.com/babylonlabs-io/pm/pull/95). This part is quite independent and the algorithm of choosing finality providers to jail can be implemented in the future


x/finality/keeper/gov_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	for i := haltingHeight; i < currentHeight; i++ {
+		ib, err := fKeeper.GetBlock(ctx, i)
+		require.NoError(t, err)
+		require.True(t, ib.Finalized)
+	}
+}
+
+func generateNFpPks(t *testing.T, r *rand.Rand, n int) []bbntypes.BIP340PubKey {
+	fpPks := make([]bbntypes.BIP340PubKey, 0)
```
Lazar955, line 86:
```suggestion
	fpPks := make([]bbntypes.BIP340PubKey, 0, n)
```

x/finality/client/cli/gov.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	fpPks := make([]bbntypes.BIP340PubKey, len(fps.FinalityProviders))
+	for i, pkStr := range fps.FinalityProviders {
+		pk, err := bbntypes.NewBIP340PubKeyFromHex(pkStr)
+		if err != nil {
+			return nil, fmt.Errorf("invalid finality provider public key %s: %w", pkStr, err)
+		}
+		fpPks[i] = *pk
+	}
```
Lazar955, line 90:
```suggestion
	fpPks := make([]bbntypes.BIP340PubKey, 0, len(fps.FinalityProviders)) // avoid wrangling indexes
	for i, pkStr := range fps.FinalityProviders {
		pk, err := bbntypes.NewBIP340PubKeyFromHex(pkStr)
		if err != nil {
			return nil, fmt.Errorf("invalid finality provider public key %s: %w", pkStr, err)
		}
		fpPks = append(fpPks, *pk)
	}
```

x/finality/gov.go
```diff
@@ -0,0 +1,29 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	govtypesv1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
+
+	"github.com/babylonlabs-io/babylon/x/finality/keeper"
+	"github.com/babylonlabs-io/babylon/x/finality/types"
+)
+
+// NewResumeFinalityProposalHandler is a handler for governance proposal on resume finality.
+func NewResumeFinalityProposalHandler(k keeper.Keeper) govtypesv1.Handler {
```
KonradStaniec, line 14: hmm why usage of the old way of handling proposals ? (which I think is discouraged) 

Wouldn't it be better to have separate message on finality provider message server which can be only executed by governance module ? (similar to parameters updates)

Lazar955 (approved):

KonradStaniec (approved):

PR #249, opened 2024-11-05
chore: Add logs to checking unbonding output

gitferry:


x/btcstaking/types/validate_parsed_message.go
```diff
@@ -129,18 +131,25 @@ func ValidateParsedMessageAgainstTheParams(
...
+		return nil, ErrInvalidUnbondingTx.
+			Wrapf(" the unbonding output value is not expected, expected: %d, got: %d",
+				unbondingTx.TxOut[0].Value, unbondingInfo.UnbondingOutput.Value)
 	}
 
 	err = btcstaking.CheckSlashingTxMatchFundingTx(
 		pm.UnbondingSlashingTx.Transaction,
 		pm.UnbondingTx.Transaction,
-		unbondingOutputIdx,
+		0, // unbonding output always has only 1 output
```
KonradStaniec, line 152: hmm if we are hardcoidng this, I feel than we should have a test case to assert that unbonding tx with more that `1` output are rejcected. Maybe in `validate_parsed_message_test.go`

gitferry: good point. Added this case in [e2916de](https://github.com/babylonlabs-io/babylon/pull/249/commits/e2916deecf59750d0c7313cc9f803af59216cd1f)

KonradStaniec (approved):

PR #253, opened 2024-11-06
chore(*): upgrade comet bft and go version

Lazar955:
Upgrades comet bft to v0.38.14


Lazar955:
Maybe someone else should do this upgrade if we don't want to bump the go version (I think we should).

KonradStaniec (approved):

gitferry (approved):
LGTM! I think we should double check the cometBFT version in deployment

gitferry (commented):
Just noticed that we should also update README about go version

Issue #243, opened 2024-10-31
Constant decoding of bech32 address


KonradStaniec:
Currently when processing rewards we spend a lot of time decoding bech32 address to bytes  - https://github.com/babylonlabs-io/babylon/blob/v0.15.0/x/finality/types/power_table.go#L200. 

We should store the address as bytes directly.

<img width="1496" alt="Screenshot 2024-10-31 at 11 49 33" src="https://github.com/user-attachments/assets/038f1554-212e-41bd-b59b-33402dd2c406">

KonradStaniec:
fixed with: https://github.com/babylonlabs-io/babylon/pull/254

PR #254, opened 2024-11-07
Avoid unnecessary bech32 decoding

KonradStaniec:
Solves issues: https://github.com/babylonlabs-io/babylon/issues/243


Lazar955 (approved):
Nice work!

PR #258, opened 2024-11-09
feat: fix go releaser and hook by github action

liam-icheng-lai:
Make github action actually make a go release package and added into release note to be able download by all the audience. Remove libwasmvm for github action using whatever is defined in the go mod file library.


SebastianElvis (approved):
Thanks!

How do we test this fix btw?

liam-icheng-lai:
> Thanks!
> 
> How do we test this fix btw?

I already make the release from the branch and it works

PR #260, opened 2024-11-12
incentive: full tokenomics implementation

SebastianElvis:
Resolves https://github.com/babylonlabs-io/pm/issues/100

This PR implements the leftover issues in the tokenmomics, including

- removing BTC timestamping related incentive mechanisms
- rewarding all active FPs rather than those who voted (we can do this because of jailing mechanism, and this is in line with Cosmos SDK)

The rest of the [tokenomics issue](https://github.com/babylonlabs-io/pm/issues/100) are mostly parameters and benchmarking, which depend on external factors and other engineering efforts.


proto/babylon/incentive/params.proto
```diff
@@ -13,18 +13,12 @@ option go_package = "github.com/babylonlabs-io/babylon/x/incentive/types";
...
-        (gogoproto.nullable)   = false
-    ];
-    // reporter_portion is the portion of rewards that goes to reporter
-    string reporter_portion = 2 [
-        (cosmos_proto.scalar)  = "cosmos.Dec",
-        (gogoproto.customtype) = "cosmossdk.io/math.LegacyDec",
-        (gogoproto.nullable)   = false
-    ];
+    // reserved 1
+    reserved 1;
```
KonradStaniec, line 17: do we have some e2e test retrieving incentive params showing they are deserialized correctly ? also will `btc_staking_portion ` be correct after upgrade or we need to run governance proposal to make it correct ? 

Asking as imo, we should handle existing params change in the same way we handle it on `x/finality` and `x/btcstaking` module i.e override exsiting params exstinsting in TGE version with new protos in upgrade handler

SebastianElvis: > do we have some e2e test retrieving incentive params showing they are deserialized correctly ? also will `btc_staking_portion ` be correct after upgrade or we need to run governance proposal to make it correct ?
>
> Asking as imo, we should handle existing params change in the same way we handle it on `x/finality` and `x/btcstaking` module i.e override exsiting params exstinsting in TGE version with new protos in upgrade handler

Hmm good point, overriding the param in TGE version is a better option. Will revise and test this in e2e later

SebastianElvis: Addressed in [`9f209ba` (#260)](https://github.com/babylonlabs-io/babylon/pull/260/commits/9f209bacd46ab5ce05684855c833d81b867fb28c)

KonradStaniec (approved):

Lazar955 (approved):
lgtm!

PR #265, opened 2024-11-13
Phase-1 stake ingesting - RFC15

KonradStaniec:
Implement:
- allow list that will expire at certain Babylon block
- upgrade to insert allowed transactions hashes
- modification in message handler to use allow list under certain Babylon height

i.e approach 4, from [RFC-15](https://github.com/babylonlabs-io/pm/blob/main/rfc/rfc-015-moderate-intake-of-phase-1-stake.md#4-fixed-list-of-allowed-transactions-with-parameter-how-long-to-keep-the-allow-list)

Rationale for putting allowlist expiration height in params:
- if we discover some problems during initial blocks of phase-2, we may extend the periond when allow list is used
- we should be able to disable allow list for testnet (setting expiration height to `0` disable allow list)


Lazar955 (approved):
Nice work!

PR #273, opened 2024-11-18
Add staking tx to the creation event

KonradStaniec:
- Add staking tx to the event 
- Removes from the event data than can be retrieved from staking tx:
  - staking tx hash
  - staking output pk script
  - staking output amount


gitferry (approved):
LGTM!

PR #278, opened 2024-11-19
chore: Allow unbonding time to be min unbonding value

### LINKED ISSUE #263: Allow choosing `min_unbonding_time` in the params as unbonding time
Currently the unbonding time in a delegation is checked by `unbonding_time > max(min_unbonding_time, checkpoint_finalization_timeout)` which means the unbonding time must be higher than `min_unbonding_time`, which is not aligned with other min values in the params where inclusive value is allowed. This causes confusion for the frontend to show a valid unbonding time to the users.

An ideal solution is to allow using a min value for the unbonding time and enforce the rule that `min_unbonding_time > checkpoint_finalization_timeout`.

gitferry:
Closes #263. In particular,
- ensure `checkpointFinalizationTimeout` can never be changed in the update params handler of the btccheckpoint module
- ensure `minUnbondingTime` can not be set to a value less than or equal to `checkpointFinalizationTimeout` in the update params handler of the btcstaking module
- allows `unbondingTime` of a delegation to be set to a value equals to `minUnbondingTime`

### DIFF
--- a/x/btccheckpoint/keeper/msg_server.go
+++ b/x/btccheckpoint/keeper/msg_server.go
@@ -4,9 +4,10 @@ import (
 	"context"
 
 	errorsmod "cosmossdk.io/errors"
-	"github.com/babylonlabs-io/babylon/x/btccheckpoint/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+
+	"github.com/babylonlabs-io/babylon/x/btccheckpoint/types"
 )
 
 var _ types.MsgServer = msgServer{}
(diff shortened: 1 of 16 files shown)


x/btccheckpoint/keeper/msg_server.go
```diff
@@ -110,6 +111,10 @@ func (ms msgServer) UpdateParams(goCtx context.Context, req *types.MsgUpdatePara
 	}
 
 	ctx := sdk.UnwrapSDKContext(goCtx)
+	if req.Params.CheckpointFinalizationTimeout != ms.k.GetParams(ctx).CheckpointFinalizationTimeout {
```
RafilxTenfen, line 114: can we add a comment here on why it can't be changed?

x/btcstaking/types/validate_parsed_message.go
```diff
@@ -206,6 +201,6 @@ func ValidateParsedMessageAgainstTheParams(
 	return &ParamsValidationResult{
 		StakingOutputIdx:   stakingOutputIdx,
 		UnbondingOutputIdx: 0, // unbonding output always has only 1 output
-		MinUnbondingTime:   minUnbondingTime,
+		MinUnbondingTime:   parameters.MinUnbondingTimeBlocks,
```
KonradStaniec, line 204: as we no longer calculate valid `MinUnbondingTime` in this functions, we can probably remove this field from `ParamsValidationResult` . (the caller has access to ` parameters.MinUnbondingTimeBlocks` eitherway)

KonradStaniec (approved):
just minor comment

x/btcstaking/types/validate_parsed_message.go
gitferry: Good point. Fixed in [6b4abc7](https://github.com/babylonlabs-io/babylon/pull/278/commits/6b4abc761e3966366f68f87982ab0c678132003d)

PR #289, opened 2024-11-21
hotfix: Invalid minUnbondingTime for verifying inclusion proof

gitferry:
The bug is caused due to setting `params.MinStakingTime` to the btcstaking module's keeper method `VerifyInclusionProofAndGetHeight` other than `params.MinUnbondingTime`. In the fix, we remove `minUnbondingTime` from the parameter list of `VerifyInclusionProofAndGetHeight` as it should retrieve the parameter within the method.

This PR also added comprehensive tests for `VerifyInclusionProofAndGetHeight`

### DIFF
--- a/x/btcstaking/keeper/msg_server.go
+++ b/x/btcstaking/keeper/msg_server.go
@@ -215,6 +215,7 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 			btcutil.NewTx(parsedMsg.StakingTx.Transaction),
 			btccParams.BtcConfirmationDepth,
 			uint32(parsedMsg.StakingTime),
+			vp.Params.MinUnbondingTimeBlocks,
 			parsedMsg.StakingTxProofOfInclusion)
 		if err != nil {
 			return nil, fmt.Errorf("invalid inclusion proof: %w", err)
@@ -314,6 +315,7 @@ func (ms msgServer) AddBTCDelegationInclusionProof(
 		btcutil.NewTx(stakingTx),
 		btccParams.BtcConfirmationDepth,
 		btcDel.StakingTime,
+		minUnbondingTime,
 		parsedInclusionProof,
 	)
 
--- a/x/btcstaking/keeper/msg_server.go
+++ b/x/btcstaking/keeper/msg_server.go
@@ -215,14 +215,13 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 			btcutil.NewTx(parsedMsg.StakingTx.Transaction),
 			btccParams.BtcConfirmationDepth,
 			uint32(parsedMsg.StakingTime),
-			vp.Params.MinStakingTimeBlocks,
 			parsedMsg.StakingTxProofOfInclusion)
 		if err != nil {
 			return nil, fmt.Errorf("invalid inclusion proof: %w", err)
 		}
 
-		startHeight = timeInfo.startHeight
-		endHeight = timeInfo.endHeight
+		startHeight = timeInfo.StartHeight
+		endHeight = timeInfo.EndHeight
 	} else {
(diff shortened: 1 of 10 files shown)


RafilxTenfen:
I am a little bit confused: 
- The `getBTCDelWithParams` returns the delegation in question and the versioned parameters respected to create the same delegation
- To verify the inclusion proof we look at the last params stored at the `x/btcstaking` 
- To update the voting power distribution we only count for the delegation end height - the versioned parameters used at the btc delegation creation


Why do we not use the latest parameter to update the voting power dist as well?

x/btcstaking/keeper/msg_server.go
```diff
@@ -315,7 +314,6 @@ func (ms msgServer) AddBTCDelegationInclusionProof(
 		btcutil.NewTx(stakingTx),
 		btccParams.BtcConfirmationDepth,
 		btcDel.StakingTime,
-		minUnbondingTime,
```
KonradStaniec, line 318: I think @RafilxTenfen comment is valid here - https://github.com/babylonlabs-io/babylon/pull/289#issuecomment-2491815295, ie we should pass here the `params.MinUnbondingTimeBlocks`  from the params used when creating the delegation, not retrieve latest one from the store inside the function

gitferry:
@RafilxTenfen @KonradStaniec good point, thanks guys! I forgot we are using versioned params. Fixed in [85689cb](https://github.com/babylonlabs-io/babylon/pull/289/commits/85689cb2b644c4729a7aebd019583e0f8c03c29a). Added a note there and relevant test

x/btcstaking/keeper/msg_server.go
```diff
@@ -215,14 +215,14 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 			btcutil.NewTx(parsedMsg.StakingTx.Transaction),
 			btccParams.BtcConfirmationDepth,
 			uint32(parsedMsg.StakingTime),
-			vp.Params.MinStakingTimeBlocks,
```
KonradStaniec, line 218: could you double check the tests ? 

I have changed this back locally to `vp.Params.MinStakingTimeBlocks` and run the tests, and they did not catch this.

gitferry: Oh that's because the tests are for `VerifyInclusionProofAndGetHeight`. Seems we need to have a special case for testing `CreateBTCDelegation`

gitferry: In [d729190](https://github.com/babylonlabs-io/babylon/pull/289/commits/d7291907ad6ec64b5451d6b22138d0d03898cb86) I added a happy case to `TestDoNotAllowDelegationWithoutFinalityProvider` so that if you change `vp.Params.MinStakingTimeBlocks`, the test won't pass. But I don't think it's comprehensive enough. We can create an issue to add comprehensive test for `CreateBTCDelegation`

gitferry: I also noticed that we don't have a validation rule to ensure `minStakingTime > minUnbondingTime` in params. I think this rule is important. Otherwise, the [verification](https://github.com/babylonlabs-io/babylon/blob/85689cb2b644c4729a7aebd019583e0f8c03c29a/x/btcstaking/keeper/inclusion_proof.go#L65) will never pass if the user choose to stake with `minStakingTime`:

```
	// ensure staking tx's timelock has more than unbonding BTC blocks left
	if btcTip.Height+minUnbondingTime >= endHeight {
		return nil, types.ErrInvalidStakingTx.
			Wrapf("staking tx's timelock has no more than unbonding(=%d) blocks left", minUnbondingTime)
	}
```

KonradStaniec: >But I don't think it's comprehensive enough. We can create an issue to add comprehensive test for CreateBTCDelegation

Sure, lets create issue for this 👍

KonradStaniec (approved):

PR #291, opened 2024-11-22
zoneconcierge: permissioned integration

SebastianElvis:
Resolves https://github.com/babylonlabs-io/pm/issues/44

This PR implements the permissioned integration and generic consumer registry, including

- [x] generic consumer registry compatible with Cosmos/ETH L2 consumers
- [x] parameter for governance gating consumer addition
- [x] ensuring IBC channel creation only when the IBC light client is in consumer registry
- [x] removing the Cosmos consumer registration upon IBC handshake in Babylon
- [x] removing the Cosmos consumer registration upon IBC handshake in Babylon contract
- [x] fuzz tests for Cosmos integration
- [x] e2e tests for Cosmos integration

ADR: https://github.com/babylonlabs-io/pm/pull/121
Accompanying PR in the contract side: https://github.com/babylonlabs-io/babylon-contract/pull/87

There are two TODOs that will be addressed in subsequent PRs:

- msg for removing registered consumer
- fuzz/e2e test for ETH L2 integrations


proto/babylon/btcstkconsumer/v1/btcstkconsumer.proto
```diff
@@ -1,15 +1,49 @@
...
+  ETH_L2 = 1;
+}
+
 // ConsumerRegister is the registration information of a consumer
 message ConsumerRegister {
   // consumer_id is the ID of the consumer
   // - for Cosmos SDK chains, the consumer ID will be the IBC client ID
+  // - for ETH L2 chains, the consumer ID will be the chain ID of the ETH L2
+  //   chain
   string consumer_id = 1;
```
maurolacy, line 23: Nit: What about adding this field to both metadatas below, with a proper description for each?

proto/babylon/btcstkconsumer/v1/btcstkconsumer.proto
```diff
@@ -1,15 +1,49 @@
...
+  oneof consumer_metadata {
+    CosmosConsumerMetadata cosmos_consumer_metadata = 4;
+    ETHL2ConsumerMetadata eth_l2_consumer_metadata = 5;
+  };
+}
+
+// CosmosConsumerMetadata is the metadata for the Cosmos integration
+message CosmosConsumerMetadata {
+  // client_id defines the IBC client ID for the consumer chain
+  string client_id = 1;
```
maurolacy, line 39: This is redundant with the main struct. Better to move `consumer_id` here, and replicate it in the Ethereum struct, with a proper description.

Moreover: Perhaps it's simpler / clearer to just make `ConsumerRegister` a `oneof`. At the expense of some redundancy in the consumer name and description fields.

proto/babylon/btcstkconsumer/v1/params.proto
```diff
@@ -10,4 +10,8 @@ option go_package = "github.com/babylonlabs-io/babylon/x/btcstkconsumer/types";
 message Params {
   option (amino.name) = "babylon/x/btcstkconsumer/v1/Params";
   option (gogoproto.equal) = true;
-}
\ No newline at end of file
+
+  // permissioned_integration is a flag to enable permissioned integration, i.e.,
+  // requiring governance proposal to approve new integrations.
+  bool permissioned_integration = 1;
```
maurolacy, line 16: Can we set a default here? Just curious.

proto/babylon/btcstkconsumer/v1/tx.proto
```diff
@@ -52,6 +52,9 @@ message MsgRegisterConsumer {
   string consumer_name = 3;
   // consumer_description is a description for the consumer (can be empty)
   string consumer_description = 4;
+  // eth_l2_finality_contract_address is the address of the finality contract
+  // (if set, then this means this is an ETH L2 integration)
+  string eth_l2_finality_contract_address = 5 [(cosmos_proto.scalar) = "cosmos.AddressString"];
```
maurolacy, line 57: This is a little obscure. This is the Cosmos address of the "assistant" finality contract in the Babylon chain for that Eth-L2 consumer, right? Better to clarify all that in the comments.

Again, what about making `MsgRegisterConsumer` a `oneof` here? Not sure this is a good idea, or even possible, just saying.

proto/babylon/zoneconcierge/v1/packet.proto
```diff
@@ -18,8 +18,7 @@ message ZoneconciergePacketData {
   oneof packet {
     BTCTimestamp btc_timestamp = 1;
     babylon.btcstaking.v1.BTCStakingIBCPacket btc_staking = 2;
-    ConsumerRegisterIBCPacket consumer_register = 3;
```
maurolacy, line 21: Bye nice automatic consumer registration.🙂

proto/babylon/zoneconcierge/v1/packet.proto
```diff
@@ -18,8 +18,7 @@ message ZoneconciergePacketData {
...
-    ConsumerSlashingIBCPacket consumer_slashing = 4;
+    ConsumerSlashingIBCPacket consumer_slashing = 3;
```
maurolacy, line 21: Nit: Again, changing the tags is not strictly needed.

maurolacy (commented):
Partial review of the build targets / scripts and proto files.

Will review the code next.

Makefile
```diff
@@ -277,7 +277,7 @@ clean-e2e:
 	docker container rm -f $(shell docker container ls -a -q) || true
 	docker network prune -f || true
 
-test-e2e-cache-bcd-consumer-integration: start-bcd-consumer-integration
+test-e2e-cache-bcd-consumer-integration:
```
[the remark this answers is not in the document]
SebastianElvis: I was basically following other e2e entries, but yes you are right that the target should not have dependencies. Added it back and renamed this entry to `test-e2e-bcd-consumer-integration`

proto/babylon/btcstaking/v1/packet.proto
```diff
@@ -58,70 +58,70 @@ message NewFinalityProvider {
 // upon a BTC delegation newly receives covenant signatures and thus becomes active
 message ActiveBTCDelegation {
   // staker_addr is the address to receive rewards from BTC delegation.
-  string staker_addr = 11;
+  string staker_addr = 1;
```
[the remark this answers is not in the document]
SebastianElvis: Yep let's have clean proto files at least before the phase-3 testnet/mainnet launch

proto/babylon/btcstkconsumer/v1/btcstkconsumer.proto
SebastianElvis: I would prefer having it here because every consumer, regardless of its type, will have a consumer ID. Wdyt?

proto/babylon/btcstkconsumer/v1/btcstkconsumer.proto
SebastianElvis: Yep this is redundant, let's remove this field. 

Personally I still prerfer to keep the current structure, as there might exist some applications that only care about id/name/description. We can follow up on this after backend/frontend comes up with concrete requirements.

proto/babylon/btcstkconsumer/v1/params.proto
SebastianElvis: It's set [here](https://github.com/babylonlabs-io/babylon/blob/de84d8817da424189fa5110ac00c2e3dc1333949/x/btcstkconsumer/types/params.go#L17-L18)

proto/babylon/zoneconcierge/v1/packet.proto
SebastianElvis: Yeah somehow this turns out to be much more complicated -- both Babylon and consumer have to pass gov prop for integration, and one cannot simply deploy a chain that integrates with Babylon without any approval

contrib/images/ibcsim-bcd/Dockerfile
```diff
@@ -53,9 +53,9 @@ COPY --from=build-env /go/bin/bcd /usr/bin/bcd
 WORKDIR /ibcsim-bcd
 COPY contrib/images/ibcsim-bcd/wrapper.sh /ibcsim-bcd/wrapper.sh
 COPY contrib/images/ibcsim-bcd/setup-bcd.sh /ibcsim-bcd/setup-bcd.sh
-COPY --from=build-env /work/babylon-sdk/tests/testdata/babylon_contract.wasm /ibcsim-bcd/babylon_contract.wasm
-COPY --from=build-env /work/babylon-sdk/tests/testdata/btc_staking.wasm /ibcsim-bcd/btc_staking.wasm
-COPY --from=build-env /work/babylon-sdk/tests/testdata/btc_finality.wasm /ibcsim-bcd/btc_finality.wasm
+COPY test/e2e/bytecode/babylon_contract.wasm /ibcsim-bcd/babylon_contract.wasm
+COPY test/e2e/bytecode/btc_staking.wasm /ibcsim-bcd/btc_staking.wasm
+COPY test/e2e/bytecode/btc_finality.wasm /ibcsim-bcd/btc_finality.wasm
```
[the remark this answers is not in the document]
SebastianElvis: Make sense. Let me create a PR for babylon-sdk, and let this Dockerfile use the latest `main` of babylon-sdk. Wdyt?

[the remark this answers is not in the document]
SebastianElvis: https://github.com/babylonlabs-io/babylon-sdk/pull/65

SebastianElvis:
Thanks @maurolacy for the comments! I have revised all of them and answered some questions. Please feel free to take a look again

TODO before merging

- [ ] merge https://github.com/babylonlabs-io/babylon-sdk/pull/65

x/btcstkconsumer/types/expected_keepers.go
```diff
@@ -18,6 +20,17 @@ type BankKeeper interface {
...
+type ClientKeeper interface {
+	GetClientState(ctx sdk.Context, clientID string) (ibcexported.ClientState, bool)
+}
+
+type WasmKeeper interface {
+	GetContractInfo(ctx context.Context, contractAddress sdk.AccAddress) *wasmtypes.ContractInfo
+}
+
+type ChannelKeeper interface {
+}
```
maurolacy, line 32: Why do we need this channel keeper if it provides no interface? Just curious.

x/btcstkconsumer/types/params.go
```diff
@@ -11,14 +11,11 @@ func ParamKeyTable() paramtypes.KeyTable {
...
-// NewParams creates a new Params instance
-func NewParams() Params {
-	return Params{}
-}
-
 // DefaultParams returns a default set of parameters
 func DefaultParams() Params {
-	return NewParams()
+	return Params{
+		PermissionedIntegration: false,
```
maurolacy, line 17: 👍🏼

Makefile
```diff
@@ -270,14 +270,14 @@ test-e2e-cache:
...
-	$(MAKE) test-e2e-cache-bcd-consumer-integration
+	$(MAKE) test-e2e-bcd-consumer-integration
 #	$(MAKE) test-e2e-cache-upgrade-v1
 
 clean-e2e:
 	docker container rm -f $(shell docker container ls -a -q) || true
 	docker network prune -f || true
 
-test-e2e-cache-bcd-consumer-integration: start-bcd-consumer-integration
+test-e2e-bcd-consumer-integration: start-bcd-consumer-integration
```
maurolacy, line 280: So, the issue or trade-off here is that this is not cached anymore?

We can still have a no-deps low level 'cache' target here if we want. The 'no-cache' target then calls the 'cache' option, which is there for convenience / user experience.

.github/workflows/ci.yml
```diff
@@ -241,9 +241,6 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: 1.21
-      - name: Start BCD consumer integration containers
-        run: |
-          sudo make -C contrib/images start-bcd-consumer-integration
```
maurolacy, line 246: 🎉

maurolacy (approved):
Looks good to me. Still have to take a look at tests (~and related PR, next~), but approving to unblock.

contrib/images/ibcsim-bcd/Dockerfile
```diff
@@ -18,7 +18,7 @@ ENV PATH $GOPATH/bin:$PATH
 
 ENV GO111MODULE on
 ENV RELAYER_TAG v2.5.2
-ENV BABYLON_SDK_TAG v0.6.0-rc.0
+ENV BABYLON_SDK_VERSION main
```
maurolacy, line 21: What about releasing `v0.6.0-rc.1`. I know it's not semantically version correct, but better than this, which defeats versioning altogether.

x/btcstkconsumer/keeper/msg_server.go
```diff
@@ -19,19 +21,61 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+		)
+		if err := ms.Keeper.RegisterConsumer(ctx, consumerRegister); err != nil {
+			return nil, err
+		}
+	} else {
+		// this is a Cosmos consumer
+
+		// ensure the IBC light client exists
+		sdkCtx := sdk.UnwrapSDKContext(ctx)
+		_, exist := ms.clientKeeper.GetClientState(sdkCtx, req.ConsumerId)
```
maurolacy, line 64: Wouldn't this require that the IBC connection is established **before** registering the consumer? What is the right / required sequence for Consumer registration?

x/btcstkconsumer/keeper/msg_server_test.go
```diff
@@ -22,3 +29,82 @@ func TestMsgServer(t *testing.T) {
...
+		// mock the wasm contract
+		contractAddr := mockSmartContract(t, ctx, babylonApp)
+		// generate a random consumer register
+		consumerRegister = datagen.GenRandomETHL2Register(r, contractAddr.String())
+		// Register the consumer
+		_, err = msgServer.RegisterConsumer(ctx, &types.MsgRegisterConsumer{
+			ConsumerId:                   consumerRegister.ConsumerId,
+			ConsumerName:                 consumerRegister.ConsumerName,
+			ConsumerDescription:          consumerRegister.ConsumerDescription,
+			EthL2FinalityContractAddress: contractAddr.String(),
```
maurolacy, line 98: At some point it would be good to check that the contract is of a particular type / adheres to some particular API.

proto/babylon/zoneconcierge/v1/packet.proto
maurolacy: On second thought, maybe a bit late for this but, what about falling back to / keeping automatic registration for Cosmos chains, in the case of permissionless integration?

maurolacy:
E2e tests are failing in my local:
```
    --- PASS: TestBCDConsumerIntegrationTestSuite/Test1ChainStartup (2.04s)
    --- PASS: TestBCDConsumerIntegrationTestSuite/Test2RegisterAndIntegrateConsumer (129.87s)
    --- PASS: TestBCDConsumerIntegrationTestSuite/Test3CreateConsumerFinalityProvider (13.45s)
    --- PASS: TestBCDConsumerIntegrationTestSuite/Test4RestakeDelegationToMultipleFPs (123.90s)
    --- FAIL: TestBCDConsumerIntegrationTestSuite/Test5ActivateDelegation (45.31s)
    --- FAIL: TestBCDConsumerIntegrationTestSuite/Test6ConsumerFPRewardsGeneration (60.96s)
    --- PASS: TestBCDConsumerIntegrationTestSuite/Test7BabylonFPCascadedSlashing (21.50s)
    --- FAIL: TestBCDConsumerIntegrationTestSuite/Test8ConsumerFPCascadedSlashing (378.03s)
```

maurolacy:
Retrying the tests. The IBC transfer back timeout should be extended still by five seconds or so.

x/btcstaking/keeper/grpc_query.go
```diff
@@ -80,13 +82,12 @@ func (k Keeper) FinalityProvider(c context.Context, req *types.QueryFinalityProv
 				return nil, err
 			}
 			// FPs for consumers are not stored in the voting power table
-			fpResp := types.NewFinalityProviderResponse(fp, 0)
+			fpResp := types.NewFinalityProviderResponse(fp, currBlockHeight)
```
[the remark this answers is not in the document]
SebastianElvis: This was a bug because the query height should be the current block height but not 0

test/e2e/bytecode/babylon_contract.wasm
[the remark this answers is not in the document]
SebastianElvis: Probably we will need to prioritise https://github.com/babylonlabs-io/pm/issues/118, i.e., all such e2e tests should be moved to Babylon SDK. Wdyt?

x/btcstkconsumer/types/expected_keepers.go
SebastianElvis: Good catch, removed channel keeper dependency in btcstkconsumer

Makefile
SebastianElvis: Yes the cache was mostly for accelerating GH action, specifically `ci.yml` pipeline.

However, now I find this pipeline conflates a number of things, including unit tests, e2e for BTC staking, e2e for BTC timestamping, e2e for integration, etc.. Probably we will need to split it into multiple ones. After doing that we might not need this cache, because the docker images will only be used for BTC staking integration e2e. Wdyt?

contrib/images/ibcsim-bcd/Dockerfile
SebastianElvis: Good idea, let's do this

x/btcstkconsumer/keeper/msg_server.go
SebastianElvis: Not really, here we are retrieving IBC *light client* but not connection or channel.

So the user story now becomes: creating an IBC lgiht client for consumer -> registering consumer using this client ID -> passing gov prop if permissioned integration is enforced -> creating IBC connection/channel

This is the approach followed by ICS, as discussed in the RFC/ADR

x/btcstkconsumer/keeper/msg_server_test.go
SebastianElvis: Good idea, added

PR #297, opened 2024-11-22
Update relayer dependency

filippos47:
Related CVE: https://avd.aquasec.com/nvd/2024/cve-2024-32972/
I don't think we use the package, but we can update it to resolve the docker vulnerability alerts.

Edit: It seems to be breaking the lint step.


filippos47:
> this can be closed as we will be getting rid of dependency on relayer in separate ticket

PR #304, opened 2024-11-26
ADR-034: Add highest voted height to finality provider

gitferry:
Implements node side of https://github.com/babylonlabs-io/pm/pull/134 by injecting `HighestVotedHeight` to finality provider


RafilxTenfen (approved):
All the changes looks good to me, can we also add a check in [FuzzAddFinalitySig](https://github.com/babylonlabs-io/babylon/blob/f93e8c1c60c455e6af2a3d1564c763c133d657d3/x/finality/keeper/msg_server_test.go#L106) that verifies the `HighestVotedHeight` is being updated?

gitferry:
> All the changes looks good to me, can we also add a check in [FuzzAddFinalitySig](https://github.com/babylonlabs-io/babylon/blob/f93e8c1c60c455e6af2a3d1564c763c133d657d3/x/finality/keeper/msg_server_test.go#L106) that verifies the HighestVotedHeight is being updated?

Not sure how we can do it because `UpdateFinalityProvider` is an expected interface from `btcstaking` module. It is mocked in test. We have relevant test here https://github.com/babylonlabs-io/babylon/blob/f93e8c1c60c455e6af2a3d1564c763c133d657d3/x/btcstaking/keeper/grpc_query_test.go#L135

PR #306, opened 2024-11-28
feat: improve performance BTC reward distribution

RafilxTenfen:
- Add new structures to track BTC delegation rewards and lazily distribute
- Add new event `BTCDelegationStatus_EXPIRED` for BTC delegations that indicates a BTC delegation is no longer contributing to security 
- Important files diff for review
  -  `proto/babylon/incentive/rewards.proto`
  -  `x/finality/keeper/power_dist_change.go`
  - `x/incentive/keeper/btc_staking_gauge.go`
  - `x/incentive/keeper/grpc_query.go`
  - `x/incentive/keeper/msg_server.go`
  - `x/incentive/keeper/reward_tracker.go`
  - `x/incentive/keeper/reward_tracker_store.go`
  - `x/btcstaking/keeper/btc_delegations.go`
  - `x/btcstaking/keeper/msg_server.go`
  - `x/btcstaking/types/btc_delegation.go`

## Benchmarks

Command to generate pprof `go tool pprof -http=:8080 http://localhost:6060/debug/pprof/profile\?seconds\=300`

- Main branch that iterates over delegations to distribute rewards with 163k BTC delegations, took 15.52s (13.1%) of the time during the 300seconds interval used to generate the pprof 

image

- Current branch `rafilx/improve-performance-btc-dist` with 100k BTC delegations, took 6.73s (1.8%) of the time during the 300seconds interval

image

Note that the percentage of the time used that is ~7x lower should be considered instead of the time, the benchmark from `main` was running in the benchmark server 48 processors and 256GB of RAM while the one in the current branch I run locally with 24 processors and 96GB of RAM.... So the percentage of the time usage instead of seconds should be the one in comparison

> Thanks @Lazar955  for help with benchmark
> The function used to show the difference is the `TallyBlocks` because the `RewardBTCStaking` does not appear in the pprof of the current branch demonstrating that the time spent there is negligible to the overall period


KonradStaniec (commented):
Some initial comments, I am still parsing how it should supposed to work tbh.

It would be great to have some e2e test with pre-defined values i.e setup some scenario with  lets ay 2 fps, each fp having 2 delegations with pre-defined values, known inflation, and split between fps and validators and calculate whether caluclated rewards for stakers makes sense.

x/finality/keeper/msg_server_test.go
```diff
@@ -486,3 +486,9 @@ func TestVerifyActivationHeight(t *testing.T) {
 		blockHeight, activationHeight,
 	).Error())
 }
+
+// func TestBtcDelegationRewards(t *testing.T) {
```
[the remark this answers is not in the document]
RafilxTenfen: This is the integration test, planning to finalize today

x/finality/keeper/power_dist_change.go
```diff
@@ -326,6 +345,20 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
 
+	k.processBtcDelegations(ctx, cacheFpByBtcPkHex, newActiveBtcDels, func(fp, del sdk.AccAddress, sats uint64) {
+		err := k.IncentiveKeeper.BtcDelegationActivated(ctx, fp, del, sats)
+		if err != nil {
+			panic(err)
+		}
+	})
+
+	k.processBtcDelegations(ctx, cacheFpByBtcPkHex, newUnbondedBtcDels, func(fp, del sdk.AccAddress, sats uint64) {
+		err := k.IncentiveKeeper.BtcDelegationUnbonded(ctx, fp, del, sats)
```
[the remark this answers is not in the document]
RafilxTenfen: nop, it is not
If we call twice, it will reduce the voting power twice

x/finality/keeper/power_dist_change.go
```diff
@@ -247,7 +261,11 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 
 		// if this finality provider is slashed, continue to avoid
 		// assigning delegation to it
-		if _, ok := slashedFPs[fpBTCPKHex]; ok {
+		_, isSlashed := slashedFPs[fpBTCPKHex]
+		if isSlashed {
+			if err := k.IncentiveKeeper.FpSlashed(ctx, fp.GetAddress()); err != nil {
```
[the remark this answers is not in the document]
RafilxTenfen: this is handle by the list of actives FPs, that will not receive rewards for the block that they were jailed 
https://github.com/babylonlabs-io/pm/pull/141#discussion_r1879170625

https://github.com/babylonlabs-io/babylon/blob/dd4009df6b7fc6a9a45f56481485f5980f6d6873/x/finality/types/power_table.go#L68

This is the part where it have as active the top N FPs  and these are the ones that are going to receive rewards

x/finality/keeper/power_dist_change.go
```diff
@@ -247,7 +261,11 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 
 		// if this finality provider is slashed, continue to avoid
 		// assigning delegation to it
-		if _, ok := slashedFPs[fpBTCPKHex]; ok {
```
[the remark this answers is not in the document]
RafilxTenfen: not sure on this one 
@SebastianElvis ?

testutil/incentives-helper/keeper.go
```diff
@@ -0,0 +1,40 @@
...
+	*btcstkhelper.Helper
+	IncentivesKeeper *keeper.Keeper
+}
+
+func NewHelper(
+	t testing.TB,
+	btclcKeeper *types.MockBTCLightClientKeeper,
+	btccKeeper *types.MockBtcCheckpointKeeper,
+) *Helper {
+	// ctrl := gomock.NewController(t)
```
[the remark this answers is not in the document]
RafilxTenfen: Removing it

x/finality/keeper/power_dist_change.go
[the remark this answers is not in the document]
RafilxTenfen: > Asking, as there is slight edge case in unbonding in which we can process two unboding events for one delegation.

Can you give more details about this one?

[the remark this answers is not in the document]
RafilxTenfen: okay, this would break the reward tracker 
checking what could I do here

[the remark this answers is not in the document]
RafilxTenfen: The ubonding events for the reward tracker was modified to be processed only to the cases where the btc delegation transactions are already inside `FinalityProviderDistInfo`

https://github.com/babylonlabs-io/babylon/blob/9ee3ba50f54f8f09257aef4a8e925da010c814a4/x/finality/keeper/power_dist_change.go#L295

@gitferry do you see any problem with this?

RafilxTenfen:
> It would be great to have some e2e test with pre-defined values i.e setup some scenario with lets ay 2 fps, each fp having 2 delegations with pre-defined values, known inflation, and split between fps and validators and calculate whether caluclated rewards for stakers makes sense.

I have added [an integration tests with rewards](https://github.com/babylonlabs-io/babylon/blob/01479f330ea98f1813eda655a4a56c5232680874/x/finality/keeper/msg_server_test.go#L496), and there is also a fuzz test that verifies the proportion distributed [FuzzRewardBTCStaking](https://github.com/babylonlabs-io/babylon/blob/aafda528440b6de3b4da254420813e57e7a261e4/x/incentive/keeper/btc_staking_gauge_test.go#L16) 

but I will write an e2e test tomorrow with 2 active fps and 2 BTC delegations, it will be a bit difficult to estimate the rewards, but I think the goal would be to see the difference between the delegations amounts

app/app.go
```diff
@@ -142,7 +142,7 @@ var (
 	// DefaultNodeHome default home directories for the application daemon
 	DefaultNodeHome string
 	// fee collector account, module accounts and their permissions
-	maccPerms = map[string][]string{
+	MaccPerms = map[string][]string{
```
SebastianElvis, line 145: Why is this made public?

proto/babylon/incentive/rewards.proto
```diff
@@ -0,0 +1,76 @@
...
+import "cosmos_proto/cosmos.proto";
+
+option go_package = "github.com/babylonlabs-io/babylon/x/incentive/types";
+
+// FinalityProviderHistoricalRewards represents the cumulative rewards ratio of the
+// finality provider per sat in that period.
+// The period is ommited here and should be part of the key used to store this structure.
+// Key: Prefix + Finality provider bech32 address + Period.
+message FinalityProviderHistoricalRewards {
+    // The cumulative rewards of that finality provider at some specific period
```
SebastianElvis, line 15: Does this sentence mean `cumulative_rewards_per_sat` = cumulative rewards during the given period / amount of Sats delegated to this FP?

testutil/coins/margin.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"testing"
+
+	"cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+)
+
+func RequireCoinsDiffInPointOnePercentMargin(t *testing.T, c1, c2 sdk.Coins) {
```
SebastianElvis, line 11: Can this be moved to `margin_test.go` as this is a test util func?

testutil/incentives-helper/keeper.go
SebastianElvis, line 1: This file seeems to have large overlap with `testutil/btcstaking-helper` or `testutil/helper`. Any reason why we cannot use those?

x/incentive/keeper/reward_tracker.go
```diff
@@ -0,0 +1,295 @@
...
+		return k.subDelegationSat(ctx, fp, del, amtSat)
+	})
+}
+
+// FpSlashed a slashed finality provider should withdraw all the rewards
+// available to it, by iterating over all the delegations for this FP
+// and sending to the gauge. After the rewards are removed, it should
+// delete every rewards tracker value in the store related to this slashed
+// finality provider.
+func (k Keeper) FpSlashed(ctx context.Context, fp sdk.AccAddress) error {
```
SebastianElvis, line 59: Why do we need to enforce withdrawal and remove the BTC delegation reward trackers? Upon slashing this FP will lose all voting power and won't get any reward anyway. It seems secure to keep the rewards in gauge as is

x/incentive/keeper/reward_tracker.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	// remove all the rewards available from the just ended period
+	keysBtcDelRwdTracker := make([][]byte, 0)
+	if err := k.IterateBTCDelegationRewardsTracker(ctx, fp, func(fp, del sdk.AccAddress) error {
+		keysBtcDelRwdTracker = append(keysBtcDelRwdTracker, del.Bytes())
+		return k.CalculateBTCDelegationRewardsAndSendToGauge(ctx, fp, del, endedPeriod)
+	}); err != nil {
+		return err
+	}
```
SebastianElvis, line 73: Yeah tbh I'm not sure if this is necessary, and this iterating process might be some perf hotspot (It iterates over all BTC delegations under a slashed FP).

x/finality/keeper/power_dist_change.go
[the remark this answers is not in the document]
SebastianElvis: No `BtcDels` field is no longer necessary. It is used for tracking the share of every BTC delegation under a given FP. Now this shall be replaced by `storeBTCDelegationRewardsTracker` store

x/incentive/keeper/reward_tracker.go
SebastianElvis, line 59: Do we need to do anything upon a slashed FP? 

- After a FP is slashed BTC delegators will unbond and withdraw anyway, and this will trigger all those lazy executions. So enforce withdrawal does not seem necessary.
- Enforcing withdrawal might cause some edge case, e.g., a FP is slashed which triggers all withdrawal. Then if a BTC delegator wants to withdraw will the code panick since some keys in KV store are removed?

x/incentive/keeper/grpc_query.go
```diff
@@ -27,6 +27,10 @@ func (k Keeper) RewardGauges(goCtx context.Context, req *types.QueryRewardGauges
 
 	// find reward gauge
 	for _, sType := range types.GetAllStakeholderTypes() {
+		if err := k.sendAllBtcDelegationTypeToRewardsGauge(ctx, sType, address); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
```
SebastianElvis, line 32: RPC should not trigger state update at all. Otherwise it's possibel that someone queries this API and the queried node's state root is updated, then in the next block this node won't have a matching state root with any other node.

x/finality/keeper/power_dist_change.go
```diff
@@ -267,9 +284,19 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 		// add all BTC delegations that are not unbonded to the new finality provider
 		for j := range dc.FinalityProviders[i].BtcDels {
 			btcDel := *dc.FinalityProviders[i].BtcDels[j]
```
SebastianElvis, line 286: Would recommend we remove any usage of `BtcDels` under `FinalityProviderDistInfo`. The original purpose of `BtcDels` is to track the share of every BTC delegator under the FP. Now this is replaced by KV stores under `x/incentive` and no need to do this anymore

x/finality/keeper/power_dist_change.go
```diff
@@ -326,6 +350,20 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 		}
 	}
 
+	k.processBtcDelegations(ctx, cacheFpByBtcPkHex, newActiveBtcDels, func(fp, del sdk.AccAddress, sats uint64) {
```
SebastianElvis, line 353: Can we do this inside the `for _, event := range events` loop? This loop was supposed to be the "event bus" that dispatches notifications to relevant modules, including `x/incentive`

SebastianElvis (commented):
That's a comprehensive PR :+1: Still parsing this. Some initial comments

Meanwhile, is it possible to consolidate some of the abstractions, e.g., in `reward_tracker.go` and `reward_tracker_store.go`? The current implementation is a bit hard to follow and some abstractions might not be necessary. See comments

app/app.go
RafilxTenfen: just to import in a test, I can duplicate the code to avoid exposure

proto/babylon/incentive/rewards.proto
RafilxTenfen: well it is the previous FinalityProviderHistoricalRewards.CumulativeRewardsPerSat + the rewards of the given period 

That was the reason I called it cumulative, in the cosmos-sdk they had the equivalent structure that was storing the amount of rewards per share of voting power and they called [CumulativeRewardRatio](https://github.com/cosmos/cosmos-sdk/blob/d9c53bfefc1e75a3c6b09065ea8b3a836cda0d18/x/distribution/types/distribution.pb.go#L97)

testutil/coins/margin.go
RafilxTenfen: if moved to a test, it can't be imported by another package

testutil/incentives-helper/keeper.go
RafilxTenfen: I did use the btcstaking-helper as a base helper inside the incentives structure

proto/babylon/incentive/rewards.proto
```diff
@@ -0,0 +1,76 @@
...
+        (gogoproto.nullable)   = false
+    ];
+}
+
+// BTCDelegationRewardsTracker represents the structure that holds information
+// from the last time this BTC delegator withdraw the rewards or modified his
+// active staked amount to one finality provider.
+// The finality provider address is ommitted here but should be part of the
+// key used to store this structure together with the BTC delegator address.
+message BTCDelegationRewardsTracker {
```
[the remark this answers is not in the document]
RafilxTenfen: Agree :+1:  this could be done, I just think this PR is already to big to introduce new changes

x/incentive/keeper/grpc_query.go
RafilxTenfen: in the query the database does not commit the modified changes to the state 
The same is done by the cosmos-sdk [DelegationRewards](https://github.com/cosmos/cosmos-sdk/blob/46be66ed4205bf109d93704dd14c0822d048972e/x/distribution/keeper/grpc_query.go#L244-L245) query

x/finality/keeper/power_dist_change.go
RafilxTenfen: we could if the events were sorted and never received more than once 

ref https://github.com/babylonlabs-io/babylon/pull/306#discussion_r1886719715

One BTC delegation can generate 2 unbonding events

x/incentive/keeper/reward_tracker.go
RafilxTenfen: This is not strictly necessary, but the FP will not receive rewards anymore under that BTC pk, so we will basically stay with all of those structures inside the database, my idea here was to use everything at once and then prune it 

> Enforcing withdrawal might cause some edge case, e.g., a FP is slashed which triggers all withdrawal. Then if a BTC delegator wants to withdraw will the code panick since some keys in KV store are removed?

If the BTC delegator wants to withdraw after the FP is slashed, it will just not exists that FP,DEL record anymore and will not appear in the iterator, it should not panic

x/finality/keeper/power_dist_change.go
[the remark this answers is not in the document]
gitferry: I think this pr works because the unbonded delegation will not appear in the cache so that unbonding events will processed twice by the incentive module

x/finality/keeper/power_dist_change.go
```diff
@@ -326,6 +350,20 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
+		err := k.IncentiveKeeper.BtcDelegationActivated(ctx, fp, del, sats)
+		if err != nil {
+			panic(err)
+		}
+	})
+
+	k.processBtcDelegations(ctx, cacheFpByBtcPkHex, newUnbondedBtcDels, func(fp, del sdk.AccAddress, sats uint64) {
```
gitferry, line 360: it seems that `cacheFpByBtcPkHex` is sorted twice in each processor. Can we pass a sorted slice here?

gitferry (commented):
Great work! Some inital comments:

app/app.go
RafilxTenfen: rollback to private maccPerms

x/incentive/keeper/reward_tracker.go
```diff
@@ -0,0 +1,295 @@
...
+
+	k.accumulateRewardGauge(ctx, types.BTCDelegationType, del, rewards)
+	return nil
+}
+
+// CalculateBTCDelegationRewards calculates the rewards entitled for this delegation
+// from the starting period cumulative reward and the ending period received as parameter
+// It returns the amount of rewards without decimals (it removes the DecimalAccumulatedRewards).
+func (k Keeper) CalculateBTCDelegationRewards(ctx context.Context, fp, del sdk.AccAddress, endPeriod uint64) (sdk.Coins, error) {
```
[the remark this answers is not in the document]
RafilxTenfen: We could, but it is very helpful to have a separate function that only calculates the rewards instead of calculating and sending them to the gauge

proto/babylon/incentive/rewards.proto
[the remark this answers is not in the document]
RafilxTenfen: updated to `// The cumulative rewards of that finality provider per sat until that period`

x/btcstaking/types/btc_delegation.go
```diff
@@ -59,6 +59,11 @@ func (d *BTCDelegation) GetFpIdx(fpBTCPK *bbn.BIP340PubKey) int {
 	return -1
 }
 
+// Address returns the bech32 fp address
```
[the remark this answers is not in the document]
RafilxTenfen: staker

x/finality/keeper/power_dist_change.go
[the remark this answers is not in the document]
RafilxTenfen: it shouldn't be processed twice, only once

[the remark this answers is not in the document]
RafilxTenfen: I will remove the btcDelegations inside the `FinalityProviderDistInfo` and add a new key inside the `x/finality` that stores the tx hash of already processed unbonded events so that expired with early unbending can be dropped when received by the second time

What do you guys think? @babylonlabs-io/core-dev

x/finality/keeper/power_dist_change.go
RafilxTenfen: `cacheFpByBtcPkHex` is never sorted, it is just a local cache to get FPs by their BTC PK, the slice of BTC delegations that is sorted to make sure it is deterministic

x/finality/keeper/power_dist_change.go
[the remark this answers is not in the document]
gitferry: sorry, typo: `so that unbonding events will processed twice by the incentive module` -> `so that unbonding events will not be processed twice by the incentive module`

I think the current impl is good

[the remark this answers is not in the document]
RafilxTenfen: So, maybe the removal of the BtcDelegations will be done in a further PR 
cc @SebastianElvis

x/finality/keeper/power_dist_change.go
RafilxTenfen: Can this be done in another PR after this one

x/finality/keeper/power_dist_change.go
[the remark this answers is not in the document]
RafilxTenfen: That makes total sense to me, I will make the modifications

proto/babylon/incentive/rewards.proto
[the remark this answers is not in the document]
SebastianElvis: Sure, let's create an issue for now

test/e2e/initialization/node.go
```diff
@@ -355,7 +355,7 @@ func (n *internalNode) initNodeConfigs(persistentPeers []string) error {
 	valConfig.P2P.ExternalAddress = fmt.Sprintf("%s:%d", n.moniker, 26656)
 	valConfig.RPC.ListenAddress = "tcp://0.0.0.0:26657"
 	valConfig.StateSync.Enable = false
-	valConfig.LogLevel = "debug"
+	valConfig.LogLevel = "info"
```
SebastianElvis, line 358: probably we can keep it at debug level for debugging purposes

x/finality/keeper/power_dist_change.go
SebastianElvis: Could we do it in this PR or merge this PR to a feature branch? imo replacing `BtcDels` with the "virtual periods" is the main point of this feature so it's better to have it before merging this feature to `main`

x/finality/keeper/power_dist_change.go
SebastianElvis: I see. Probably a better fix is then distinguishing between two types of unbonded events, but this is certainly out of this PR's scope.

Also, why is `newActiveBtcDels` needed given that a BTC del cannot be active twice?

x/finality/keeper/power_dist_change.go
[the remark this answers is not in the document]
RafilxTenfen: modified at https://github.com/babylonlabs-io/babylon/pull/306/commits/dcee28ef1438a50c6fdc060b4e559993aa3151a3

proto/babylon/incentive/rewards.proto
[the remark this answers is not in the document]
RafilxTenfen: https://github.com/babylonlabs-io/babylon/issues/357

x/incentive/keeper/reward_tracker.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// sendAllBtcRewardsToGauge iterates over all the finality providers associated
+// with the delegator and withdraw the rewards available to the gauge.
+// This creates new periods for each delegation and finality provider.
+func (k Keeper) sendAllBtcRewardsToGauge(ctx context.Context, del sdk.AccAddress) error {
+	return k.iterBtcDelegationsByDelegator(ctx, del, func(del, fp sdk.AccAddress) error {
+		return k.btcDelegationModified(ctx, fp, del)
+	})
+}
```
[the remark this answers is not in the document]
RafilxTenfen: It iterates over all the delegations from one address and acts as if the btc delegation was modified (which was, since BTC rewards are being withdraw) it just doesn't modifies the amount staked

x/finality/keeper/power_dist_change.go
RafilxTenfen: I didn't had in mind that the events were deterministic, my bad 

I can process the reward tracker right at the moment the event are received, I will update tomorrow and remove the `newActiveBtcDels` and `newUnbondedBtcDels`

x/finality/keeper/power_dist_change.go
RafilxTenfen: Removed `BTCDelDistInfo` structure

x/finality/keeper/power_dist_change.go
RafilxTenfen: removed

x/incentive/keeper/reward_tracker.go
```diff
@@ -0,0 +1,295 @@
...
+	newFp := types.NewFinalityProviderCurrentRewards(sdk.NewCoins(), 1, sdkmath.ZeroInt())
+	return newFp, k.setFinalityProviderCurrentRewards(ctx, fp, newFp)
+}
+
+// initializeBTCDelegation creates a new BTCDelegationRewardsTracker from the previous acumulative rewards
+// period of the finality provider and it should be called right after a BTC delegator withdraw his rewards
+// (in our case send the rewards to the reward gauge). Reminder that at every new modification to the amount
+// of satoshi staked from this btc delegator to this finality provider (activivation or unbonding) of BTC
+// delegations, it should withdraw all rewards (send to gauge) and initialize a new BTCDelegationRewardsTracker.
+// TODO: add reference count to keep track of possible prunning state of val rewards
```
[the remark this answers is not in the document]
RafilxTenfen: I will create the issue for the reference count for pruning 
and another one for removing used `GetBTCStakingGauge` which stores the amount of rewards distributed every block https://github.com/babylonlabs-io/babylon/blob/96f5a1fd34d6b09697bbb7fa7b4f5090fc864d2f/x/incentive/keeper/btc_staking_gauge.go#L17

But once it is used, should not be in the database anymore, if anyone wants to know the amounts from old heights it should index events

x/finality/keeper/power_dist_change.go
```diff
@@ -320,6 +351,16 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 			fpDistInfo.AddBTCDel(d)
 		}
 
+		// edge case where we might be processing an unbonded event
```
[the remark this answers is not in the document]
RafilxTenfen: I will add another explicity test case for this as well

x/finality/types/power_table.go
```diff
@@ -155,29 +154,17 @@ func (v *FinalityProviderDistInfo) GetAddress() sdk.AccAddress {
...
-	}
-	v.BtcDels = append(v.BtcDels, btcDelDistInfo)
-	v.TotalBondedSat += btcDelDistInfo.TotalSat
+	v.TotalBondedSat += btcDel.TotalSat
 }
 
-func (v *FinalityProviderDistInfo) AddBTCDelDistInfo(d *BTCDelDistInfo) {
-	v.BtcDels = append(v.BtcDels, d)
-	v.TotalBondedSat += d.TotalSat
+func (v *FinalityProviderDistInfo) UnbondBTCDel(btcDel *bstypes.BTCDelegation) {
```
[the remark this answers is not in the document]
RafilxTenfen: modified functions to sats amounts at https://github.com/babylonlabs-io/babylon/pull/306/commits/33f1b7f0ee42b5ffa6867cf04f171743db2a7885

Issue to add test https://github.com/babylonlabs-io/babylon/issues/361

x/incentive/keeper/reward_tracker.go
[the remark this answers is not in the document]
RafilxTenfen: Reference count https://github.com/babylonlabs-io/babylon/issues/362
`BTCStakingGauge` https://github.com/babylonlabs-io/babylon/issues/363

x/finality/keeper/power_dist_change.go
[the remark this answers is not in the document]
RafilxTenfen: test added https://github.com/babylonlabs-io/babylon/pull/306/commits/20c1c08f5a4a3014ee3789370a641463f35248f9

x/btcstaking/types/btc_delegation.go
```diff
@@ -127,12 +134,18 @@ func (d *BTCDelegation) GetStatus(
...
 		// staking tx's timelock has not begun, or is less than unbonding time BTC
-		// blocks left, or is expired
+		// blocks left
 		return BTCDelegationStatus_UNBONDED
 	}
 
+	// if the endheight is lower than the btc height + unbonding time
+	// the btc delegation should be considered expired
+	if btcHeight+d.UnbondingTime > d.EndHeight {
+		return BTCDelegationStatus_EXPIRED
```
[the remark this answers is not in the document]
RafilxTenfen: good catch, done at https://github.com/babylonlabs-io/babylon/pull/306/commits/5ed4b1c43d41ac8f149252806dfcb0c7c54a43e6

KonradStaniec (approved):
looks good to me 👍 great work pushing it forward :) 

Lets wait for at leas one more approval before merging given size of the pr  :)

x/incentive/keeper/reward_tracker_store.go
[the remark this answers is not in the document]
RafilxTenfen: modified to use collections in most parts

RafilxTenfen:
@SebastianElvis @gitferry Addressed all the comments :rocket:

x/finality/keeper/power_dist_change.go
```diff
@@ -264,24 +283,29 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 			fp.IsJailed = false
 		}
 
-		// add all BTC delegations that are not unbonded to the new finality provider
-		for j := range dc.FinalityProviders[i].BtcDels {
-			btcDel := *dc.FinalityProviders[i].BtcDels[j]
-			if _, ok := unbondedBTCDels[btcDel.StakingTxHash]; !ok {
-				fp.AddBTCDelDistInfo(&btcDel)
+		// process all new BTC delegations under this finality provider
```
gitferry, line 286:
```suggestion
		// process all new BTC delegations under this finality provider
```
Aren't them unbonding?

x/finality/keeper/power_dist_change.go
```diff
@@ -264,24 +283,29 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
+		if fpUnbondedBTCDels, ok := unbondedSatsByFpBtcPk[fpBTCPKHex]; ok {
+			// handle unbonded delegations for this finality provider
+			for _, unbodedSats := range fpUnbondedBTCDels {
+				fp.RemoveBondedSats(unbodedSats)
```
gitferry, line 290: here we process unbondedSats earlier than bondedSats, will it make the sats overflow (uin64 cannot be a negative value)? Better have a test case for it

x/finality/keeper/power_dist_change.go
```diff
@@ -294,30 +318,34 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
-			panic(err) // only programming error
-		}
+		newFP := k.loadFP(ctx, fpByBtcPkHex, fpBTCPKHex)
 		fpDistInfo := ftypes.NewFinalityProviderDistInfo(newFP)
 
 		// add each BTC delegation
-		fpActiveBTCDels := activeBTCDels[fpBTCPKHex]
-		for _, d := range fpActiveBTCDels {
-			fpDistInfo.AddBTCDel(d)
+		fpActiveBTCDels := activedSatsByFpBtcPk[fpBTCPKHex]
```
gitferry, line 336: should use sats other than dels

x/finality/keeper/power_dist_change.go
```diff
@@ -294,30 +318,34 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
+
+		// edge case where we might be processing an unbonded event
+		// from a newly active finality provider in the same slice
+		// of events received.
+		fpUnbondedBTCDels, ok := unbondedSatsByFpBtcPk[fpBTCPKHex]
+		if ok {
+			for _, unbodedSats := range fpUnbondedBTCDels {
+				fpDistInfo.RemoveBondedSats(unbodedSats)
+			}
 		}
```
gitferry, line 349:
```suggestion
		fpUnbondedBTCDels := unbondedSatsByFpBtcPk[fpBTCPKHex]
		for _, unbodedSats := range fpUnbondedBTCDels {
			fpDistInfo.RemoveBondedSats(unbodedSats)
		}
```

gitferry (approved):
Great work on the improvement! Some minor comments, no blocker:

x/finality/keeper/power_dist_change.go
RafilxTenfen: modified to sats

x/finality/keeper/power_dist_change.go
RafilxTenfen: modified to handle active first, test should be added either way in another PR :sweat_smile:

x/finality/keeper/power_dist_change.go
RafilxTenfen: good catch

proto/babylon/btcstaking/v1/btcstaking.proto
```diff
@@ -167,8 +167,8 @@ message BTCDelegatorDelegationIndex {
 
 // BTCDelegationStatus is the status of a delegation.
 // There are two possible valid state transition paths for a BTC delegation:
-// - PENDING -> ACTIVE -> UNBONDED
-// - PENDING -> VERIFIED -> ACTIVE -> UNBONDED
+// - PENDING -> ACTIVE -> UNBONDED -> EXPIRED
+// - PENDING -> VERIFIED -> ACTIVE -> UNBONDED/EXPIRED
```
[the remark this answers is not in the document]
RafilxTenfen: updated

Lazar955 (approved):
Amazing work!

PR #307, opened 2024-11-28
benchmark single list vs iterator for active fps (DNM)

KonradStaniec:
Compare storing active finality provider as list with iterator vs as list in db;

Results for 80fps (on my machine):
```
80 FInality Providers
Get all finality providers
Old: 20243087 ns/op
New: 19692 ns/op // 3 orders of magnitude faster

Get one finality provider
Old: 1072 ns/op
New: 7441 ns/op // 7 times slower
```

I am still not sure whether all data is flushed from memory to the disk, but in setup I do:
- commit which should flush the data from memory to the goleveldb
- write pretty large array to goleveldb which should flush internal lsm memtable to the disk


RafilxTenfen:
benchmark was proved, we can close now and reopen if later is needed

PR #309, opened 2024-12-02
feat(ADR-036): custom withdrawal address

Lazar955:
Adds the ability to specify custom address to separate operational and rewards address

[References ADR-036](https://github.com/babylonlabs-io/pm/pull/133)


RafilxTenfen (approved):
LGTM, but I think we should also add a query to verify which is the current withdrawal addr

x/incentive/keeper/msg_server_test.go
```diff
@@ -63,3 +63,51 @@ func FuzzWithdrawReward(f *testing.F) {
...
+
+		// invoke withdraw and assert consistency
+		resp, err := ms.WithdrawReward(ctx, &types.MsgWithdrawReward{
+			Type:    sType.String(),
+			Address: sAddr.String(),
+		})
+		require.NoError(t, err)
+		require.Equal(t, withdrawableCoins, resp.Coins)
+
+		// ensure reward gauge is now empty
```
gitferry, line 108: should we also assert that the reward gauge is not empty before withdrawal?

x/incentive/keeper/query_delegator.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func (k Keeper) DelegatorWithdrawAddress(goCtx context.Context, req *types.QueryDelegatorWithdrawAddressRequest) (*types.QueryDelegatorWithdrawAddressResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	if req.DelegatorAddress == "" {
+		return nil, status.Error(codes.InvalidArgument, "empty delegator address")
+	}
```
gitferry, line 18: This check is already covered in `sdk.AccAddressFromBech32`

gitferry (approved):
Great work!

PR #311, opened 2024-12-02
Enforce that unbonding tx always has version 2

KonradStaniec:
- Enforces that unbonding tx always have version `2`


Lazar955 (approved):
nice work

PR #317, opened 2024-12-04
Enforce correct parameters version when verifing EOI

KonradStaniec:
- EOI created delegations parameter version should be the some one for given block range


Lazar955 (approved):
nice work!

gitferry (approved):
LGTM!

PR #327, opened 2024-12-06
feat: add the slashing msg handler for op consumer fp

lesterli:
## Summary

Per the comment https://github.com/babylonlabs-io/babylon-contract/pull/92#discussion_r1868538354, this PR adds a new message handler `EquivocationEvidence` that upon the evidence, slash the given FP.

## Test Plan

```
make lint
make test
```


proto/babylon/finality/v1/tx.proto
```diff
@@ -103,3 +107,18 @@ message MsgUnjailFinalityProvider {
...
 // MsgUnjailFinalityProviderResponse defines the Msg/UnjailFinalityProvider response type
 message MsgUnjailFinalityProviderResponse {}
+
+// MsgEquivocationEvidence is the message for handling evidence of equivocation
+message MsgEquivocationEvidence {
+    option (cosmos.msg.v1.signer) = "signer";
+
+    string signer = 1;
+    // evidence is the evidence of equivocation
+    Evidence evidence = 2 [(gogoproto.nullable) = false];
```
SebastianElvis, line 117: Suggest we have all fields in `Evidence` here rather than using a `Evidence` object directly. This is because `Evidence` is a DB object and it's better not conflate DB objects and message objects

proto/babylon/finality/v1/tx.proto
```diff
@@ -103,3 +107,18 @@ message MsgUnjailFinalityProvider {
...
+}
+
+// MsgEquivocationEvidenceResponse is the response for MsgEquivocationEvidence
+message MsgEquivocationEvidenceResponse {
+    // success is true if the slashing was successfully applied
+    bool success = 1;
```
SebastianElvis, line 123: This is actually not needed because if it fails the server will return error anyway

x/finality/keeper/msg_server.go
```diff
@@ -333,3 +333,24 @@ func (ms msgServer) validateActivationHeight(ctx sdk.Context, height uint64) (ui
 	}
 	return activationHeight, nil
 }
+
+// EquivocationEvidence handles the evidence of equivocation message sent from the finality gadget cw contract
+func (ms msgServer) EquivocationEvidence(goCtx context.Context, req *types.MsgEquivocationEvidence) (*types.MsgEquivocationEvidenceResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	evidence := req.Evidence
```
SebastianElvis, line 340: After we have all fields of `Evidence` in `req`, we can validate it, construct the `Evidence` object, and then do the following logic

SebastianElvis (commented):
Thank you!

Also it would be great if we have a fuzz test for this

proto/babylon/finality/v1/tx.proto
lesterli: `Evidence` is defined in [here](https://github.com/babylonlabs-io/babylon/blob/base/consumer-chain-support/proto/babylon/finality/v1/finality.proto#L92-L113), so it is a message object, right?

x/finality/keeper/msg_server.go
lesterli: > we can validate it

[evidence.IsSlashable()](https://github.com/babylonlabs-io/babylon/blob/base/consumer-chain-support/x/finality/types/finality.go#L73-L100) is enough, right?

bap2pecs: since we will not be using `Evidence`, we can't use this function:
image

@SebastianElvis do you mean we should copy the logic in `ValidateBasic` here?

SebastianElvis: Yep, in theory `MsgEquivocationEvidence` can be anything and users might submit invalid `MsgEquivocationEvidence`. It needs to go through the same validation process like this.

lesterli:
> Also it would be great if we have a fuzz test for this

added

x/finality/keeper/msg_server.go
```diff
@@ -333,3 +333,56 @@ func (ms msgServer) validateActivationHeight(ctx sdk.Context, height uint64) (ui
...
 }
+
+// EquivocationEvidence handles the evidence of equivocation message sent from the finality gadget cw contract
+func (ms msgServer) EquivocationEvidence(goCtx context.Context, req *types.MsgEquivocationEvidence) (*types.MsgEquivocationEvidenceResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// validate the evidence
+	if err := validateEvidence(req); err != nil {
+		return &types.MsgEquivocationEvidenceResponse{}, err
+	}
```
SebastianElvis, line 344:
```suggestion
	// validate the evidence
	if err := req.ValidateBasics(); err != nil {
		return &types.MsgEquivocationEvidenceResponse{}, err
	}
```

SebastianElvis (approved):
:+1:

x/finality/keeper/msg_server.go
lesterli: The helper function `validateEvidence` is created in `msg_server.go`, and there is the below comments for `ValidateBasic` in Cosmos SDK
```
// ValidateBasic is deprecated and now facultative.
// Prefer validating messages directly in the msg server.
```

PR #333, opened 2024-12-09
enable more linters from golanci-lint

KonradStaniec:
- enables more linters from golang ci lint
- fixes some low hanging fruits


Lazar955 (approved):
Love the gradual cleanup!

crypto/eots/error.go
```diff
@@ -4,17 +4,17 @@ import (
 	ecdsa_schnorr "github.com/decred/dcrd/dcrec/secp256k1/v4/schnorr"
 )
 
-// ErrorKind identifies a kind of error.  It has full support for errors.Is
+// KindError identifies a kind of error.  It has full support for errors.Is
```
gitferry, line 7:
```suggestion
// KindOfError identifies a kind of error. It has full support for errors.Is
```
Curious why linter is not happy with `ErrorKind`?

gitferry (approved):

crypto/eots/error.go
KonradStaniec: apparently convention for error types is to end name with suffix `Error` (underlying `ecdsa_schnorr.ErrorKind` does not follow this convention)

PR #334, opened 2024-12-09
chore: change IAVL size

Lazar955:
Based on benchmark measurements recommended default size of IAVL cache will be 5000, with fastnode enabled.

image

More context in [confluence](https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/83460126/Babylon+node+benchmark+analysis).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Improvements
 
 - [#316](https://github.com/babylonlabs-io/babylon/pull/316) Add testnet upgrade data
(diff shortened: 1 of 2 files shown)


Lazar955:
@filippos47 @babylonlabs-io/devops tagging if you guys are doing any `sed`'s based on these values

filippos47:
As shared with Lazar offline, we've seen higher memory usages on devnets with high BTC delegation load:
- For nodes with IAVL cache and fastnode disabled, memory went up to 20G+
- For nodes using the default Cosmos SDK settings (781,250 IAVL nodes and fastnode enabled), memory reached 80-90G

Thus, I assume that during this benchmark the cache didn't fill up.

PR #337, opened 2024-12-10
Add scaffold to test replaying the block of the app

KonradStaniec:
Adds initial scaffold for tests which enable full execution of the block by comet-bft utilities:
- `BlockExecutor`

Those are test can be used in various scenarios:
- e2e application test  asserting  that after certain ordered operations, we reached expected state
- deterministic execution tests i.e execute some random operations and replay them and check that app hash matches
- early block execution tests (if we introduce other backends than memdb)


gitferry (approved):
Nice scaffolding!

PR #338, opened 2024-12-10
chore: add print bip 340 to debug pubkey-raw

RafilxTenfen:
- Helps to convert keys from pub key 

```shell
babylond debug pubkey-raw "A2h06lj/GAPW1RvJE3HU9yetMGCR9K7a+tcwwTYT8t7z"
Parsed key as secp256k1
Address: 3565AEBC1D0CF089A4C2B04500EFE5007701604E
JSON (base64): {"type":"tendermint/PubKeySecp256k1","value":"A2h06lj/GAPW1RvJE3HU9yetMGCR9K7a+tcwwTYT8t7z"}
Bech32 Acc: bbnpub1addwnpepqd58f6jcluvq84k4r0y3xuw57un66vrqj862akh66ucvzdsn7t00x5n57a8
Bech32 Validator Operator: bbnvaloperpub1addwnpepqd58f6jcluvq84k4r0y3xuw57un66vrqj862akh66ucvzdsn7t00x4p4spq
Bech32 Validator Consensus: 
BIP340 Hex: 036874ea58ff1803d6d51bc91371d4f727ad306091f4aedafad730c13613f2def3
```

### DIFF
--- a/cmd/babylond/cmd/debug.go
+++ b/cmd/babylond/cmd/debug.go
@@ -7,12 +7,14 @@ import (
 	"strings"
 
 	"github.com/babylonlabs-io/babylon/types"
+	"github.com/cosmos/cosmos-sdk/client/debug"
 	"github.com/cosmos/cosmos-sdk/crypto/keys/ed25519"
 	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
 	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 
-	"github.com/cosmos/cosmos-sdk/client/debug"
-	legacybech32 "github.com/cosmos/cosmos-sdk/types/bech32/legacybech32" //nolint:staticcheck // we do old keys, they're keys after all.
+	// this pkg was deprecated but still needs support
+	bech32 "github.com/cosmos/cosmos-sdk/types/bech32/legacybech32" //nolint:staticcheck
+
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/babylond/cmd/debug.go
```diff
@@ -0,0 +1,130 @@
...
+	if err == nil {
+		return pk, nil
+	}
+
+	pk, err = legacybech32.UnmarshalPubKey(legacybech32.ValPK, pkstr) //nolint:staticcheck // we do old keys, they're keys after all.
+	if err == nil {
+		return pk, nil
+	}
+
+	pk, err = legacybech32.UnmarshalPubKey(legacybech32.ConsPK, pkstr) //nolint:staticcheck // we do old keys, they're keys after all.
```
KonradStaniec, line 110: I must say that phrase `we do old keys, they're keys after all.` sounds like some kind of mantra or something 😅 

Maybe we can have something more descriptive in one place maybe near the import ? (I imagine the problem is that `staticcheck` flags this during linting due to `legacy` in the name ?)

KonradStaniec (approved):

Lazar955 (approved):

cmd/babylond/cmd/debug.go
RafilxTenfen: the `staticcheck` of no lint is due to the `// Deprecated:` note inside `cosmos-sdk/types/bech32/legacybech32/pk.go`

RafilxTenfen: > `we do old keys, they're keys after all.` 

kkkk I just copied the import, but yeah I can write something else just in the import

PR #341, opened 2024-12-12
select params for pre approval flow based on BTC LC tip

KonradStaniec:
Pr changes way how babylon choses parameters for pre-approval flow

- parameters will be chosen by BTC LC tip
- to avoid attacks in which somebody will first send tx to BTC and then report it as EOI, additional validation during activation is added to check that staking transaction is included after the BTC LC tip during delegation creation.


x/btcstaking/keeper/msg_server.go
```diff
@@ -148,6 +148,55 @@ func (ms msgServer) isAllowListEnabled(ctx sdk.Context, p *types.Params) bool {
...
+			return nil, nil, 0, err
+		}
+
+		return &DelegationTimeRangeInfo{
+			StartHeight: 0,
+			EndHeight:   0,
+			TipHeight:   btcTip.Height,
+		}, paramsByHeight, version, nil
+	}
+}
```
RafilxTenfen, line 198: only suggestion
```suggestion
	}
	// staking tx is not included on BTC, retrieve params for the current tip height
	// and return info about the tip
	btcTip := ms.btclcKeeper.GetTipInfo(ctx)

	paramsByHeight, version, err := ms.GetParamsForBtcHeight(ctx, uint64(btcTip.Height))
	if err != nil {
		return nil, nil, 0, err
	}

	return &DelegationTimeRangeInfo{
		StartHeight: 0,
		EndHeight:   0,
		TipHeight:   btcTip.Height,
	}, paramsByHeight, version, nil
}
```

RafilxTenfen (approved):
LGTM :tada:

PR #342, opened 2024-12-13
fix: Non-determinism while jailing

gitferry:
The voting power table is a map which causes random order of processing jailing events. If more than one fps jailed at one block, non-determinism could happen. The PR fixed this by introducing ordered voting power table which retains the order from interating the `votingPowerBbnBlockHeightStore`


x/finality/keeper/liveness.go
```diff
@@ -21,7 +22,15 @@ func (k Keeper) HandleLiveness(ctx context.Context, height int64) {
 	// Iterate over all the finality providers which *should* have signed this block
 	// store whether or not they have actually signed it, identify sluggish
 	// ones, and apply punishment (TBD)
+	// Create a sorted slice of keys for deterministic iteration
+	fpPkHexes := make([]string, 0, len(fpSet))
 	for fpPkHex := range fpSet {
```
KonradStaniec, line 27: From what I remember the elements in `votingPowerBbnBlockHeightStore` are ordered by voting power (sorting and saving happens during block tallying)

What do you think about introducing new function to finality keeper like `GetVotingPowerTableOrdered` that will return something like `[]*FpWithVotingPower` where:
```
type FpWithVotingPower struct {
  fpKeyHex string
  votingPower uint64
}
```
that will just collect values from `votingPowerBbnBlockHeightStore` into list.

This way:
- we will avoid this intermediate `map[string]uint64` steap
- we will avoid this additional sorting and use the same order we have in `votingPowerBbnBlockHeightStore`

gitferry: Good point. I thought the ordering of the iterator is implicit but the benefits are clear if we can assume it is already ordered. Fixed in [68d5397](https://github.com/babylonlabs-io/babylon/pull/342/commits/68d5397caf257058230aea58616ae446b20367d9)

KonradStaniec (approved):

gitferry:
@KonradStaniec Local tested with https://github.com/babylonlabs-io/babylon/pull/344 and worked. Merging it now

PR #343, opened 2024-12-13
add btc staking relevant replay test

KonradStaniec:
Add determinism test for whole staking activation flow pending/verified/active

Next steps:
- finalizing epochs
- finality provider voting and randomness registration


test/replay/btcstaking_test.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/babylonlabs-io/babylon/testutil/datagen"
+	bstypes "github.com/babylonlabs-io/babylon/x/btcstaking/types"
+	"github.com/stretchr/testify/require"
+)
+
+func FuzzCreatingAndActivatingDelegations(f *testing.F) {
```
gitferry, line 12: Seems we can extend it to supporting post-approval?

gitferry (approved):
Brilliant!

test/replay/btcstaking_test.go
KonradStaniec: yup, we can :) though first I want to work towards having votes from finality provider so we have full flow, and afterwards focus on adding different scenarios (pre-approval, post-approval, jailing, no-timestamped randomnes, slashing, unbonding...)

PR #344, opened 2024-12-13
andomnes commit replay test

KonradStaniec:
- Add fuzzing test case showing our jailing is non-deterministic now


KonradStaniec:
replay test shows that it is jailing that it is culprit of non-determinism. After https://github.com/babylonlabs-io/babylon/pull/342 is merged this branch will be re-based on top of main, and then test should pass.

gitferry (approved):
Tested in my local end with the non-determinism fix #342 and worked. Amazing work!

PR #348, opened 2024-12-16
chore: refactory

RafilxTenfen:
- Move `PrivateSigner` to specific package
- Created vars for constant used Mod accounts in appparams

### DIFF
--- a/testutil/signer/private.go
+++ b/testutil/signer/private.go
@@ -1,0 +1,43 @@
+package signer
+
+import (
+	"os"
+
+	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
+	cosmosed "github.com/cosmos/cosmos-sdk/crypto/keys/ed25519"
+
+	"github.com/babylonlabs-io/babylon/app/signer"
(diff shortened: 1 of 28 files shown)


testutil/signer/private.go
```diff
@@ -0,0 +1,43 @@
...
+// SetupTestPrivSigner sets up a PrivSigner for testing
+func SetupTestPrivSigner() (*signer.PrivSigner, error) {
+	// Create a temporary node directory
+	nodeDir, err := os.MkdirTemp("", "tmp-signer")
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		_ = os.RemoveAll(nodeDir)
+	}()
```
Lazar955, line 23:
```suggestion
	nodeDir := t.TempDir()
```

Lazar955: t.TempDir() internally calls t.CleanUp, so a bit cleaner

Lazar955 (approved):
lgtm

testutil/signer/private.go
RafilxTenfen: problem is, I don't have the `*testing.T` there

PR #349, opened 2024-12-16
refactor: use errors.New to replace fmt.Errorf with no parameters

ChengenH:


gitferry:
Don't see any advance of using `errors.New` over `fmt.Errorf`

PR #352, opened 2024-12-17
fix: withdrawal cli for rewards

samricotta:
There has been an issue relative to https://github.com/babylonlabs-io/finality-provider/pull/231 and this pr adds `MsgSetWithdrawAddress` to the necessary codec file plus adds it to `types/msg.go`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@ Update btcstaking module docs to include EOI
 
 - [#324](https://github.com/babylonlabs-io/babylon/pull/324) Fix decrementing
 jailed fp counter
+- [#352](https://github.com/babylonlabs-io/babylon/pull/352) Fix: withdrawal cli 
+for rewards
 
 ### State Machine Breaking
 


x/incentive/client/cli/tx.go
```diff
@@ -65,6 +67,8 @@ func NewSetWithdrawAddressCmd() *cobra.Command {
 				return err
 			}
 
+			clientCtx = clientCtx.WithAccountRetriever(authtypes.AccountRetriever{})
```
gitferry, line 70: why do we need this?

x/incentive/types/codec.go
```diff
@@ -18,9 +18,14 @@ func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
...
 		&MsgWithdrawReward{},
 		&MsgUpdateParams{},
+		&MsgSetWithdrawAddress{},
 	)
 
 	msgservice.RegisterMsgServiceDesc(registry, &_Msg_serviceDesc)
+
+	registry.RegisterImplementations((*sdk.Msg)(nil),
+		&MsgWithdrawReward{},
+	)
```
gitferry, line 28: don't think this is needed but we should reigister the type in `RegisterCodec` above`

gitferry (approved):
Nice fix!

PR #353, opened 2024-12-17
chore: Bump SDK to 0.50.11

vitsalis:


crypto/ecdsa/ecdsa.go
```diff
@@ -34,7 +34,7 @@ func magicHash(msg string) chainhash.Hash {
 	return chainhash.DoubleHashH(bytes)
 }
 
-func Sign(sk *btcec.PrivateKey, msg string) ([]byte, error) {
+func Sign(sk *btcec.PrivateKey, msg string) []byte {
```
vitsalis, line 37: @SebastianElvis can you verify that this change does not lead to any issues? For example, we have to do other validity checks now instead of checking the error?

SebastianElvis: Checked btcec package and this modification is fine -- `SignCompact` shall not return error anyway so they removed the returned error

SebastianElvis (approved):

Lazar955:
This feels a bit hacky, and it might be better to have a separate repo for client package

Lazar955 (approved):

PR #355, opened 2024-12-18
fix: remove unused deps in `.proto` files

lesterli:
## Summary

This PR removes unused dependencies in `.proto` files, addressing the protobuf generation issue outlined in [babylon-contract/issues/101](https://github.com/babylonlabs-io/babylon-contract/issues/101).

## Test Plan

```
make proto-lint
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -419,7 +419,7 @@ protoImage=$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace $(pro
 
 proto-all: proto-gen proto-swagger-gen ## Generate all protobuf related files
 
-proto-gen: ## Generate protobuf files
+proto-gen: proto-lint ## Generate protobuf files
 	@echo "Generating Protobuf files"
 	@$(protoImage) sh ./proto/scripts/protocgen.sh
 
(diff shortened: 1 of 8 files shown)


maurolacy (approved):
LGTM.

Adding the `proto-lint` target as part of protobuf generation or in CI tests looks like a good idea.

In particular since `buf` is failing on the Rust side because of these unused messages (see https://github.com/babylonlabs-io/babylon-contract/issues/101). And in general because it's good practice.

lesterli:
> Adding the `proto-lint` target as part of protobuf generation or in CI tests looks like a good idea.

yes, added

Makefile
```diff
@@ -419,7 +419,7 @@ protoImage=$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace $(pro
 
 proto-all: proto-gen proto-swagger-gen ## Generate all protobuf related files
 
-proto-gen: ## Generate protobuf files
+proto-gen: proto-lint ## Generate protobuf files
```
maurolacy, line 422: 👍🏼

maurolacy:
Yes, we can backport this to main branch as well.

PR #360, opened 2024-12-19
chore(adr-39): Refactor rewarding

gitferry:
Implements https://github.com/babylonlabs-io/pm/pull/145 including:
- Rewards are calculated with a timeout `finality_sig_timeout`
- Rewards should not be assigned to finality providers that have not voted for the block when calculating rewards.


x/incentive/keeper/btc_staking_gauge.go
```diff
@@ -28,6 +29,12 @@ func (k Keeper) RewardBTCStaking(ctx context.Context, height uint64, dc *ftypes.
 		if i >= int(dc.NumActiveFps) {
 			break
 		}
+
+		// skip if finality provider didn't vote
+		if _, ok := voters[fp.BtcPk.MarshalHex()]; !ok {
+			continue
+		}
```
SebastianElvis, line 36: Does this mean that the reward for unvoted FPs remains in the `feeCollector` account? For example, assuming 4 FPs with equal voting power, 3 of them voted for a block, and 1 did not. Then 3/4 of the reward in the reward gauge is distributed to the 3 FPs, whereas the rest 1/3 remains.

Is this expected or we want to distribute the whole reward to voted FPs?

gitferry: Good point. I didn't think of that. I think we should split all the rewards to the voted FPs but what happens if the rewards remain in the `feeCollector` account?

KonradStaniec: >I think we should split all the rewards to the voted FPs

Agree with that. We should split the rewards between all fps that voted on finalized block.

PR #365, opened 2024-12-20
chore(adr-39): Reject out-dated finality votes

gitferry:


Lazar955 (approved):
lgtm 🚀

x/finality/keeper/msg_server.go
```diff
@@ -210,6 +218,24 @@ func (ms msgServer) AddFinalitySig(goCtx context.Context, req *types.MsgAddFinal
 	return &types.MsgAddFinalitySigResponse{}, nil
 }
 
+func (ms msgServer) ShouldAcceptSigForHeight(ctx context.Context, height uint64) (bool, error) {
+	// check if the block is finalized and timestamped
+	indexedBlock, err := ms.GetBlock(ctx, height)
```
KonradStaniec, line 223: when verifying sig we retrieve block:

```
	// the public randomness is good, set the public randomness
	ms.SetPubRand(ctx, req.FpBtcPk, req.BlockHeight, *req.PubRand)
	// verify whether the voted block is a fork or not
	indexedBlock, err := ms.GetBlock(ctx, req.BlockHeight)
	if err != nil {
		return nil, err
	}
```

later for different purposes.

Maybe we can pass indexed block to this function to avoid one db read ?

KonradStaniec (approved):
Non blocking suggestiong to avoid unnecessary read from db.

PR #367, opened 2024-12-22
refactor(btccheckpoint): remove unnecessary index boundary check

lukema95:
Remove unnecessary index boundary check in checkpoint data processing since:
- Loop index `i` is always non-negative in Go
- [`NumberOfParts`](https://github.com/babylonlabs-io/babylon/blob/bf53a4aaa564afae9676a37f998ac0eeb2b96890/x/btccheckpoint/types/msgs.go#L30) is fixed at 2, so `i` will never exceed `math.MaxUint8`


RafilxTenfen:
unverified

PR #370, opened 2024-12-23
chore: use errors.New to replace fmt.Errorf with no parameters

RiceChuan:
use errors.New to replace fmt.Errorf with no parameters


laciferin2024 (commented):
fmt.Errorf : is not deprecated AFAIK and is more convenient and idiomatic for programmatic errors.

Issue #362, opened 2024-12-19
chore: add refence count to `FinalityProviderHistoricalRewards`


RafilxTenfen:
[`FinalityProviderHistoricalRewards`](https://github.com/babylonlabs-io/babylon/blob/3504b67998d5715e00f89f7a017f49f208bd347c/x/incentive/types/rewards.pb.go#L34)  can have a reference count (How many structures are using it) and when it reaches zero it should delete from the database 
Just like it is done in the [cosmos-sdk](https://github.com/cosmos/cosmos-sdk/blob/e76102f885b71fd6e1c1efb692052173c4b3c3a3/x/distribution/keeper/validator.go#L120)

RafilxTenfen:
https://github.com/babylonlabs-io/babylon/pull/306/files#diff-f0b54e0790639c54248093f342f488933a716a6f851fda23afa7648a0e2ddea8R24

PR #371, opened 2024-12-23
fix: slashed fp and BTC rewards tracker

RafilxTenfen:
`EventPowerDistUpdate_SlashedFp` and [`RewardBTCStaking`](https://github.com/babylonlabs-io/babylon/blob/b132d16483d2e12ac68be74a524a4b7121edac65/x/incentive/keeper/btc_staking_gauge.go#L19) are called at different blocks. 
The rewards are being called a few blocks behind and if the slashed fp is pruned in BTC reward tracker data, it fails to find and panics 

On babylond-deployments an fp is being slashed at block height 46, the call of `RewardBTCStaking` happens at Babylon block 48 but is called to distribute rewards of voting power table from block 44 (which didn't had the FP slashed yet) which then checks for fp rewards for the FP that was pruned at the slashing processed event.

```
6:04PM INF finalizing commit of block hash=153E869F7B0056A9CD8208301F2019FFACA2BEFBBB35AEADF70DCA14C15F3972 height=47 module=consensus num_txs=2 root=62B5F025459DEEBE2EE6F5D869C5BCBE903299FF80825FB9DE62551E6B725D52
 ERR CONSENSUS FAILURE!!! err="failed to add fp rewards for btc delegation bbn1wwkaq6z3kdkekm7ltwxng4ftvzux8gzp2xjx8d at height 44: finality provider current rewards not found
``` 

```shell
$~ babylond q btcstaking finality-providers -o json | jq
...
    {
      "description": {
        "moniker": "Finality Provider 0",
        "identity": "",
        "website": "",
        "security_contact": "",
        "details": ""
      },
      "commission": "0.050000000000000000",
      "addr": "bbn1wwkaq6z3kdkekm7ltwxng4ftvzux8gzp2xjx8d",
      "btc_pk": "6ba387c315dc6105ec02b50684593505103a9f13452d3a597c16ac2f22b924dc",
      "pop": {
        "btc_sig_type": "BIP340",
        "btc_sig": "Q3PcfiKT0fwWOVx95rjV+mQvI/juTbYIV84Hfz2NAmQByiHRXz8OWnFbyugJgOrd/TrqR+6BOgtr10bZUFDFbA=="
      },
      "slashed_babylon_height": "46",
      "slashed_btc_height": 131,
      "height": "46",
      "jailed": false,
      "highest_voted_height": 45
    },
```

Basically, this means we are processing the rewards a few blocks down the road after the slash and then it is not possible to prune the data for reward tracker until that fp slashed height is processed.

The fix for the time being: do nothing at the btc reward tracker structures if some fp gets slashed just continue to store that data. In the future during the issue https://github.com/babylonlabs-io/babylon/issues/362 we can properly handle the pruning of slashed finality provider data

### DIFF
--- a/test/e2e/btc_rewards_distribution_e2e_test.go
+++ b/test/e2e/btc_rewards_distribution_e2e_test.go
@@ -402,13 +402,15 @@ func (s *BtcRewardsDistribution) Test7CheckRewards() {
 	// (del2) => 10_00000000
 	fp1RewardGaugePrev, fp2RewardGaugePrev, btcDel1RewardGaugePrev, btcDel2RewardGaugePrev := s.QueryRewardGauges(n2)
 	// wait a few block of rewards to calculate the difference
-	n2.WaitForNextBlocks(2)
 	s.AddFinalityVoteUntilCurrentHeight()
 	n2.WaitForNextBlocks(2)
 	s.AddFinalityVoteUntilCurrentHeight()
 	n2.WaitForNextBlocks(2)
 	s.AddFinalityVoteUntilCurrentHeight()
 	n2.WaitForNextBlocks(2)
+	s.AddFinalityVoteUntilCurrentHeight()
+	n2.WaitForNextBlocks(2)
+	s.AddFinalityVoteUntilCurrentHeight()
 
 	fp1RewardGauge, fp2RewardGauge, btcDel1RewardGauge, btcDel2RewardGauge := s.QueryRewardGauges(n2)
 
--- a/test/e2e/btc_rewards_distribution_e2e_test.go
+++ b/test/e2e/btc_rewards_distribution_e2e_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 
+	"github.com/babylonlabs-io/babylon/crypto/eots"
 	"github.com/babylonlabs-io/babylon/test/e2e/configurer"
 	"github.com/babylonlabs-io/babylon/test/e2e/configurer/chain"
 	"github.com/babylonlabs-io/babylon/testutil/coins"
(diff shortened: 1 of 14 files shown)


x/finality/keeper/power_dist_change.go
```diff
@@ -320,6 +323,11 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 	// sort new finality providers in activeBTCDels to ensure determinism
```
KonradStaniec, line 323: comment about a line in which we set `fp.IsSlashed = true`  i.e
```
		if isSlashed {
			fp.IsSlashed = true
			continue
		}
```
shouldn't we also set `fp.TotalBondedSats = 0` there ? this would be inline with previous logic in which we set
`fp.TotalBondedSats = 0` at the begininng and the loop and later do not add any delegations this fp so effectively leaving it with `0`

x/finality/keeper/power_dist_change.go
```diff
@@ -320,6 +323,11 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
 	fpActiveBtcPkHexList := make([]string, 0, len(activedSatsByFpBtcPk))
 	for fpBTCPKHex := range activedSatsByFpBtcPk {
+		// if the fp was slashed, should not even be added to the list
+		_, isSlashed := slashedFPs[fpBTCPKHex]
```
KonradStaniec, line 327: hmm I think we should add it the active list and and later replicate the logic from above loop i.e 
```
		if isSlashed {
			fp.IsSlashed = true
			continue
		}
```

as otherwise we won't set jailed flag for new slashed fp. (not sure this is criticial, but this is definitly inconsisent)

x/finality/keeper/power_dist_change.go
RafilxTenfen: we can, but that would not modify the `newDc` since we are calling `continue` if the fp is slashed

x/finality/keeper/power_dist_change.go
RafilxTenfen: good point

x/finality/keeper/power_dist_change.go
```diff
@@ -233,10 +233,13 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 			types.EmitSlashedFPEvent(sdkCtx, typedEvent.SlashedFp.Pk)
 			fpBTCPKHex := typedEvent.SlashedFp.Pk.MarshalHex()
 			slashedFPs[fpBTCPKHex] = struct{}{}
-			fp := k.loadFP(ctx, fpByBtcPkHex, fpBTCPKHex)
-			if err := k.IncentiveKeeper.FpSlashed(ctx, fp.Address()); err != nil {
-				panic(err)
-			}
```
RafilxTenfen, line 239: This was the important part

test/e2e/btc_rewards_distribution_e2e_test.go
```diff
@@ -442,7 +445,50 @@ func (s *BtcRewardsDistribution) Test7CheckRewards() {
 	s.NotEmpty(del2DiffRewardsStr)
 }
 
-// TODO(rafilx): Slash a FP and expect rewards to be withdraw.
+// Test8SlashFp slashes the finality provider, but should continue to produce blocks
+func (s *BtcRewardsDistribution) Test8SlashFp() {
```
RafilxTenfen, line 449: test that verifies by slashing it doesn't panic the rewards anymore

x/finality/keeper/power_dist_change.go
```diff
@@ -332,6 +340,14 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 		newFP := k.loadFP(ctx, fpByBtcPkHex, fpBTCPKHex)
```
KonradStaniec, line 340: I think you are righ about about in above comments (https://github.com/babylonlabs-io/babylon/pull/371/files#r1896052253)

Though one thing that worries me is that we should probably ignore fp here if he is already slashed as even though we do not accept delegations for slashed finality providers, it is still possible to generate voting power events for them:
- it is still possible to activate delegation which was created without inclusion proof
- it is sittl possible to unbond delegation which would create unbonding event

Could you try to add some tests for this cases to verify that it is handled properly currently and add fixe if they dont ? (basically we do not want for slashed fp to ever end up in voting power set ever again)

KonradStaniec (approved):
Fix looks good 👍 

Added one comment about thing we are possibly missing.

Lazar955 (approved):

gitferry (commented):
Maybe an alternative solution is to add a condition to break the loop if the fp.Slashed in https://github.com/babylonlabs-io/babylon/blob/main/x/finality/types/power_table.go#L77?

RafilxTenfen:
> Maybe an alternative solution is to add a condition to break the loop if the fp.Slashed in https://github.com/babylonlabs-io/babylon/blob/main/x/finality/types/power_table.go#L77?

I totally agree

x/finality/keeper/power_dist_change.go
RafilxTenfen: I am verifying if they are slashed, in the creation of `activedSatsByFpBtcPk` list, but yeah adding the check as Gai [suggested](https://github.com/babylonlabs-io/babylon/pull/371#pullrequestreview-2521909396) seems to make a lot of sense to me

PR #374, opened 2025-01-02
fix block tally in finality module

KonradStaniec:
https://github.com/babylonlabs-io/babylon/pull/350 Introduced regression due to change from if/else to switch. `break` which earlier was jumping out of the loop, after change would only exit the `switch` statement. 

Fix:
- add `finalizationLoop` label and use it during `break`
- add proper test to check that finalization in consecutive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
 
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


gitferry (approved):
Nice fix!

RafilxTenfen (approved):

RafilxTenfen (commented):
~~Should we also remove the continue?~~


```
		case fpSet == nil && !ib.Finalized:
			// does not have finality providers, non-finalised: not finalisable,
			// increment the next height to finalise and continue
			k.setNextHeightToFinalize(ctx, ib.Height+1)
			continue
```

edit: Ops, it goes to the `for` to continue, understood

PR #377, opened 2025-01-03
chore: add new query to return next block to be rewarded

RafilxTenfen:
helps to debug


KonradStaniec (commented):
Wdyt about making this a metric ? 

It feel a bit weird to introduce whole query, without other use case than debugging.

PR #378, opened 2025-01-03
fix: `HandleRewarding` gaps of unfinalized blocks

RafilxTenfen:
Prior to this fix, if we had an unfinalized block it would stuck to it as a next block to be rewarded

By using `continue` instead of `break`, if there is any finalized block in the interval (`nextBlockToBeRewarded` and `targetHeight`) it will give out rewards and update the next block to be rewarded

### DIFF
--- a/x/finality/keeper/rewarding.go
+++ b/x/finality/keeper/rewarding.go
@@ -30,10 +30,9 @@ func (k Keeper) HandleRewarding(ctx context.Context, targetHeight int64) {
 		if err != nil {
 			panic(err)
 		}
-		if !block.Finalized {
-			break
+		if block.Finalized {
+			k.rewardBTCStaking(ctx, height)
 		}
-		k.rewardBTCStaking(ctx, height)
 		nextHeightToReward = height + 1
 	}
 
--- a/x/finality/keeper/rewarding.go
+++ b/x/finality/keeper/rewarding.go
@@ -9,6 +9,7 @@ import (
 	"github.com/babylonlabs-io/babylon/x/finality/types"
 )
 
+// HandleRewarding calls the reward to stakers if the block is finalized
 func (k Keeper) HandleRewarding(ctx context.Context, targetHeight int64) {
 	// rewarding is executed in a range of [nextHeightToReward, heightToExamine]
 	// this is we don't know when a block will be finalized and we need ensure
--- a/x/finality/keeper/rewarding.go
+++ b/x/finality/keeper/rewarding.go
@@ -30,9 +30,10 @@ func (k Keeper) HandleRewarding(ctx context.Context, targetHeight int64) {
 		if err != nil {
 			panic(err)
 		}
-		if block.Finalized {
-			k.rewardBTCStaking(ctx, height)
+		if !block.Finalized {
+			continue
 		}
+		k.rewardBTCStaking(ctx, height)
 		nextHeightToReward = height + 1
 	}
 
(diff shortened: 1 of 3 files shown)


gitferry (approved):
Lgtm!

x/finality/keeper/rewarding_test.go
```diff
@@ -108,3 +109,57 @@ func FuzzHandleRewarding(f *testing.F) {
...
+	require.NoError(t, err)
+	fKeeper.SetVotingPower(ctx, fpPK.MustMarshal(), 1, 1)
+
+	// starts rewarding at block 1
+	fKeeper.SetNextHeightToReward(ctx, 1)
+
+	fKeeper.SetBlock(ctx, &types.IndexedBlock{
+		Height:    1,
+		AppHash:   datagen.GenRandomByteArray(r, 32),
+		Finalized: false,
```
gitferry, line 135: I would set it to `true` to see the `NextHeightToReward` to grow to `2`

gitferry (approved):
Lgtm!

x/finality/keeper/rewarding_test.go
RafilxTenfen: modified

x/finality/keeper/rewarding.go
```diff
@@ -30,7 +31,7 @@ func (k Keeper) HandleRewarding(ctx context.Context, targetHeight int64) {
 			panic(err)
 		}
 		if !block.Finalized {
-			break
+			continue
```
KonradStaniec, line 34: hmm the only concern I would have is that in case finality lags behind many blocks, this operation would be quite expensive i.e if we have 100 unfinalized blocks this will iterate over all of them. 

This can be a bit expensive. Of course if we are lagging 100 blocks it means something is happening already so maybe this non issue? 

In our case of all jailed fps, where there was no finalization for 5k blocks , this would iterate over all of them. 

Wdyt @RafilxTenfen @gitferry ?

KonradStaniec (approved):

x/finality/keeper/rewarding.go
RafilxTenfen: yeap, that is true 

but it would be updated to the closest finalized block height once we have it

gitferry: Good point. It would mitigate the issue if we increase `nextHeightToReward` regardless block finalization (see [commit 10b90f8](https://github.com/babylonlabs-io/babylon/pull/378/commits/10b90f827a1f2f426c22b0e108f73f0a167d52ea)). If we do so, it means that we will never reward blocks back if they are not finalized. I think it also make sense to not reward those late votes rewardless block finalization

KonradStaniec: >Good point. It would mitigate the issue if we increase nextHeightToReward regardless block finalization (see [commit 10b90f8](https://github.com/babylonlabs-io/babylon/pull/378/commits/10b90f827a1f2f426c22b0e108f73f0a167d52ea)). If we do so, it means that we will never reward blocks back if they are not finalized. I think it also make sense to not reward those late votes rewardless block finalization

Hmm this seems like protocol change (not rewarding those late votes) it definitely should not be done in this bug fix pr.

gitferry: Agreed. Let's merge this pr first and record this as an issue. Then bring it to the team for better visibility. Imo, we should make decision/implementation quick as it would cause breaking change

PR #380, opened 2025-01-04
Fix: fix error messages in LoadWrappedFilePV to include file paths

meetrick:
Improved error messages in LoadWrappedFilePV by appending file paths.
When this code is applied, it allows users to identify the root cause of errors more easily.
Even though it may seem basic, this is a common point of confusion for many operators.

- The code has been successfully built and tested in the local environment.


RafilxTenfen:
func removed

PR #383, opened 2025-01-06
Fix: Fix broken BabylonChain links in README.md to point to BabylonLabs

meetrick:
Updated outdated links referencing babylonchain.io to babylonlabs.io. Adjusted documentation and user guide URLs to ensure correct redirection. Verified that all modified links are functional and accurate.


RafilxTenfen:
fixed already

Issue #336, opened 2024-12-10
Separate BLS key from `priv_validator_key.json`


gitferry:
Should consider utilizing some keyring for bls keys storage like ethereum bls key store - https://github.com/ethereum/ercs/blob/master/ERCS/erc-2335.md, or most recent versions of comet bft have support for bls keys - https://github.com/cometbft/cometbft/tree/main/crypto/bls12381

gitferry:
This should be done in two steps:
1. Naively separate BLS key from `priv_validator_key.json`
2. Store the BLS key securely following https://github.com/ethereum/ercs/blob/master/ERCS/erc-2335.md

gitferry:
Some context of why and how BLS key is used can be found [here](https://github.com/babylonlabs-io/babylon/blob/5f8af8ced17d24f3f0c6172293cd37fb3d055807/x/checkpointing/README.md).

The BLS key should be treated as important as the ed25519 key used in Comet consens. Currently it is embeded in the `priv_validator_key.json` and validators should use `babylond create-bls-key` to generate the BLS key and bond with the ed25519 key generated in pior. However, cosmos tooling (e.g., tmkms and horcrux) cannot recognize the bls key in the `priv_validator_key.json`, so now we decided to separate it from `priv_validator_key.json`

In order for secure management of the BLS key, after separation, the BLS key should be kept following standard proposed by Ethereum https://github.com/ethereum/ercs/blob/master/ERCS/erc-2335.md so that it can work with related toolings.

dongsam:
Hi, I am an engineer at B-Harvest. Our team plans to work on this issue, and we are considering proceeding as follows. If there are any issues or alternative suggestions, I’d appreciate it if you could share them. 

### 1. Preserve `priv_validator_key.json` in its original form for compatibility with consensus-related logic and tools

- Modify [`WrappedFilePVKey`](https://github.com/babylonlabs-io/babylon/blob/a3b749d7cd6f9d46fc484508bba9aea719eb94b2/privval/file.go#L49) to ensure it does not overwrite the existing `priv_validator_key.json`.

```diff
{
-  "acc_address": "bbn17m08...",
  "address": "78C5FF6E441E3...",
  "pub_key": {
    "type": "tendermint/PubKeyEd25519",
    "value": "IAnqptCTD3pUKhnt...="
  },
  "priv_key": {
    "type": "tendermint/PrivKeyEd25519",
[...]

PR #396, opened 2025-01-10
feat: BLS Key Separation and ERC2335 Implementation

wnjoon:
## Description

This PR was created from epic #336, separates BLS key from priv_validator_key.json, and applies ERC2335.

There are still issues in CI that are being resolved, but I'm uploading a PR to first discuss the changed structure and ideas.

### ERC2335 structure

- Use [keystorev4](github.com/wealdtech/go-eth2-wallet-encryptor-keystorev4) with [prysm](https://github.com/prysmaticlabs/prysm) as reference
- Add ERC2335 encryption/decryption and password generation functions

### Change WrappedFilePV to include BLS structure and use cometbft's FilePV directly

```go
type WrappedFilePVKey struct {
	CometPVKey privval.FilePVKey
	BlsPVKey   BlsPVKey
}
```

### BLS separation

- At the time of `babylond init`, only priv_validator_key.json is created.
- `babylond create-bls-key` creates a BLS key with the password entered as the prompt and encrypts it into an ERC2335 structure.
- The generated BLS key in ERC2335 format is saved as `bls_key.json`, and the password is saved as `bls_password.txt`.

An example of a `bls_key.json` is below:

```json
{
  "crypto": {
    "checksum": {
      "function": "sha256",
      "message": "d291102754ece0a5ab9d4c3c75f452069d192df6c90d616e466d7e7884f41b3a",
      "params": {}
    },
    "cipher": {
      "function": "aes-128-ctr",
      "message": "e638d3f04b803c3ec934c31749ed20c1ffcd48ab481b4fad6b303aa2e7ff991c",
      "params": {
        "iv": "473e315746ec9997f0d0e423f73e2d3d"
      }
    },
    "kdf": {
      "function": "pbkdf2",
      "message": "",
      "params": {
        "c": 262144,
        "dklen": 32,
        "prf": "hmac-sha256",
        "salt": "3f1b139f3663082f582acc6d48fd9bf8caf4ea9c4e51c6ef75c9bc5b5c04ee9e"
      }
    }
  },
  "version": 4,
  "uuid": "",
  "path": "",
  "pubkey": "a28b189e76e8d61461811eec22bc8d5c926bf1276afde5344a3b9d3a2769a1a61ea42471a867d4811a85640dfde8fed702d81bcc9f56494ec65471151fac6e3c262541245f1b16c9e22000eef2142c43f6ad5ca8c31fc16081261cc54d3f034c",
  "description": "bbn196xcr99y5eehdg6lr38u8re0j6xeq0twup5rl4"
}
```

An example of a `priv_validator_key.json` with BLS separated is below:

```json
{
  "address": "3141C6C8EF5FDAE49931AE2F9EB295F5C9E2212A",
  "pub_key": {
    "type": "tendermint/PubKeyEd25519",
    "value": "CwCehdhn8DmcVTKn4I1RHzlmur5LbEZxSEr+zvf7Ifo="
  },
  "priv_key": {
    "type": "tendermint/PrivKeyEd25519",
    "value": "HIHCMsU9+J3FjEU7gooSed5moKRjar6TfLa3Nb1CSCULAJ6F2GfwOZxVMqfgjVEfOWa6vktsRnFISv7O9/sh+g=="
  }
}
```

### Change DelegatorAddress to be stored in the description field of ERC2335 structure instead of storing it in `priv_validator_key.json`.

- DelegatorAddress is stored in the description field of the `bls_key.json` file in erc2335 format.

```go
type BlsPVKey struct {
	PubKey  bls12381.PublicKey  `json:"bls_pub_key"`
	PrivKey bls12381.PrivateKey `json:"bls_priv_key"`
	DelegatorAddress string
        // ...
}
```

## Subsequent tasks

1. Ensure the BLS key file is created atomically with priv_validator_key.json
2. BLS key non-validation logic for non-validator
3. Remove DelegatorAddress from BlsPV structure and obtain it from GetValidatorByConsAddr function of the Epoch(Staking) module.
4. Modified to not keep `WrappedFilePV` since LastSignState is not used

```go
type WrappedFilePV struct {
	Key           WrappedFilePVKey
	LastSignState privval.FilePVLastSignState
}
```

5. Add option to use --bls-password flag when creating bls key


crypto/erc2335/erc2335.go
```diff
@@ -0,0 +1,83 @@
...
+	Version     uint                   `json:"version"`
+	UUID        string                 `json:"uuid"`
+	Path        string                 `json:"path"`
+	Pubkey      string                 `json:"pubkey"`
+	Description string                 `json:"description"`
+}
+
+// wonjoon: encrypt key pair to erc2335 keystore
+// available to handle all keys in []byte format
+func Encrypt(privKey, pubKey []byte, password string) ([]byte, error) {
```
gitferry, line 25: why not return the results of `encrypto.Encrypt` directly? Seems constructing `Erc2335KeyStore` is not relevant to `Encrypt`

app/signer/private.go
```diff
@@ -15,17 +15,27 @@ type PrivSigner struct {
...
+
+	cometPV := cmtprivval.LoadFilePV(pvKeyFile, pvStateFile)
+	blsPV := privval.LoadBlsPV(blsKeyFile, blsPasswordFile)
+
+	wrappedPV := &privval.WrappedFilePV{
+		Key: privval.WrappedFilePVKey{
+			CometPVKey: cometPV.Key,
+			BlsPVKey:   blsPV.Key,
+		},
+		LastSignState: cometPV.LastSignState,
```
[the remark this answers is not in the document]
wnjoon: As I mentioned before, It will be done in the next PR. 
Even though it can be started now, the number of codes that should be reviewed might be increased.

client/flags/flags.go
```diff
@@ -0,0 +1,5 @@
+package flags
+
+const (
+	FlagBlsPassword = "bls-password"
```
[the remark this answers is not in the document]
wnjoon: while `create-bls-key` cmd, bls-password flag will be used optionally to get password from command line. And the flag will be set inside `create-bls-key.go` statically since only this code will use that flag. So that file will be removed and no longer be used.

cmd/babylond/cmd/create_bls_key.go
```diff
@@ -57,28 +53,13 @@ $ babylond create-bls-key %s1f5tnl46mk4dfp4nx3n2vnrvyw2h2ydz6ykhk3r --home ./
...
 func CreateBlsKey(home string, addr sdk.AccAddress) error {
-	nodeCfg := cmtconfig.DefaultConfig()
-	keyPath := filepath.Join(home, nodeCfg.PrivValidatorKeyFile())
-	statePath := filepath.Join(home, nodeCfg.PrivValidatorStateFile())
 
-	pv, err := LoadWrappedFilePV(keyPath, statePath)
-	if err != nil {
-		return err
-	}
+	blsCfg := privval.DefaultBlsConfig()
```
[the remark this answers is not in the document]
wnjoon: No, only the default config is used. 
This is a direct imitation of the method used in FilePV in cometbft, and this structure is unnecessary in BLS, which uses the same directory path as the Comet file.

cmd/babylond/cmd/create_bls_key.go
```diff
@@ -57,28 +53,13 @@ $ babylond create-bls-key %s1f5tnl46mk4dfp4nx3n2vnrvyw2h2ydz6ykhk3r --home ./
...
+	keyPath := filepath.Join(home, blsCfg.BlsKeyFile())
+	passwordPath := filepath.Join(home, blsCfg.BlsPasswordFile())
```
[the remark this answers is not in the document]
wnjoon: No, there is no separate flag for BLS keys and each file name will be used as a constant value. 

crypto/erc2335/erc2335.go
wnjoon: Currently, we are storing the DelegatorAddress in the Description field of the BlsPV structure, which is in ERC2335 format. So when calling Encrypt, we are temporarily returning an Erc2335KeyStore. This part is planned to be comprehensively fixed in the next PR.

gitferry (commented):
Thank for the answers. One thing as I mentioned in the pr is the change the target branch. Other than that the code lgtm! Will approve once it's ready for review. Also, let's wait for review from @KonradStaniec.

How is CI going (I don't know why I can's see the ci results)?

wnjoon:
Commit summary [1f14b31](https://github.com/babylonlabs-io/babylon/pull/396/commits/1f14b31db3dc3aef8655e416aae90ca12e0e8557)

- Added the bls-password flag to be optionally provided when executing the create-bls-key CLI.
- Removed code related to the DefaultConfig of BLS and updated the implementation accordingly.
- Deleted unused flags.
- Renamed the function for checking the existence of a file directory path (IsValidFilePath -> EnsureDirs).
- Fixed linting errors.

Additionally, in the CI pipeline, there is an issue with a specific E2E test: test-e2e-cache-upgrade-v1. The problem appears to stem from the recent change requiring the BLS key to be created exclusively during the create-bls-key process. As a result, the node does not start if the key is missing. We plan to address this issue further.

Furthermore, aside from the test-e2e-cache-ibc-transfer test, which requires DockerHub login, all other E2E tests, and unit tests have been successfully executed locally.

crypto/erc2335/erc2335.go
```diff
@@ -0,0 +1,83 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/pkg/errors"
+	keystorev4 "github.com/wealdtech/go-eth2-wallet-encryptor-keystorev4"
+)
+
+type Erc2335KeyStore struct {
+	Crypto      map[string]interface{} `json:"crypto"`
```
KonradStaniec, line 15: Some doc for this field would be nice like what does this map hold (as type ` map[string]interface{}` does not tell much to the reader)

crypto/erc2335/erc2335.go
```diff
@@ -0,0 +1,83 @@
...
+
+	return json.Marshal(keystoreJSON)
+}
+
+func LoadKeyStore(filePath string) (Erc2335KeyStore, error) {
+	var keystore Erc2335KeyStore
+
+	keyJSONBytes, err := os.ReadFile(filePath)
+	if err != nil {
+		return Erc2335KeyStore{}, err
```
KonradStaniec, line 51: lets wrap errors here with some additional useful info given that this is public function which is pretty low level

crypto/erc2335/erc2335.go
```diff
@@ -0,0 +1,83 @@
...
+// available to handle all keys in []byte format
+func Encrypt(privKey, pubKey []byte, password string) ([]byte, error) {
+	if privKey == nil {
+		return nil, errors.New("private key cannot be nil")
+	}
+
+	encryptor := keystorev4.New()
+	cryptoFields, err := encryptor.Encrypt(privKey, password)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to encrypt private key")
```
KonradStaniec, line 33: usually in our codebases we are wrapping our errors with `fmt.Errorf("error msg: %w", err)` (which imo looks nicer though this is highly personal opinion), any reason for using `errors` ?

crypto/erc2335/erc2335.go
```diff
@@ -0,0 +1,83 @@
...
+func LoadKeyStore(filePath string) (Erc2335KeyStore, error) {
+	var keystore Erc2335KeyStore
+
+	keyJSONBytes, err := os.ReadFile(filePath)
+	if err != nil {
+		return Erc2335KeyStore{}, err
+	}
+
+	if err := json.Unmarshal(keyJSONBytes, &keystore); err != nil {
+		return Erc2335KeyStore{}, err
```
KonradStaniec, line 55: lets wrap the error here

crypto/erc2335/erc2335.go
```diff
@@ -0,0 +1,83 @@
...
+func LoadPaswordFromFile(filePath string) (string, error) {
+	password, err := os.ReadFile(filePath)
+	return string(password), err
+}
+
+func CreateRandomPassword() string {
+	password := make([]byte, 32)
+	_, err := rand.Read(password)
+	if err != nil {
+		panic(err)
```
KonradStaniec, line 80: this is a bit risky in such low level public operation which does not know context of the application.

Maybe:
- if it is only for testing, lets make it private
- it is supposed to be public, lets return error

privval/bls.go
```diff
@@ -0,0 +1,140 @@
...
+	return password
+}
+
+// Save bls key using password
+// Check both paths of bls key and password inside function
+func (k *BlsPVKey) Save(password, addr string) {
+	// encrypt the bls12381 key to erc2335 type
+	erc2335BlsPvKey, err := erc2335.Encrypt(k.PrivKey, k.PubKey.Bytes(), password)
+	if err != nil {
+		panic(err)
```
KonradStaniec, line 104: seems weird to be panicking in low level library code. Lets wrap and return errors ? (and panic at callsites if necessary)

KonradStaniec (approved):
few nits here and there, but in general LGTM! great progress 👍

crypto/erc2335/erc2335.go
Lazar955, line 25: nit: proper go doc

privval/bls.go
wnjoon: The logic for saving BLS keys to files is written to be consistent with [FilePV in cometbft](https://github.com/cometbft/cometbft/blob/8f861366f08c96c5fc43d4fc0c08fa7c14857e66/privval/file.go#L135). 

In addition, since Save is not executed during state transition, but only when setting keys before the chain is executed, the risk of panic should not be great. 

Nevertheless, if necessary, I will rewrite it to return an error. Please give me your opinion.

wnjoon:
Commit Summary [9dd45f3](https://github.com/babylonlabs-io/babylon/pull/396/commits/9dd45f35b08314b5c93bc2fdcd541344ffefc3db)

- Add comments based on go doc in erc2335.go and bls.go
- Remove unused function in erc2335.go
  - Replace not to generate random password during unit tests.
  - Remove the save/load password function in erc2335 and modify related codes in bls.go
- Wrapping errors to use fmt.Errorf in privval package

crypto/erc2335/erc2335.go
wnjoon: [9dd45f3](https://github.com/babylonlabs-io/babylon/pull/396/commits/9dd45f35b08314b5c93bc2fdcd541344ffefc3db)

crypto/erc2335/erc2335.go
wnjoon: [9dd45f3](https://github.com/babylonlabs-io/babylon/pull/396/commits/9dd45f35b08314b5c93bc2fdcd541344ffefc3db)

crypto/erc2335/erc2335.go
wnjoon: [9dd45f3](https://github.com/babylonlabs-io/babylon/pull/396/commits/9dd45f35b08314b5c93bc2fdcd541344ffefc3db)

crypto/erc2335/erc2335.go
wnjoon: [9dd45f3](https://github.com/babylonlabs-io/babylon/pull/396/commits/9dd45f35b08314b5c93bc2fdcd541344ffefc3db)

crypto/erc2335/erc2335.go
wnjoon: [9dd45f3](https://github.com/babylonlabs-io/babylon/pull/396/commits/9dd45f35b08314b5c93bc2fdcd541344ffefc3db)

privval/bls.go
gitferry: I'm fine with either way

wnjoon:
Here is the plan for the second PR:

This second PR is expected to be set R4R until next week.

### Remove WrappedFilePV:

Since the LastSignState field has been removed, the WrappedFilePV structure is no longer necessary. Methods that previously relied on WrappedFilePV will be updated to directly reference FilePV and BlsPV from CometBFT.

### Ensure Atomic Creation of BLS Key and priv_validator_key.json:

Further discussion is needed on “ensuring the atomic creation of the BLS key file along with priv_validator_key.json and removing the create-bls-key command”.

- Since priv_validator_key.json can be automatically generated without running init due to Comet’s behavior, it may be necessary to modify Comet’s code to synchronize the timing of BLS key generation.
- Alternatively, the BLS key could be generated on demand (e.g., during block generation) based on the node’s validator status. This would ensure that the Comet keys and the BLS key always coexist in the app.
- The final implementation approach depends on whether key atomicity should be enforced at the Comet or app level.

### Update Delegator Address Handling:

Modify the logic so that the delegator address is provided directly during the ExtendVote() function instead of being stored in BlsPV. This will follow the [pseudo-code](https://github.com/babylonlabs-io/babylon/compare/v1.0.0-rc.1...b-harvest:babylon:poc-remove-del-addr-on-file-pv?expand=1) shared earlier.

### Check Validator Status During InitPrivSigner:

Add a check in InitPrivSigner to determine if the node is a validator by verifying whether the Consensus public key(from FilePV) is included in the validator list. If the node is a validator, generate the BLS key; otherwise, skip key generation.

gitferry:
Thanks a lot @wnjoon! So, for each point:

- Remove WrappedFilePV
  - Sounds good to me!
- Ensure Atomic Creation of BLS Key and priv_validator_key.json
  -  what about we implement a new `init` cmd that wraps Cosmos SDK's native `init` cmd here
- Update Delegator Address Handling
  - Actually I was thinking the necessity of having `DelegatorAddress`. It is needed to check whether the validator is in the validator set. However, given that executing `ExtendVote()` already ensures that this is a validator, so this check is probably not needed. Wdyt @KonradStaniec?
- Check Validator Status During InitPrivSigner
  - Does it involve a query for the current validator set? What if the user wants to become a validator later? I was thinking maybe we can introduce a flag like `--no-bls` to indicate the user's intention

PR #402, opened 2025-01-14
feat: merge base branch to main

SebastianElvis:
This PR merges the base branch with phase-3 codebase to `main`.

Comparison between base and main: https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/5905879


x/btcstaking/keeper/msg_server.go
```diff
@@ -220,18 +220,22 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
-		fp, err := ms.GetFinalityProvider(ctx, fpBTCPK)
-		if err != nil {
-			return nil, err
-		}
-		// ensure the finality provider is not slashed
-		if fp.IsSlashed() {
-			return nil, types.ErrFpAlreadySlashed.Wrapf("finality key: %s", fpBTCPK.MarshalHex())
-		}
+	// verify proof of possession
+	if err := req.Pop.Verify(parsedMsg.StakerAddress, req.BtcPk, ms.btcNet); err != nil {
```
KonradStaniec, line 224: duplicated pop verification ?

Pop is already verified above in point two:
```

	// 2. Basic stateless checks
	// - verify proof of possession
	if err := parsedMsg.ParsedPop.Verify(parsedMsg.StakerAddress, parsedMsg.StakerPK.BIP340PubKey, ms.btcNet); err != nil {
		return nil, types.ErrInvalidProofOfPossession.Wrap(err.Error())
	}
```

x/btcstaking/keeper/msg_server.go
```diff
@@ -220,18 +220,22 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
+
+	// Ensure all finality providers
+	// - are known to Babylon,
+	// - at least 1 one of them is a Babylon finality provider,
+	// - are not slashed, and
+	// - their registered epochs are finalised
+	// and then check whether the BTC stake is restaked to FPs of consumers
+	// TODO: ensure the BTC delegation does not restake to too many finality providers
+	// (pending concrete design)
+	restakedToConsumers, err := ms.validateRestakedFPs(ctx, req.FpBtcPkList)
```
KonradStaniec, line 236: `req.FpBtcPkList`  -> `parsedMsg.FinalityProviderKeys.PublicKeysBbnFormat`

x/btcstkconsumer/types/expected_keepers.go
```diff
@@ -0,0 +1,35 @@
...
+type ClientKeeper interface {
+	GetClientState(ctx sdk.Context, clientID string) (ibcexported.ClientState, bool)
+}
+
+type WasmKeeper interface {
+	GetContractInfo(ctx context.Context, contractAddress sdk.AccAddress) *wasmtypes.ContractInfo
+}
+
+// ParamSubspace defines the expected Subspace interface for parameters.
+type ParamSubspace interface {
```
KonradStaniec, line 32: not used ?

KonradStaniec (commented):
Some initial comments.

Also question, does all coninspect security issues related to integration and zoneconcierge module from previous audit are already solved (by permissioned integration) or not yet ?

app/app.go
```diff
@@ -542,10 +556,11 @@ func NewBabylonApp(
 
 	// At startup, after all modules have been registered, check that all proto
 	// annotations are correct.
-	protoFiles, err := proto.MergedRegistry()
-	if err != nil {
-		panic(err)
-	}
+	// FIXME (https://github.com/babylonlabs-io/babylon-private/issues/266): This is a temporary fix
```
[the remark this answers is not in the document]
SebastianElvis: It was. Created https://github.com/babylonlabs-io/babylon/issues/416

proto/babylon/btcstaking/v1/tx.proto
```diff
@@ -53,6 +53,9 @@ message MsgCreateFinalityProvider {
   bytes btc_pk = 4 [ (gogoproto.customtype) = "github.com/babylonlabs-io/babylon/types.BIP340PubKey" ];
   // pop is the proof of possession of btc_pk over the FP signer address.
   ProofOfPossessionBTC pop = 5;
+  // consumer_id is the ID of the consumer
+  // If it's empty, it's assumed to be Babylon's chain id
+  string consumer_id = 7;
```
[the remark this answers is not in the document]
SebastianElvis: This was because the `main` branch seems to have removed some fields, and then the ID did not change during the rebase. Can chaneg to 6

x/btcstaking/keeper/btc_delegations.go
```diff
@@ -185,6 +224,69 @@ func (k Keeper) setBTCDelegation(ctx context.Context, btcDel *types.BTCDelegatio
...
+	}
+	if !restakedToBabylon {
+		// a BTC delegation has to stake to at least 1 Babylon finality provider
+		return false, types.ErrNoBabylonFPRestaked
+	}
+	return restakedToConsumers, nil
+}
+
+// restakedFPConsumerIDs returns the unique consumer IDs of non-Babylon finality providers
+func (k Keeper) restakedFPConsumerIDs(ctx context.Context, fpBTCPKs []bbn.BIP340PubKey) ([]string, error) {
```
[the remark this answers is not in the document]
SebastianElvis: You are right. Although the current use cases of this function do not introduce non-determinism in DB, better avoid map usage completely

x/btcstaking/keeper/btc_delegators.go
```diff
@@ -61,6 +61,57 @@ func (k Keeper) getBTCDelegatorDelegations(ctx context.Context, fpBTCPK *bbn.BIP
...
+// It automatically determines and selects the appropriate KV store based on the finality provider type.
+//
+// Parameters:
+// - ctx: The context for the operation
+// - fpBTCPK: The Bitcoin public key of the finality provider
+//
+// Returns:
+// - A slice of BTCDelegation pointers representing all delegations for the given finality provider
+// - An error if the finality provider is not found or if there's an issue retrieving the delegations
+func (k Keeper) GetFPBTCDelegations(ctx context.Context, fpBTCPK *bbn.BIP340PubKey) ([]*types.BTCDelegation, error) {
```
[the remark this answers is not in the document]
SebastianElvis: You are right -- upon a FP with too many BTC delegations slashed, this function is not very scalable, even though it is not triggered often.

One improvement we can do here is to replace this function with an iterator function that takes a handler function as input. Then upon slashing a FP we only need to iterate over each BTC delegation and keep it in memory, rather than keeping the whole thing in memory. This reduces memory footprint.

The thing that is unfortunately hard to improve is the complexity -- upon slashing a FP we will have to iterate over all BTC delegations anyway.

Probably we can create a Github issue to track this potential improvement?

x/btcstaking/keeper/msg_server.go
SebastianElvis: Yeah this is duplicated -- with the new `ParseCreateDelegationMessage` we don't need to do PoP verification anymore.

x/btcstaking/keeper/msg_server.go
```diff
@@ -440,6 +454,12 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
 		return nil, err
 	}
 
+	// check whether the BTC stake is restaked to FPs of consumers
+	_, err = ms.validateRestakedFPs(ctx, btcDel.FpBtcPkList)
```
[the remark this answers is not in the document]
SebastianElvis: Good catch, so many duplicated check due to rebase :sweat:

x/btcstkconsumer/types/expected_keepers.go
SebastianElvis: You are right, this was the old way of handling params. Can remove

proto/scripts/protocgen.sh
```diff
@@ -19,4 +19,4 @@ cd ..
 cp -r github.com/babylonlabs-io/babylon/* ./
 rm -rf github.com
 
-go mod tidy -compat=1.23
```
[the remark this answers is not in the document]
SebastianElvis: As it does not seem to do anything meaningful yet increasing the `make proto-gen` execution time quite a lot.

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/cometbft/cometbft-db v0.15.0
 	github.com/cosmos/cosmos-sdk v0.50.11
 	github.com/cosmos/ibc-go/modules/light-clients/08-wasm v0.0.0-20240429153234-e1e6da7e4ead
-	github.com/cosmos/relayer/v2 v2.5.3
+	github.com/cosmos/relayer/v2 v2.5.2
```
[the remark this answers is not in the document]
SebastianElvis: Nice, just found out that the `main` has bumped relayer and grpc to the latest :+1:

SebastianElvis:
Thanks for the comments @KonradStaniec ! I have resolved the comments at https://github.com/babylonlabs-io/babylon/pull/418, feel free to take a look

> Also question, does all coninspect security issues related to integration and zoneconcierge module from previous audit are already solved (by permissioned integration) or not yet ?

yes, https://github.com/babylonlabs-io/pm/issues/14 and https://github.com/babylonlabs-io/pm/issues/12 are already closed

x/btcstaking/keeper/btc_staking_consumer_events_test.go
```diff
@@ -0,0 +1,344 @@
...
+		require.Equal(t, actualDel.StakingTx, ev.StakingTx)
+		require.Equal(t, actualDel.StakingOutputIdx, ev.StakingOutputIdx)
+		require.Equal(t, actualDel.UnbondingTime, ev.UnbondingTime)
+	})
+}
+
+func FuzzSetBTCStakingEventStore_UnbondedDel(f *testing.F) {
+	datagen.AddRandomSeedsToFuzzer(f, 10)
+
+	f.Fuzz(func(t *testing.T, seed int64) {
```
Lazar955, line 171: would be nice to add `t.Parallel()` in all unit/fuzz tests

test/e2e/configurer/chain/commands_btcstaking.go
```diff
@@ -125,8 +129,10 @@ func (n *NodeConfig) CreateBTCDelegation(
 		// broadcast stuff
 		cmd = append(cmd, "-b=sync", "--yes")
 	}
-
+	// cmd = append(cmd, fmt.Sprintf("--chain-id=%s", n.chainId), "-b=sync", "--yes", "--keyring-backend=test", "--log_format=json", "--home=/home/babylon/babylondata")
```
Lazar955, line 132: remove commented out code

Lazar955 (commented):
a bit overwhelming pr :D

gitferry (commented):
Amazing work. Some initial comments:

Makefile
```diff
@@ -263,20 +264,20 @@ endif
...
+	$(MAKE) test-e2e-cache-btc-timestamping
+	$(MAKE) test-e2e-cache-btc-staking
+	$(MAKE) clean-e2e
+	$(MAKE) test-e2e-cache-btc-staking-pre-approval
+	$(MAKE) test-e2e-cache-ibc-transfer
+#	$(MAKE) test-e2e-cache-upgrade-v1
 
-test-e2e-cache-ibc-transfer:
-	go test -run TestIBCTranferTestSuite -mod=readonly -timeout=60m -v $(PACKAGES_E2E) --tags=e2e
+clean-e2e:
```
RafilxTenfen, line 274: nice!

app/keepers/keepers.go
```diff
@@ -397,6 +411,30 @@ func (ak *AppKeepers) InitKeepers(
...
+		ak.IBCKeeper.ChannelKeeper,
+		ak.IBCKeeper.PortKeeper,
+		scopedWasmKeeper,
+		ak.TransferKeeper,
+		bApp.MsgServiceRouter(),
+		bApp.GRPCQueryRouter(),
+		homePath,
+		wasmConfig,
+		WasmCapabilities(),
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
```
RafilxTenfen, line 434:
```suggestion
		appparams.AccGov.String(),
```

app/keepers/keepers.go
```diff
@@ -469,26 +507,14 @@ func (ak *AppKeepers) InitKeepers(
...
-	)
-	ak.BtcCheckpointKeeper = btcCheckpointKeeper
-	ak.BTCLightClientKeeper = *btclightclientKeeper.SetHooks(
-		btclightclienttypes.NewMultiBTCLightClientHooks(ak.BtcCheckpointKeeper.Hooks()),
+		runtime.NewKVStoreService(keys[bsctypes.StoreKey]),
+		ak.AccountKeeper,
+		ak.BankKeeper,
+		ak.IBCKeeper.ClientKeeper,
+		ak.WasmKeeper,
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
```
RafilxTenfen, line 517:
```suggestion
		appparams.AccGov.String(),
```

app/keepers/keepers.go
```diff
@@ -508,10 +535,65 @@ func (ak *AppKeepers) InitKeepers(
...
 	)
 
+	monitorKeeper := monitorkeeper.NewKeeper(
+		appCodec,
+		runtime.NewKVStoreService(keys[monitortypes.StoreKey]),
+		&btclightclientKeeper,
+	)
+
+	// create querier for KVStore
+	storeQuerier, ok := bApp.CommitMultiStore().(storetypes.Queryable)
```
RafilxTenfen, line 549: it seems out of scope to get the store query root into a keeper, and it seems like the zone concierge already imports the needed keeper epoching and checkpointing

app/keepers/keepers.go
```diff
@@ -571,13 +630,18 @@ func (ak *AppKeepers) InitKeepers(
 	transferStack = transfer.NewIBCModule(ak.TransferKeeper)
 	transferStack = ibcfee.NewIBCMiddleware(transferStack, ak.IBCFeeKeeper)
 
+	var zoneConciergeStack porttypes.IBCModule
+	zoneConciergeStack = zoneconcierge.NewIBCModule(ak.ZoneConciergeKeeper)
```
RafilxTenfen, line 634:
```suggestion
	zoneConciergeStack := zoneconcierge.NewIBCModule(ak.ZoneConciergeKeeper)
```

contrib/images/Makefile
```diff
@@ -29,4 +29,4 @@ cosmos-relayer: cosmos-relayer-rmi
 cosmos-relayer-rmi:
 	docker rmi babylonlabs-io/cosmos-relayer 2>/dev/null; true
 
-.PHONY: all babylond cosmos-relayer e2e-init-chain babylond-rmi babylond-rmi-upgrade cosmos-relayer-rmi
+.PHONY: all babylond babylond-before-upgrade cosmos-relayer e2e-init-chain babylond-rmi babylond-rmi-upgrade cosmos-relayer-rmi
```
RafilxTenfen, line 32:
```suggestion
.PHONY: all babylond cosmos-relayer e2e-init-chain babylond-rmi babylond-rmi-upgrade cosmos-relayer-rmi 
```

proto/babylon/btcstaking/v1/btcstaking.proto
```diff
@@ -39,6 +39,9 @@ message FinalityProvider {
     // highest_voted_height is the highest height for which the
     // finality provider has voted
     uint32 highest_voted_height = 9;
+    // consumer_id is the ID of the consumer the finality provider is operating on.
+    // If it's missing / empty, it's assumed the finality provider is operating in the Babylon chain.
+    string consumer_id = 10;
```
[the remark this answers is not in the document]
RafilxTenfen: yeap, no migration needed

proto/babylon/btcstkconsumer/v1/query.proto
```diff
@@ -0,0 +1,155 @@
...
+// QueryFinalityProviderRequest requests information about a finality provider
+message QueryFinalityProviderRequest {
+  // consumer id is the consumer id this finality provider is registered to
+  string consumer_id = 1;
+  // fp_btc_pk_hex is the hex str of Bitcoin secp256k1 PK of the finality provider
+  string fp_btc_pk_hex = 2;
+}
+
+// QueryFinalityProviderResponse contains information about a finality provider
+message QueryFinalityProviderResponse {
```
RafilxTenfen, line 109: why don't we use the finality provider from `x/btcstaking`

proto/babylon/finality/v1/tx.proto
```diff
@@ -25,6 +25,9 @@ service Msg {
     rpc UnjailFinalityProvider(MsgUnjailFinalityProvider) returns (MsgUnjailFinalityProviderResponse);
```
RafilxTenfen, line 25: remove `    // TODO: msg for evidence of equivocation. this is not specified yet`

proto/babylon/zoneconcierge/v1/zoneconcierge.proto
```diff
@@ -0,0 +1,148 @@
...
+  // babylon_header_hash is the hash of the babylon block that includes this CZ
+  // header
+  bytes babylon_header_hash = 5;
+  // babylon_header_height is the height of the babylon block that includes this CZ
+  // header
+  uint64 babylon_header_height = 6;
+  // epoch is the epoch number of this header on Babylon ledger
+  uint64 babylon_epoch = 7;
+  // babylon_tx_hash is the hash of the tx that includes this header
+  // (babylon_block_height, babylon_tx_hash) jointly provides the position of
```
RafilxTenfen, line 37: what happens if in the same tx two headers are indexed?

x/zoneconcierge/keeper/hooks.go
```diff
@@ -0,0 +1,65 @@
...
+func (h Hooks) AfterRawCheckpointFinalized(ctx context.Context, epoch uint64) error {
+	headersToBroadcast := h.k.getHeadersToBroadcast(ctx)
+
+	// send BTC timestamp to all open channels with ZoneConcierge
+	h.k.BroadcastBTCTimestamps(ctx, epoch, headersToBroadcast)
+
+	// Update the last broadcasted segment
+	h.k.setLastSentSegment(ctx, &types.BTCChainSegment{
+		BtcHeaders: headersToBroadcast,
+	})
```
RafilxTenfen, line 49: I did not understood why we keep the last sent segment

test/e2e/containers/containers.go
```diff
@@ -272,7 +272,7 @@ func (m *Manager) RunNodeResource(chainId string, containerName, valCondifDir st
 		Entrypoint: []string{
 			"sh",
 			"-c",
-			"babylond start " + FlagHome,
+			"babylond start " + FlagHome + " --log_level trace --trace",
```
RafilxTenfen, line 275: if we want to use trace log level should be modified at https://github.com/babylonlabs-io/babylon/blob/ee744545e9e6369cd1544f8f6d1bb4f64b634643/test/e2e/initialization/node.go#L358

But in my opinion trace is too much :sweat_smile:

test/e2e/ibc_transfer_e2e_test.go
```diff
@@ -96,7 +96,7 @@ func (s *IBCTransferTestSuite) Test1IBCTransfer() {
 		}
 		return math.Abs(float64(balanceA.Sub(transferCoin).AmountOf(denom).Int64()-
 			balanceA2.AmountOf(denom).Int64())) < delta
-	}, 10*time.Second, 1*time.Second, "Transfer was not successful")
+	}, 1*time.Minute, 1*time.Second, "Transfer was not successful")
```
RafilxTenfen, line 99: why increase that much?

wasmbinding/testdata/artifacts/testgrpc.wasm
RafilxTenfen, line 1: should we have tests for our https://github.com/babylonlabs-io/babylon-contract?

wasmbinding/wasm.go
```diff
@@ -160,8 +166,9 @@ func RegisterCustomPlugins(
 	ek *epochingkeeper.Keeper,
 	ck *checkpointingkeeper.Keeper,
 	lcKeeper *lcKeeper.Keeper,
+	zcKeeper *zckeeper.Keeper,
 ) []wasmkeeper.Option {
-	wasmQueryPlugin := NewQueryPlugin(ek, ck, lcKeeper)
+	wasmQueryPlugin := NewQueryPlugin(ek, ck, lcKeeper, zcKeeper)
```
RafilxTenfen, line 171: if we had custom msg handler for wasm would also eliminate the possibility to people bypass the epoch staking? @maurolacy

x/btcstaking/keeper/msg_server.go
```diff
@@ -295,6 +294,13 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 	if err := ms.AddBTCDelegation(ctx, newBTCDel); err != nil {
 		panic(fmt.Errorf("failed to add BTC delegation that has passed verification: %w", err))
 	}
+	// if this BTC delegation is restaked to consumers' FPs, add it to btcstkconsumer indexes
+	// TODO: revisit the relationship between BTC staking module and BTC staking consumer module
+	if restakedToConsumers {
+		if err := ms.IndexBTCConsumerDelegation(ctx, newBTCDel); err != nil {
```
RafilxTenfen, line 300: should it be indexed even if the BTC staking tx is not confirmed?

x/btcstkconsumer/keeper/finality_provider_registry.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// SetConsumerFinalityProvider adds the given finality provider to CZ chains KVStore
+func (k Keeper) SetConsumerFinalityProvider(ctx context.Context, fp *btcstaking.FinalityProvider) {
+	store := k.finalityProviderStore(ctx, fp.ConsumerId)
+	fpBytes := k.cdc.MustMarshal(fp)
+	store.Set(fp.BtcPk.MustMarshal(), fpBytes)
+	// Update the finality provider chain store
+	fpChainStore := k.finalityProviderChainStore(ctx)
+	fpChainStore.Set(fp.BtcPk.MustMarshal(), []byte(fp.ConsumerId))
```
RafilxTenfen, line 23: We should consider use `collections.map` for this one as it seems we store the `fp_btc_pk => chain ID`
or maybe just get it from the btcstaking?

x/btcstkconsumer/keeper/finality_provider_registry.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// GetConsumerFinalityProvider gets the finality provider in the chain id with the given finality provider Bitcoin PK
+func (k Keeper) GetConsumerFinalityProvider(ctx context.Context, consumerId string, fpBTCPK *bbn.BIP340PubKey) (*btcstaking.FinalityProvider, error) {
+	if !k.HasConsumerFinalityProvider(ctx, fpBTCPK) {
+		return nil, btcstaking.ErrFpNotFound
+	}
+	store := k.finalityProviderStore(ctx, consumerId)
```
RafilxTenfen, line 31: seems like we are duplicating the fp data

x/btcstkconsumer/keeper/finality_provider_registry.go
```diff
@@ -0,0 +1,126 @@
...
+
+		shouldContinue := handler(&fp)
+		if !shouldContinue {
+			return
+		}
+	}
+}
+
+// HasFPs checks if the FPs list exists for a given chain id
+func (k Keeper) HasFPs(ctx context.Context, chainID string) bool {
```
RafilxTenfen, line 75: remove as it is not being used

x/btcstkconsumer/keeper/finality_provider_registry.go
```diff
@@ -0,0 +1,126 @@
...
+	store := k.finalityProviderStore(ctx, chainID)
+	iter := store.Iterator(nil, nil)
+	defer iter.Close()
+	return iter.Valid()
+}
+
+// GetFPs gets the list of FPs, i.e., the finality provider set for a given chain id
+func (k Keeper) GetFPs(ctx context.Context, chainID string) map[string]btcstaking.FinalityProvider {
```
RafilxTenfen, line 83: not used

proto/babylon/btcstaking/v1/events.proto
```diff
@@ -73,19 +78,27 @@ message EventPowerDistUpdate {
   // is slashed
   // TODO: unify with existing slashing events
   message EventSlashedFinalityProvider {
-    bytes pk = 1 [ (gogoproto.customtype) = "github.com/babylonlabs-io/babylon/types.BIP340PubKey" ];
+    bytes pk = 1 [(gogoproto.customtype) = "github.com/babylonlabs-io/babylon/types.BIP340PubKey"];
+  }
+
+  // EventSlashedBTCDelegation is emitted for each BTC delegation that restakes to a slashed consumer finality provider.
+  // It indicates that the voting power of affected Babylon finality providers will be discounted for this delegation.
+  message EventSlashedBTCDelegation {
```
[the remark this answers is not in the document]
SebastianElvis: Each BTC delegation has to multi-stake to a Babylon FP so it's a bit hard to say the BTC delegation is from a consumer. What do you think we keep it as is?

proto/babylon/btcstkconsumer/v1/query.proto
```diff
@@ -0,0 +1,155 @@
...
+  // if it's 0 then the finality provider is not slashed
+  uint64 slashed_babylon_height = 6;
+  // slashed_btc_height indicates the BTC height when
+  // the finality provider is slashed.
+  // if it's 0 then the finality provider is not slashed
+  uint32 slashed_btc_height = 7;
+  // height is the queried Babylon height
+  uint64 height = 8;
+  // voting_power is the voting power of this finality provider at the given height
+  uint64 voting_power = 9;
```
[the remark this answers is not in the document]
SebastianElvis: Not yet. Nevertheless, the consumer side won't share the query protobuf objects with Babylon; rather it will implement its own queries (e.g., [here](https://github.com/babylonlabs-io/babylon-contract/blob/f73fb6a8535c00858f6dd6ddb5e86e6388b01da1/contracts/btc-staking/src/queries.rs))

test/e2e/btc_staking_pre_approval_e2e_test.go
```diff
@@ -108,13 +108,14 @@ func (s *BTCStakingPreApprovalTestSuite) Test1CreateFinalityProviderAndDelegatio
...
+	delBtcPK := bbn.NewBIP340PubKeyFromBTCPK(s.delBTCSK.PubKey())
 	nonValidatorNode.CreateBTCDelegation(
-		bbn.NewBIP340PubKeyFromBTCPK(s.delBTCSK.PubKey()),
+		delBtcPK,
 		pop,
 		stakingTx,
 		// We are passing `nil` as inclusion proof will be provided in separate tx
 		nil,
-		s.cacheFP.BtcPk,
+		[]bbn.BIP340PubKey{*s.cacheFP.BtcPk},
```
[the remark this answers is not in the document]
SebastianElvis: We did this in https://github.com/babylonlabs-io/babylon-sdk/blob/dcef36fa85168bfc6d331527960a8870e8c3b8fd/tests/e2e/bcd_consumer_integration_test.go.

x/btcstaking/client/cli/tx.go
```diff
@@ -190,7 +194,7 @@ func NewEditFinalityProviderCmd() *cobra.Command {
 
 func NewCreateBTCDelegationCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "create-btc-delegation [btc_pk] [pop_hex] [staking_tx] [inclusion_proof] [fp_pk] [staking_time] [staking_value] [slashing_tx] [delegator_slashing_sig] [unbonding_tx] [unbonding_slashing_tx] [unbonding_time] [unbonding_value] [delegator_unbonding_slashing_sig]",
+		Use:   "create-btc-delegation [btc_pk] [pop_hex] [staking_tx] [inclusion_proof] [fp_pk1],[fp_pk2],... [staking_time] [staking_value] [slashing_tx] [delegator_slashing_sig] [unbonding_tx] [unbonding_slashing_tx] [unbonding_time] [unbonding_value] [delegator_unbonding_slashing_sig]",
```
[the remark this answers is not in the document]
SebastianElvis: I prefer the original notation in which `[x]` denotes a variable. Wdyt?

x/btcstaking/keeper/btc_consumer_delegations.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// get staking tx hash
+	stakingTxHash, err := btcDel.GetStakingTxHash()
+	if err != nil {
+		return err
+	}
+
+	// for each finality provider the delegation restakes to, update its index
+	for i := range btcDel.FpBtcPkList {
+		fpBTCPK := btcDel.FpBtcPkList[i]
```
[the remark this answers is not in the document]
SebastianElvis: This is to avoid https://github.com/golang/go/discussions/56010. Although this seems fixed in Go 1.23, wdyt we keep the existing code to make the assignment explicit?

x/btcstaking/keeper/btc_delegations.go
```diff
@@ -185,6 +225,72 @@ func (k Keeper) setBTCDelegation(ctx context.Context, btcDel *types.BTCDelegatio
...
+	consumerIDMap := make(map[string]struct{})
```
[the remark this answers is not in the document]
SebastianElvis: This is not necessarily true as it's possible that multiple FPs are in the same consumer, i.e., `len(fpBTCPKs)` is bigger than `len(consumerIDMap)`

x/btcstaking/keeper/finality_providers.go
```diff
@@ -113,6 +129,111 @@ func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error
...
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// Send collected events to each involved consumer chain
+	for consumerID, events := range consumerEvents {
+		if err := k.AddBTCStakingConsumerEvents(ctx, consumerID, events); err != nil {
```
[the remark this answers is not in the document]
SebastianElvis: Here `seenConsumers` map is only used for checking existence and we never iterate over it, so the function is deterministic.

app/keepers/keepers.go
SebastianElvis: This is necesasry to get the inclusion proof of some metadata in BTC timestamping, see [here](https://github.com/babylonlabs-io/babylon/blob/873f1232365573a97032037af4ac99b5e3fcada8/x/zoneconcierge/keeper/query_kvstore.go#L19-L20).

app/keepers/keepers.go
SebastianElvis: This seems necessary for L635. If I apply this then L635 will not compile

proto/babylon/btcstkconsumer/v1/query.proto
SebastianElvis: We created this endpoint for consumer FPs explicitly, but yes merging two is surely possible

proto/babylon/zoneconcierge/v1/query.proto
```diff
@@ -0,0 +1,202 @@
...
+  string consumer_id = 1;
+  // pagination defines whether to have the pagination in the request
+  cosmos.base.query.v1beta1.PageRequest pagination = 2;
+}
+
+// QueryListHeadersResponse is response type for the Query/ListHeaders RPC
+// method.
+message QueryListHeadersResponse {
+  // headers is the list of headers
+  repeated babylon.zoneconcierge.v1.IndexedHeader headers = 1;
```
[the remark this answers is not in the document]
SebastianElvis: sg, https://github.com/babylonlabs-io/babylon/issues/422

proto/babylon/zoneconcierge/v1/zoneconcierge.proto
SebastianElvis: Good question. It's fine actually as we have consumer ID as the KV store prefix ([ref](https://github.com/babylonlabs-io/babylon/blob/873f1232365573a97032037af4ac99b5e3fcada8/x/zoneconcierge/keeper/canonical_chain_indexer.go#L59-L62))

x/zoneconcierge/keeper/hooks.go
SebastianElvis: Babylon needs to send the *new* BTC headers during the last finalised epoch to consumer chains. This means

- Babylon needs to at least remember the last BTC header sent in the last finalisd epoch
- However, BTC chain can get reorg'ed. So Babylon needs to remember at least the last k BTC headers

test/e2e/containers/containers.go
SebastianElvis: This seems to be debug leftover, removed

test/e2e/ibc_transfer_e2e_test.go
SebastianElvis: Hmm probably IBC transfer packets take some time? This was for fixing the flaky e2e test AFAIK

wasmbinding/testdata/artifacts/testgrpc.wasm
SebastianElvis: We do have e2e tests in https://github.com/babylonlabs-io/babylon-sdk/blob/dcef36fa85168bfc6d331527960a8870e8c3b8fd/tests/e2e/bcd_consumer_integration_test.go

x/btcstaking/keeper/btc_consumer_delegations.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		// get BTC delegation index under this finality provider
+		btcDelIndex := k.getBTCConsumerDelegatorDelegationIndex(ctx, &fpBTCPK, btcDel.BtcPk)
+		if btcDelIndex == nil {
+			btcDelIndex = bstypes.NewBTCDelegatorDelegationIndex()
+		}
+		// index staking tx hash of this BTC delegation
+		if err := btcDelIndex.Add(stakingTxHash); err != nil {
+			return bstypes.ErrInvalidStakingTx.Wrapf("%s", err.Error())
```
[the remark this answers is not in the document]
SebastianElvis: Hmm actually this will make linter unhappy `non-constant format string in call to (*cosmossdk.io/errors.Error).Wrapf`

x/btcstaking/keeper/msg_server.go
SebastianElvis: Yes, this needs to be indexed at the beginning

x/btcstkconsumer/keeper/finality_provider_registry.go
SebastianElvis: We can do this in a future PR. At the time of implementing this code we haven't started exploring the `collections` package. Created https://github.com/babylonlabs-io/babylon/issues/423 to track this

SebastianElvis:
Thanks @gitferry @RafilxTenfen @Lazar955 for the comprehensive comments! I have addressed or answered the comments and please find the PR [here](https://github.com/babylonlabs-io/babylon/pull/424)

proto/babylon/zoneconcierge/v1/zoneconcierge.proto
RafilxTenfen: I mean two headers for the same consumer/height/tx (as tx can contain multiple msgs) being indexed 

Seems like we are only overwriting it by the last one that passes

x/btcstaking/keeper/msg_server.go
```diff
@@ -295,6 +294,13 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 	if err := ms.AddBTCDelegation(ctx, newBTCDel); err != nil {
```
RafilxTenfen, line 294: Here we are adding the BTC delegation normally as it is the same for BTC delegations to Babylon vs Consumers, that means for both cases events `EventBTCDelegationStateUpdate` are being added and later processed in `ProcessAllPowerDistUpdateEvents` - https://github.com/babylonlabs-io/babylon/blob/be9eb6b9f4db544eb35db8b4767852d12910b9eb/x/finality/keeper/power_dist_change.go#L170

Since rewards of consumer chains will not be handled by the Babylon chain we should handle and do not store the changes of those finality providers in the incentive store (do not call `processRewardTracker` https://github.com/babylonlabs-io/babylon/blob/be9eb6b9f4db544eb35db8b4767852d12910b9eb/x/finality/keeper/power_dist_change.go#L216) for consumers BTC delegations and FPs

x/btcstaking/keeper/finality_providers.go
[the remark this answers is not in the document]
RafilxTenfen: humm, if for some reason the consumer event fails `AddBTCStakingConsumerEvents` it will stop the execution and with different orders it will not be deterministic 

We should order a slice by the consumer ID presents in the `consumerEvents`

x/btcstaking/keeper/finality_providers.go
```diff
@@ -113,6 +129,111 @@ func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error
...
+		// Track consumers seen for this delegation
+		seenConsumers := make(map[string]struct{})
+
+		for _, delegationFPBTCPK := range delegation.FpBtcPkList {
+			fpBTCPKHex := delegationFPBTCPK.MarshalHex()
+			consumerID, exists := fpToConsumerMap[fpBTCPKHex]
+			if !exists {
+				// If not in map, check if it's a Babylon FP or get its consumer
+				if _, err := k.GetFinalityProvider(ctx, delegationFPBTCPK); err == nil {
+					continue // It's a Babylon FP, skip
```
RafilxTenfen, line 206: we already know if the finality provider `fpBTCPK` is from the Babylon when querying [HasFinalityProvider](https://github.com/babylonlabs-io/babylon/blob/873f1232365573a97032037af4ac99b5e3fcada8/x/btcstaking/keeper/btc_delegators.go#L79)

It would reduce one read by FP in the list, can be addressed in a future task

x/btcstaking/keeper/grpc_query.go
```diff
@@ -67,12 +67,27 @@ func (k Keeper) FinalityProvider(c context.Context, req *types.QueryFinalityProv
 	}
 
 	ctx := sdk.UnwrapSDKContext(c)
+	currBlockHeight := uint64(ctx.BlockHeight())
+
 	fp, err := k.GetFinalityProvider(ctx, key)
 	if err != nil {
+		// Try in the btcstkconsumer module
+		if k.BscKeeper.HasConsumerFinalityProvider(ctx, fpPK) {
```
RafilxTenfen, line 75: if we are going to have 2 different stores with FPs we should return the same structure/interface and have a function to query a FP by their fpPK to avoid handling in the error case

x/btcstaking/types/keys.go
```diff
@@ -27,4 +27,6 @@ var (
 	PowerDistUpdateKey        = []byte{0x08}             // key prefix for power distribution update events
 	AllowedStakingTxHashesKey = collections.NewPrefix(9) // key prefix for allowed staking tx hashes
 	HeightToVersionMapKey     = []byte{0x10}             // key prefix for height to version map
+	BTCConsumerDelegatorKey   = []byte{0x11}             // key prefix for the Consumer BTC delegators
+	BTCStakingEventKey        = []byte{0x12}             // key prefix for the BTC staking events
```
RafilxTenfen, line 31: could we put this inside the `x/btcstkconsumer` ? @gusin13

x/zoneconcierge/types/genesis.go
```diff
@@ -0,0 +1,33 @@
+package types
+
+import (
+	host "github.com/cosmos/ibc-go/v8/modules/core/24-host"
+)
+
+// DefaultGenesis returns the default genesis state
+func DefaultGenesis() *GenesisState {
+	return &GenesisState{
```
RafilxTenfen, line 9: We have a lot of keys 
```go

var (
	PortKey               = []byte{0x11} // PortKey defines the key to store the port ID in store
	ChainInfoKey          = []byte{0x12} // ChainInfoKey defines the key to store the chain info for each CZ in store
	CanonicalChainKey     = []byte{0x13} // CanonicalChainKey defines the key to store the canonical chain for each CZ in store
	ForkKey               = []byte{0x14} // ForkKey defines the key to store the forks for each CZ in store
	EpochChainInfoKey     = []byte{0x15} // EpochChainInfoKey defines the key to store each epoch's latests chain info for each CZ in store
	LastSentBTCSegmentKey = []byte{0x16} // LastSentBTCSegmentKey is key holding last btc light client segment sent to other cosmos zones
	ParamsKey             = []byte{0x17} // key prefix for the parameters
	SealedEpochProofKey   = []byte{0x18} // key prefix for proof of sealed epochs
)
```

Every data that needs to be know in case of a chain failure should be inside the `GenesisState`, a few of our modules don't have that, but still we should take note and create issues

x/btcstkconsumer/client/cli/tx.go
```diff
@@ -0,0 +1,75 @@
...
+	return cmd
+}
+
+func NewRegisterConsumerCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "register-consumer <consumer-id> <name> [description]",
+		Args:  cobra.RangeArgs(2, 3),
+		Short: "Registers a CZ consumer",
+		Long: strings.TrimSpace(
+			`Registers a CZ consumer.`,
```
RafilxTenfen, line 39: long is the same as short

x/btcstkconsumer/genesis.go
```diff
@@ -0,0 +1,23 @@
...
+
+// InitGenesis initializes the module's state from a provided genesis state.
+func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
+	if err := k.SetParams(ctx, genState.Params); err != nil {
+		panic(err)
+	}
+}
+
+// ExportGenesis returns the module's exported genesis.
+func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
```
RafilxTenfen, line 18: Same, genesis state should have every data needed to relaunch the chain in case of failure

x/btcstkconsumer/keeper/finality_provider_registry.go
```diff
@@ -0,0 +1,126 @@
...
+	if !k.HasConsumerFinalityProvider(ctx, fpBTCPK) {
+		return "", btcstaking.ErrFpNotFound
+	}
+	store := k.finalityProviderChainStore(ctx)
+	chainID := store.Get(*fpBTCPK)
+	return string(chainID), nil
+}
+
+// IterateFPs iterates over all finality providers for a given chain
+func (k Keeper) IterateFPs(ctx context.Context, chainId string, handler func(fp *btcstaking.FinalityProvider) bool) {
```
RafilxTenfen, line 59: also not used

x/btcstkconsumer/keeper/query.go
```diff
@@ -0,0 +1,7 @@
+package keeper
+
+import (
+	"github.com/babylonlabs-io/babylon/x/btcstkconsumer/types"
+)
+
+var _ types.QueryServer = Keeper{}
```
RafilxTenfen, line 7: we already have this in [grpc_query](https://github.com/babylonlabs-io/babylon/blob/ce31233233d496868bd1c83794c30ec3456c0b16/x/btcstkconsumer/keeper/grpc_query.go#L18)
File could be removed

x/btcstkconsumer/types/chain_register.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func (m *MsgRegisterConsumer) ValidateBasic() error {
+	if len(m.ConsumerId) == 0 {
+		return fmt.Errorf("ConsumerId must be non-empty")
+	}
+	if len(m.ConsumerName) == 0 {
+		return fmt.Errorf("ConsumerName must be non-empty")
+	}
+	if len(m.ConsumerDescription) == 0 {
```
RafilxTenfen, line 14: why do we allow not specify the description in CLI? 
https://github.com/babylonlabs-io/babylon/blob/ce31233233d496868bd1c83794c30ec3456c0b16/x/btcstkconsumer/client/cli/tx.go#L56

x/btcstkconsumer/types/expected_keepers.go
```diff
@@ -0,0 +1,29 @@
...
+type AccountKeeper interface {
+	GetAccount(context.Context, sdk.AccAddress) sdk.AccountI // only used for simulation
+	// Methods imported from account should be defined here
+}
+
+// BankKeeper defines the expected interface for the Bank module.
+type BankKeeper interface {
+	SpendableCoins(context.Context, sdk.AccAddress) sdk.Coins
+	// Methods imported from bank should be defined here
+}
```
RafilxTenfen, line 21: not used

x/btcstkconsumer/keeper/keeper.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+type Keeper struct {
+	cdc          codec.BinaryCodec
+	storeService store.KVStoreService
+
+	accountKeeper types.AccountKeeper
+	bankKeeper    types.BankKeeper
+	clientKeeper  types.ClientKeeper
+	wasmKeeper    types.WasmKeeper
```
RafilxTenfen, line 21: not being used
```suggestion
	clientKeeper  types.ClientKeeper
```

x/btcstkconsumer/types/genesis.go
```diff
@@ -0,0 +1,17 @@
+package types
+
+// DefaultIndex is the default global index
+const DefaultIndex uint64 = 1
```
RafilxTenfen, line 4: remove?

x/btcstkconsumer/types/keys.go
```diff
@@ -0,0 +1,26 @@
...
+var (
+	ParamsKey                        = []byte("p_btcstkconsumer")
+	ConsumerRegisterKey              = []byte{0x01} // ConsumerRegisterKey defines the key to the chain register for each consumer in store
+	ConsumerFinalityProviderKey      = []byte{0x02} // ConsumerFinalityProviderKey defines the key to the consumer finality providers store
+	ConsumerFinalityProviderChainKey = []byte{0x03} // ConsumerFinalityProviderChainKey defines the key to the consumer chains per FP BTC PK store
+)
+
+func KeyPrefix(p string) []byte {
+	return []byte(p)
+}
```
RafilxTenfen, line 26: not being used

x/btcstkconsumer/types/types.go
RafilxTenfen, line 1: delete?

x/finality/client/cli/tx.go
```diff
@@ -167,37 +165,3 @@ func NewAddFinalitySigCmd() *cobra.Command {
 
 	return cmd
 }
-
-func NewUnjailFinalityProviderCmd() *cobra.Command {
-	cmd := &cobra.Command{
-		Use:   "unjail-finality-provider [fp_btc_pk]",
```
RafilxTenfen, line 173: why remove it? do we not have the unjail option @gitferry

x/zoneconcierge/client/cli/tx.go
```diff
@@ -0,0 +1,23 @@
+package cli
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	// "github.com/cosmos/cosmos-sdk/client/flags"
```
RafilxTenfen, line 8: can we remove this file?

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -0,0 +1,403 @@
...
+	if err := bbntypes.CheckForDuplicatesAndEmptyStrings(req.ConsumerIds); err != nil {
+		return nil, status.Error(codes.InvalidArgument, types.ErrInvalidConsumerIDs.Wrap(err.Error()).Error())
+	}
+
+	ctx := sdk.UnwrapSDKContext(c)
+	resp := &types.QueryFinalizedChainsInfoResponse{FinalizedChainsInfo: []*types.FinalizedChainInfo{}}
+
+	// find the last finalised epoch
+	lastFinalizedEpoch := k.GetLastFinalizedEpoch(ctx)
+	for _, ConsumerId := range req.ConsumerIds {
```
RafilxTenfen, line 242: Thinking if doing all of this to 100 consumers at once wouldn't it be too much? 
Maybe we should only allow to query one by one

x/zoneconcierge/keeper/header_handler.go
```diff
@@ -0,0 +1,73 @@
...
+
+	var (
+		chainInfo *types.ChainInfo
+		err       error
+	)
+	if !k.HasChainInfo(ctx, indexedHeader.ConsumerId) {
+		// chain info does not exist yet, initialise chain info for this chain
+		chainInfo, err = k.InitChainInfo(ctx, indexedHeader.ConsumerId)
+		if err != nil {
+			panic(fmt.Errorf("failed to initialize chain info of %s: %w", indexedHeader.ConsumerId, err))
```
RafilxTenfen, line 37: all the panics here only can happens due to a programming error?

x/zoneconcierge/keeper/ibc_header_decorator.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// NewIBCHeaderDecorator creates a new IBCHeaderDecorator
+func NewIBCHeaderDecorator(k Keeper) *IBCHeaderDecorator {
+	return &IBCHeaderDecorator{
+		k: k,
+	}
+}
+
+func (d *IBCHeaderDecorator) getHeaderAndClientState(ctx sdk.Context, m sdk.Msg) (*types.HeaderInfo, *ibctmtypes.ClientState) {
```
RafilxTenfen, line 25: can we add some documentation on this function on what it does? or maybe only to `func (d *IBCHeaderDecorator) PostHandle`

x/zoneconcierge/keeper/keeper.go
```diff
@@ -0,0 +1,131 @@
...
+type (
+	Keeper struct {
+		cdc              codec.BinaryCodec
+		storeService     corestoretypes.KVStoreService
+		ics4Wrapper      types.ICS4Wrapper
+		clientKeeper     types.ClientKeeper
+		connectionKeeper types.ConnectionKeeper
+		channelKeeper    types.ChannelKeeper
+		portKeeper       types.PortKeeper
+		authKeeper       types.AccountKeeper
```
RafilxTenfen, line 26: authKeeper, bankKeeper not used

x/zoneconcierge/module.go
```diff
@@ -0,0 +1,164 @@
...
+// AppModule
+// ----------------------------------------------------------------------------
+
+// AppModule implements the AppModule interface that defines the inter-dependent methods that modules need to implement
+type AppModule struct {
+	AppModuleBasic
+
+	keeper        keeper.Keeper
+	accountKeeper types.AccountKeeper
+	bankKeeper    types.BankKeeper
```
RafilxTenfen, line 101: remove unused keepers

x/zoneconcierge/module_ibc.go
```diff
@@ -0,0 +1,272 @@
...
+			return errorsmod.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal packet acknowledgement with protobuf (error: %v) or json (error: %v)", errProto, errJson)
+		}
+	}
+
+	// // TODO (Babylon): Dispatch and process packet
+	// switch packet := modulePacketData.Packet.(type) {
+	// default:
+	// 	errMsg := fmt.Sprintf("unrecognized %s packet type: %T", types.ModuleName, packet)
+	// 	return errorsmod.Wrap(sdkerrors.ErrUnknownRequest, errMsg)
+	// }
```
RafilxTenfen, line 225: leftover?

x/zoneconcierge/module_ibc.go
```diff
@@ -0,0 +1,272 @@
...
+
+	switch resp := ack.Response.(type) {
+	case *channeltypes.Acknowledgement_Result:
+		im.keeper.Logger(ctx).Info("received an Acknowledgement message", "result", string(resp.Result))
+		ctx.EventManager().EmitEvent(
```
RafilxTenfen, line 230: should we emit typed events?

x/zoneconcierge/README.md
```diff
@@ -0,0 +1,500 @@
...
+the long range attacks, thus unbonding requests in this prefix can be safely
+finished, leading to much shorter unbonding period compared to that in existing
+PoS blockchains (e.g., 21 days in Cosmos SDK chains).
+
+## State
+
+The Zone Concierge module keeps handling IBC headers of PoS blockchains, and
+maintains the following KV stores.
+
+### Parameters
```
RafilxTenfen, line 193: structures are out of date with the protos

x/btcstaking/keeper/msg_server.go
SebastianElvis: We can do this because the protocol requires that every BTC delegation has to at least mult-stake to a Babylon FP ([ref](https://github.com/babylonlabs-io/babylon/blob/1a0f10f4f7918a5b792a42dcc1b18c6b13c17d73/x/btcstaking/keeper/btc_delegations.go#L261-L264)). This is a product decision in order to protect Babylon value proposition.

x/btcstaking/keeper/finality_providers.go
SebastianElvis: Added a TODO in the code

x/btcstaking/keeper/grpc_query.go
SebastianElvis: Suggest we iterate over this part based on the feedbacks from web team later. It's not clear to me whether this interface should dedicate to Babylon FP or be a unified one for both Babylon/consumer FP. Wdyt?

x/zoneconcierge/types/genesis.go
SebastianElvis: Created https://github.com/babylonlabs-io/babylon/issues/426 to track this. Good point

x/btcstkconsumer/client/cli/tx.go
SebastianElvis: reworded

x/btcstkconsumer/genesis.go
SebastianElvis: Sg, covered in https://github.com/babylonlabs-io/babylon/issues/426

x/btcstkconsumer/types/chain_register.go
SebastianElvis: This is the default value, and the CLI input will be taken [here](https://github.com/babylonlabs-io/babylon/blob/1a0f10f4f7918a5b792a42dcc1b18c6b13c17d73/x/btcstkconsumer/client/cli/tx.go#L58-L61)

SebastianElvis: Good point, let's enforce this in CLI

x/btcstkconsumer/types/errors.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	sdkerrors "cosmossdk.io/errors"
+)
+
+// x/btcstkconsumer module sentinel errors
+var (
+	ErrInvalidSigner                = sdkerrors.Register(ModuleName, 1100, "expected gov account as only signer for proposal message")
+	ErrConsumerNotRegistered        = sdkerrors.Register(ModuleName, 1101, "consumer not registered")
+	ErrInvalidConsumerRegister      = sdkerrors.Register(ModuleName, 1102, "invalid consumer register")
```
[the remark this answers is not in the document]
SebastianElvis: Which plugin do you use for finding all these unused stuff? just curious

x/btcstkconsumer/keeper/keeper.go
SebastianElvis: Let's keep them for now as I see every module depends on accountKeeper / bankKeeper by default. Wdyt?

x/btcstkconsumer/types/keys.go
```diff
@@ -0,0 +1,26 @@
...
+
+const (
+	// ModuleName defines the module name
+	ModuleName = "btcstkconsumer"
+
+	// StoreKey defines the primary module store key
+	StoreKey = ModuleName
+
+	// MemStoreKey defines the in-memory store key
+	MemStoreKey = "mem_btcstkconsumer"
```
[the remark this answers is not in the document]
SebastianElvis: Every module has `MemStoreKey`. Maybe keep it for now in case we need a mem KV store for this module?

x/btcstkconsumer/types/keys.go
SebastianElvis: Same, many modules have this, not sure why though

x/finality/client/cli/tx.go
SebastianElvis: Oops maybe this is an error due to rebase. Adding it back

x/zoneconcierge/README.md
```diff
@@ -0,0 +1,500 @@
...
+
+- **Phase 1 integration:** Babylon receives PoS blockchain headers via standard
+  `MsgUpdateClient` messages in IBC light client protocol, timestamps them, and
+  functions as a canonical chain oracle for the PoS blockchain.
+  [Babylonscan](https://babylonscan.io/) shows PoS blockchains with phase 1
+  integration.
+- **Phase 2 integration:** In addition to phase 1, phase 2 allows a PoS
+  blockchain to receive BTC timestamps from Babylon via an IBC channel, such
+  that the PoS blockchain can use BTC timestamps to detect and resolve forks, as
+  well as other use cases such as Bitcoin-assisted fast unbonding.
```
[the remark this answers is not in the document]
SebastianElvis: Subjected to revision. Created https://github.com/babylonlabs-io/babylon/issues/427 and https://github.com/babylonlabs-io/babylon/issues/428

x/zoneconcierge/keeper/grpc_query.go
SebastianElvis: Added a TODO for paginating this for loop. This is not pressiong though as we expect the integration to be permissioned in the beginning and there won't be too many of them

x/zoneconcierge/keeper/header_handler.go
SebastianElvis: Yes

x/zoneconcierge/keeper/ibc_header_decorator.go
SebastianElvis: added docstring for both

x/zoneconcierge/keeper/keeper.go
SebastianElvis: ditto, maybe keep them for now as by default modules depend on these two?

x/zoneconcierge/module_ibc.go
SebastianElvis: added TODO on emitting typed event for each `EmitEvent` occurence

x/zoneconcierge/README.md
SebastianElvis: captured in https://github.com/babylonlabs-io/babylon/issues/428

SebastianElvis:
Thanks @RafilxTenfen for the second wave of comments. Addressed them in https://github.com/babylonlabs-io/babylon/pull/424

proto/babylon/btcstaking/v1/events.proto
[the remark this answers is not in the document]
gitferry: Makes sense. All good then

x/btcstkconsumer/keeper/keeper.go
RafilxTenfen: In my opinion it only increases the complexity if it is not being used, but it is your call on this ^^

x/btcstaking/client/cli/tx.go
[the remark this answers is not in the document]
gitferry: yeah, so in this case, `[x]` is a variable, which is a string, right? I worry that people may think each fp pk is a separate arg and type them as `"fp1_pk", "fp2_pk"` but we expect `"fp1_pk, fp2_pk", right?

x/btcstaking/keeper/msg_server.go
RafilxTenfen: yeap, but we would be tracking for multiple fps https://github.com/babylonlabs-io/babylon/blob/be9eb6b9f4db544eb35db8b4767852d12910b9eb/x/finality/keeper/power_dist_change.go#L422
while the only ones that matter for the incentives reward tracker are the FPs from Babylon, not from consumers

x/btcstaking/keeper/btc_consumer_delegations.go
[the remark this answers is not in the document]
gitferry: Seems we should do
```suggestion
	for _, fpBTCPK := range btcDel.FpBtcPkList {
		fpBTCPK := fpBTCPK
```
no need to have `i`

x/btcstaking/keeper/finality_providers.go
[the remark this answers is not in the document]
gitferry: hmm OK, seems no non-determinism. This still makes me nervous as it seems introduce an assumption that `AddBTCStakingConsumerEvents` will not cause non-determinism even if it is called in a random order

> humm, if for some reason the consumer event fails AddBTCStakingConsumerEvents it will stop the execution and with different orders it will not be deterministic

If errors are observed during a tx execution, all the state changes will be rolled back

x/btcstaking/client/cli/tx.go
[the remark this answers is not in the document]
SebastianElvis: > yeah, so in this case, [x] is a variable, which is a string, right?

Yes

> I worry that people may think each fp pk is a separate arg and type them as "fp1_pk", "fp2_pk" but we expect `"fp1_pk, fp2_pk", right?

This is something I'm not sure how to make it clear :sweat_smile: let's add a TODO here for now?

x/btcstaking/keeper/finality_providers.go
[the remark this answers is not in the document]
SebastianElvis: Good catch. I think if `AddBTCStakingConsumerEvents` fails then the code should panick. If `AddBTCStakingConsumerEvents` fails, it means that the event type is not recognised [here](https://github.com/babylonlabs-io/babylon/blob/62120b1be8fd7682bf676b88639a3160801e09eb/x/btcstaking/keeper/btc_staking_consumer_events.go#L37-L38), which has to be a programming error

x/btcstaking/keeper/btc_consumer_delegations.go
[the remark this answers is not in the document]
SebastianElvis: This seems to be same as the original code right?

x/btcstaking/keeper/msg_server.go
SebastianElvis: Then this means the `processRewardTracker` function or the new lazy reward protocol is not compatible with multi-staking since it assumes all FPs are Babylon FPs. In this case wdyt we modify `processRewardTracker` and `ProcessAllPowerDistUpdateEvents` to only handle Babylon FPs, like here https://github.com/babylonlabs-io/babylon/pull/424/commits/1e825db167b72efc5468fff24799606bf4ef91e2?

proto/babylon/zoneconcierge/v1/zoneconcierge.proto
SebastianElvis: Hmm I see what you mean, but in this case wouldn't IBC modules reject the duplicated submission before the tx goes to the PostHandler?

wasmbinding/wasm.go
maurolacy: Not sure. I would say, no.

x/btcstaking/keeper/msg_server.go
RafilxTenfen: sounds good to me

maurolacy (approved):
LGTM.

Reviewed more or less everything, except for the re-added `zoneconcierge` module. Perhaps it's a good idea to compare the `x/zoneconcierge` directory here with the latest version (before deletion) from `main`, to confirm we're not losing any potential new addition.

x/btcstkconsumer/keeper/keeper.go
KonradStaniec: this is a bit artefact of the past i.e first Babylon modules we created, we copied them from ignite tool (from what I remember) and ignite by default added those keepers, so we kinda rolled with it as we did not know any better back then.

We should clean this up, and every keeper should only depend on what is needed. Maybe separate task ?

CHANGELOG.md
```diff
@@ -31,7 +31,35 @@ Ref: https://keepachangelog.com/en/1.0.0/
...
+## [Unreleased](https://github.com/babylonchain/babylon-private/tree/HEAD)
+
+[Full Changelog](https://github.com/babylonchain/babylon-private/compare/euphrates-0.2.0-rc.0...HEAD)
+
+
+## [euphrates-v0.2.0-rc.0](https://github.com/babylonchain/babylon-private/tree/euphrates-v0.2.0-rc.0) (2024-05-17)
+
+[Full Changelog](https://github.com/babylonchain/babylon-private/compare/euphrates-0.1.0-rc.1...euphrates-v0.2.0-rc.0)
+
+## [euphrates-0.1.0-rc.1](https://github.com/babylonchain/babylon-private/tree/euphrates-0.1.0-rc.1) (2024-03-25)
```
KonradStaniec, line 43: tbh I would delete all those sub points and in `## Unreleased` section add `Babylon Multi staking support`
pointing to this pr. (probably under State breaking sub titlte`)

And in this pr description add links to most relevat pr's

KonradStaniec (approved):
lgtm! just minor comments about changelog.

Lets wait for other core team members approval before merging though.

RafilxTenfen (approved):
LGTM after the merge of https://github.com/babylonlabs-io/babylon/pull/424

Lazar955 (approved):
Great work! Let's not forget to merge #424

x/btcstaking/client/cli/tx.go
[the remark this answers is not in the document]
SebastianElvis: Hmm it's a bit difficult for someone to know the string should be pk1,pk2,... upon seeing `[fp_pks]` :sweat:

SebastianElvis:
Created https://github.com/babylonlabs-io/babylon/issues/431 to track the task for tightening the cross-module dependencies. Cannot find the discussion anymore so put it here

proto/babylon/zoneconcierge/v1/zoneconcierge.proto
RafilxTenfen: not sure, it is worth to test it out

SebastianElvis: Duplicated header will get rejected, see https://github.com/cosmos/ibc-go/blob/91169d134d659b87acaf4903652cdcb1a55f5657/modules/light-clients/07-tendermint/update.go#L85-L91

x/btcstaking/keeper/btc_consumer_delegations.go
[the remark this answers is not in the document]
gitferry: Yes but avoiding the index `i`

gitferry (approved):
Thanks for addressing all the comments. No blocker from my side. Let's go!

PR #419, opened 2025-01-17
fix: swagger add new modules to config

RafilxTenfen:
- Each module need to be listed in the config to generate the data in swagger


Lazar955 (approved):
utACK

KonradStaniec (approved):

KonradStaniec:
hmm is there linter check that would check the swagger re-generation on each pr ? 🤔  (otherwise this problem will return really quickly 😅 )

PR #429, opened 2025-01-21
chore: remove `cosmos/relayer` dependency 

Lazar955:
- Migrates the minimal amount of code needed from the [relayer](https://github.com/cosmos/relayer) repo
- Removes dependency for `cosmos/relayer`
- Removes dependency for `strangelove-ventures/cometbft-client`

~Todo:~
 - ~As this is modified code from the [relayer](https://github.com/cosmos/relayer), what license if any should we include in the header of those files? I've also modified them (deleted a bunch of stuff, and added more code quality improvements)~
 - ~Is feegrant.go needed, feels like it can be removed and other parts related to it~
 - ~Future PRs can remove references to strangelove cometbft client, tho we do have to satisfy the `CometRPC interface`, `ABCIClient interface` etc. (not sure if it's worth the extra effort)~
 - ~Decide on the name of the relayer package, the clients would currently use `relayerclient.Whatever`~
 
 [Closes](https://github.com/babylonlabs-io/pm/issues/149)


KonradStaniec:
> Decide on the name of the relayer package, the clients would currently use relayerclient.Whatever

tbh I would name it `babylonclient` or `cosmosclient`

>As this is modified code from the [relayer](https://github.com/cosmos/relayer), what license if any should we include in the header of those files? I've also modified them (deleted a bunch of stuff, and added more code quality improvements)

Something like that - https://law.stackexchange.com/questions/59999/can-i-copy-pieces-of-apache-licensed-source-code-if-i-attribute ? (though I am not a license expert) 

>Future PRs can remove references to strangelove cometbft client, tho we do have to satisfy the CometRPC interface, ABCIClient interface etc. (not sure if it's worth the extra effort)

I think one of the goals of this is to, remove dependency on `strangelove cometbft client` , so maybe lets:
- create feature branch with `feat/` prefix for the whole feature
- make this pr to the feature branch
- then create next pr removing dependency on `strangelove cometbft clien`
- then create pr from feature branch to main

wdyt ?

Lazar955:
@KonradStaniec 
> Or maybe remove `strangelove cometbft client` directly in this pr. wdyt ?

Removed it in this PR, it wasn't a big effort, [commit](https://github.com/babylonlabs-io/babylon/pull/429/commits/0ba6efc2e7c3c0532315d4098be5f26e66e82fb6)

Lazar955:
> The package `relayerclient` could be in a separate repository in my opinion
@RafilxTenfen 
For now, I would keep it within the Babylon repo. The only "issue" I personally see is that updating client code would mean that we have to release a new version of Babylon, I don't see this as something that should happen often (as up to this point it wasn't the case), as the only real dependencies of that new package are cometbft and cosmos SDK. If we start updating the client often and as a consequence we have to release new version of Babylon, I'm down for splitting this into it's own repo.

RafilxTenfen:
> > The package `relayerclient` could be in a separate repository in my opinion
> > @RafilxTenfen
> > For now, I would keep it within the Babylon repo. The only "issue" I personally see is that updating client code would mean that we have to release a new version of Babylon, I don't see this as something that should happen often (as up to this point it wasn't the case), as the only real dependencies of that new package are cometbft and cosmos SDK. If we start updating the client often and as a consequence we have to release new version of Babylon, I'm down for splitting this into it's own repo.

Yeah, but also like we are constantly getting babylon node code checked and this is basically not a part of it, just a consumer 
Anyway, if we intend to maintain this code in the Babylon repo, the PR looks good to me ^^

RafilxTenfen (approved):
LGTM

KonradStaniec (approved):
LGTM! 

One question, did you test this branch with some service (like use this version in staker and run e2e tests) to see whether change is really not breaking ? 

Also, lets merge this one after - https://github.com/babylonlabs-io/babylon/pull/402

Lazar955:
> LGTM!
> 
> One question, did you test this branch with some service (like use this version in staker and run e2e tests) to see whether change is really not breaking ?
> 
> Also, lets merge this one after - #402

I've tested it with Vigilante e2e tests

client/babylonclient/codec.go
```diff
@@ -0,0 +1,17 @@
...
+
+package babylonclient
+
+import (
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/codec"
+	"github.com/cosmos/cosmos-sdk/codec/types"
+)
+
+type Codec struct {
```
RafilxTenfen, line 12: This seems the same as our [EncodingConfig](https://github.com/babylonlabs-io/babylon/blob/00bf5307f8aa783aff16ecae4a0bae2bdfbcdff6/app/params/encoding.go#L11)

Can we use that?

client/client/client.go
```diff
@@ -1,21 +1,20 @@
...
 	rpchttp "github.com/cometbft/cometbft/rpc/client/http"
-	"github.com/cosmos/relayer/v2/relayer/chains/cosmos"
 	"go.uber.org/zap"
 )
 
 type Client struct {
 	*query.QueryClient
 
-	provider *cosmos.CosmosProvider
+	provider *babylonclient.CosmosProvider
```
RafilxTenfen, line 17: Can we make this public? 
```suggestion
	Provider *babylonclient.CosmosProvider
```

client/babylonclient/account.go
```diff
@@ -0,0 +1,78 @@
...
+func (cc *CosmosProvider) EnsureExists(clientCtx client.Context, addr sdk.AccAddress) error {
+	if _, err := cc.GetAccount(clientCtx, addr); err != nil {
+		return err
+	}
+	return nil
+}
+
+// GetAccountNumberSequence returns sequence and account number for the given address.
+// It returns an error if the account couldn't be retrieved from the state.
+func (cc *CosmosProvider) GetAccountNumberSequence(clientCtx client.Context, addr sdk.AccAddress) (uint64, uint64, error) {
```
RafilxTenfen, line 72: can we make a function that increases the account sequence also in the WalletState

client/babylonclient/tx.go
```diff
@@ -0,0 +1,650 @@
...
+		return
+	}
+	sequenceGuard.NextAccountSequence = nextSeq
+}
+
+// SendMessagesToMempool simulates and broadcasts a transaction with the given msgs and memo.
+// This method will return once the transaction has entered the mempool.
+// In an async goroutine, will wait for the tx to be included in the block unless asyncCtx exits.
+// If there is no error broadcasting, the asyncCallback will be called with success/failure of the wait for block inclusion.
+func (cc *CosmosProvider) SendMessagesToMempool(
```
RafilxTenfen, line 344: Can we add another option here to specify the sequence in the msg and if it is specified the PrepareFactory would not load from the account sequence

RafilxTenfen:
Added a new comments here https://github.com/babylonlabs-io/babylon/pull/429 due to comment in https://github.com/babylonlabs-io/covenant-emulator/pull/98#discussion_r1926541104

client/babylonclient/tx.go
Lazar955: Hm it's going to be a bit much returning and passing account sequence to build messages, maybe we a separate SendMessagesToMempool?

client/babylonclient/codec.go
Lazar955: sure, nice spot

client/babylonclient/tx.go
RafilxTenfen: could be as well ;D

Lazar955: Maybe all this is a bit out of the scope of this PR, comparing what's needed in [cov emulator PR](https://github.com/babylonlabs-io/covenant-emulator/pull/98)?

RafilxTenfen: it could be done in another PR, but just adding the account sequence would help to reduce a lot the code copy in covd

PR #435, opened 2025-01-22
chore: modified to gentx to generate MsgWrappedCreateValidator

RafilxTenfen:
- Copy partial implementation from sdk [gentx](https://github.com/cosmos/cosmos-sdk/blob/eba0b9e9c2b5da19ccb8b4b189c1f0747830a8fa/x/genutil/client/cli/gentx.go#L31) but modified to generate `MsgWrappedCreateValidator` and create/add BLS key if needed
- Removed commands: 
  - `babylond gen-helpers create-bls`
  - `babylond gen-helpers add-bls`
- Modified `x/checkpointing` msg server to call the `x/staking` msg server on receiving `MsgWrappedCreateValidator` if the context block height is zero (genutil) calls the handler

It partially helps to remove the `x/staking` msg server from the router


RafilxTenfen:
@babylon-devops after this PR the commands `babylond gen-helpers create-bls` and `babylond gen-helpers add-bls` will not be available and executing the `babylond gentx ...` will handle 

Also, `babylond create-bls-key` prior to gentx will be optional, if the bls key is not found it creates one

cmd/babylond/cmd/genhelpers/collect.go
```diff
@@ -0,0 +1,272 @@
...
+	return printInfo{
+		Moniker:    moniker,
+		ChainID:    chainID,
+		NodeID:     nodeID,
+		GenTxsDir:  genTxsDir,
+		AppMessage: appMessage,
+	}
+}
+
+func displayInfo(info printInfo) error {
```
Lazar955, line 263: nit: could we return a string/bytes, and the the `cmd.Printf`

cmd/babylond/cmd/bls/create_bls_key.go
```diff
@@ -56,20 +57,20 @@ $ babylond create-bls-key %s1f5tnl46mk4dfp4nx3n2vnrvyw2h2ydz6ykhk3r --home ./
 	return cmd
 }
 
-func CreateBlsKey(home string, addr sdk.AccAddress) error {
+func CreateBlsKey(home string, addr sdk.AccAddress) (wrappedPV *privval.WrappedFilePV, err error) {
```
Lazar955, line 60: nit: don't see the benefit of the named return here

cmd/babylond/cmd/genhelpers/collect.go
Lazar955: in tests when can verify the output (if make sense)

cmd/babylond/cmd/validate_genesis.go
```diff
@@ -112,8 +111,8 @@ func CheckCorrespondence(ctx client.Context, genesis map[string]json.RawMessage,
 		if len(msgs) == 0 {
 			return errors.New("invalid genesis transaction")
 		}
-		msgCreateValidator := msgs[0].(*stakingtypes.MsgCreateValidator)
-		if _, exists := addresses[msgCreateValidator.ValidatorAddress]; !exists {
+		msgCreateValidator := msgs[0].(*checkpointingtypes.MsgWrappedCreateValidator)
```
Lazar955, line 114:
```suggestion
		msgCreateValidator, ok := msgs[0].(*checkpointingtypes.MsgWrappedCreateValidator)
		// check !ok here
```

cmd/babylond/cmd/validate_genesis.go
```diff
@@ -86,7 +85,7 @@ func validateGenDoc(importGenesisFile string) (*cmttypes.GenesisDoc, error) {
 // CheckCorrespondence checks that each genesis tx/BLS key should have one
 // corresponding BLS key/genesis tx
 func CheckCorrespondence(ctx client.Context, genesis map[string]json.RawMessage, validator genutiltypes.MessageValidator) error {
-	checkpointingGenState := types.GetGenesisStateFromAppState(ctx.Codec, genesis)
+	checkpointingGenState := checkpointingtypes.GetGenesisStateFromAppState(ctx.Codec, genesis)
 	gks := checkpointingGenState.GetGenesisKeys()
 	genTxState := genutiltypes.GetGenesisStateFromAppState(ctx.Codec, genesis)
 	addresses := make(map[string]struct{}, 0)
```
Lazar955, line 91:
```suggestion
	addresses := make(map[string]struct{}, len(gks))
```

Lazar955: Also is this check necessary/possible to happen after the loop, I don't think so. `if len(addresses) != len(gks) {`

cmd/babylond/cmd/genhelpers/collect.go
RafilxTenfen: removed the display of info func

cmd/babylond/cmd/bls/create_bls_key.go
RafilxTenfen: removed named return

cmd/babylond/cmd/validate_genesis.go
RafilxTenfen: basically, it verifies that all the addresses inside gks are unique, I will not modify to add the size to the map

cmd/babylond/cmd/genhelpers/collect.go
```diff
@@ -0,0 +1,266 @@
...
+	"github.com/cosmos/cosmos-sdk/x/genutil/types"
+)
+
+const flagGenTxDir = "gentx-dir"
+
+// CollectGenTxsCmd - return the cobra command to collect genesis transactions
+// Code coppied from https://github.com/cosmos/cosmos-sdk/blob/46be66ed4205bf109d93704dd14c0822d048972e/x/genutil/client/cli/collect.go#L22
+// The reason is that the genutil parses the gen tx message as stakingtypes.MsgCreateValidator to check balances in
+// https://github.com/cosmos/cosmos-sdk/blob/46be66ed4205bf109d93704dd14c0822d048972e/x/genutil/collect.go#L134
+func CollectGenTxsCmd(genBalIterator types.GenesisBalancesIterator, defaultNodeHome string, validator types.MessageValidator, valAddrCodec sdkruntime.ValidatorAddressCodec) *cobra.Command {
```
KonradStaniec, line 37: as I understand this is copy paste from cosmos-sdk, so maybe we need some licence attribution on top of the file ?

KonradStaniec (approved):

PR #436, opened 2025-01-22
chore: epoch handle staking msg server

RafilxTenfen:
- Removed `x/staking` msgs from router
- Add msg server from `x/staking` to epoch keeper to call the necessary functions directly instead of from the router
- Add `MsgWrappedEditValidator` and `MsgWrappedStakingUpdateParams` to epoching txs so all the msgs from staking can be called even with the router removed
- Removed `x/staking` tx subcommand


cmd/babylond/cmd/root.go
```diff
@@ -130,11 +132,46 @@ func NewRootCmd() *cobra.Command {
 		TextualCoinMetadataQueryFn: authtxconfig.NewGRPCCoinMetadataQueryFn(initClientCtx),
 	}
 
+	EnhanceRootCommandWithoutTxStaking(autoCliOpts, rootCmd)
+
+	return rootCmd
+}
+
+// EnhanceRootCommandWithoutTxStaking excludes staking tx commands
+func EnhanceRootCommandWithoutTxStaking(autoCliOpts autocli.AppOptions, rootCmd *cobra.Command) {
```
RafilxTenfen, line 141: since we don't have the `x/staking` msg server routes, we still add the staking commands and later manually finds to remove it

It is done this way to avoid need to overwrite `x/staking` app module

app/keepers/keepers.go
```diff
@@ -412,7 +412,7 @@ func (ak *AppKeepers) InitKeepers(
 		appparams.AccGov.String(),
 	)
 
-	wasmOpts = append(owasm.RegisterCustomPlugins(&ak.EpochingKeeper, &ak.CheckpointingKeeper, &ak.BTCLightClientKeeper, &ak.ZoneConciergeKeeper), wasmOpts...)
+	wasmOpts = append(owasm.RegisterCustomPlugins(&epochingKeeper, &ak.CheckpointingKeeper, &ak.BTCLightClientKeeper, &ak.ZoneConciergeKeeper), wasmOpts...)
```
RafilxTenfen, line 415: ak.EpochingKeeper was not set yet

x/epoching/keeper/epoch_msg_queue_test.go
```diff
@@ -97,7 +97,7 @@ func FuzzHandleQueuedMsg_MsgWrappedDelegate(f *testing.F) {
 		require.Equal(t, uint64(0), lc.ValLife[0].BlockHeight)
 
 		// delegate a random amount of tokens to the validator
-		numNewDels := r.Int63n(1000) + 1
+		numNewDels := r.Int63n(15) + 1
```
RafilxTenfen, line 100: if left as 1k new dels it breaks the `cosmos-sdk/types.TypedEventToEvent` to parse the event, was not able to find the root of the issue 

```shell
goroutine 41 [chan receive, 1 minutes]:
testing.(*F).Fuzz.func1({0x507ab20, 0xc0018a5c00}, {{0x0, 0x0}, {0xc0006d9420, 0x6}, {0x0, 0x0, 0x0}, {0xc000f98c20, ...}, ...})
	/home/rafilx/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.1.linux-amd64/src/testing/fuzz.go:337 +0x58a
testing.(*F).Fuzz(0xc0018a5c00, {0x316d8a0, 0x4bfe638})
	/home/rafilx/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.1.linux-amd64/src/testing/fuzz.go:402 +0xab7
github.com/babylonlabs-io/babylon/x/epoching/keeper_test.FuzzHandleQueuedMsg_MsgWrappedDelegate(0xc0018a5c00)
	/home/rafilx/projects/github.com/babylonlabs-io/babylon/x/epoching/keeper/epoch_msg_queue_test.go:67 +0x35
testing.fRunner(0xc0018a5c00, 0x4bfe490)
	/home/rafilx/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.1.linux-amd64/src/testing/fuzz.go:724 +0xb9
created by testing.runFuzzTests in goroutine 1
	/home/rafilx/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.1.linux-amd64/src/testing/fuzz.go:527 +0x838

goroutine 42 [runnable]:
encoding/json.stateInStringEsc(0xc06c9f4388, 0x5c?)
	/home/rafilx/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.1.linux-amd64/src/encoding/json/scanner.go:353 +0x1a5
encoding/json.checkValid({0xc1bb52a000, 0x7ef6, 0x8?}, 0xc06c9f4388)
	/home/rafilx/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.1.linux-amd64/src/encoding/json/scanner.go:35 +0x9d
encoding/json.Unmarshal({0xc1bb52a000, 0x7ef6, 0x8000}, {0x30eaae0, 0xc0017af498})
	/home/rafilx/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.1.linux-amd64/src/encoding/json/decode.go:102 +0x54
github.com/cosmos/cosmos-sdk/types.TypedEventToEvent({0x5096970, 0xc06c1dd200})
	/home/rafilx/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.50.11/types/events.go:97 +0xc5
github.com/cosmos/cosmos-sdk/types.(*EventManager).EmitTypedEvent(0xc001eb51a0, {0x5096970?, 0xc06c1dd200?})
	/home/rafilx/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.50.11/types/events.go:64 +0x3d
github.com/babylonlabs-io/babylon/x/epoching.EndBlocker({0x50bed48, 0xc0021a7188}, {{0x50dd160, 0xc001200950}, {0x507b180, 0xc000fb01d8}, {0x50abcd0, 0xc000719878}, {0x5090f28, 0xc000778508}, ...})
	/home/rafilx/projects/github.com/babylonlabs-io/babylon/x/epoching/abci.go:116 +0x9ee
```

app/app.go
```diff
@@ -588,6 +586,45 @@ func NewBabylonApp(
...
+		panic(err)
+	}
+
+	// adds the staking module back it back
+	app.ModuleManager.Modules[stakingtypes.ModuleName] = stkModTemp
+}
+
+// RegisterStakingQueryAndMigrations registrates in the configurator
+// the x/staking query server and its migrations
+func (app *BabylonApp) RegisterStakingQueryAndMigrations() {
```
RafilxTenfen, line 606: We still need to register the staking migrations, so every new migration in https://github.com/cosmos/cosmos-sdk/blob/32969cf7cc602bb380f78dee23f0f0a747728791/x/staking/module.go#L141 we need to apply here as well

x/epoching/keeper/msg_server.go
```diff
@@ -16,13 +18,135 @@ type msgServer struct {
...
+	}
+	blockTime := ctx.HeaderInfo().Time
+
+	txid := tmhash.Sum(ctx.TxBytes())
+	queuedMsg, err := types.NewQueuedMessage(blockHeight, blockTime, txid, msg)
+	if err != nil {
+		return nil, err
+	}
+
+	ms.EnqueueMsg(ctx, queuedMsg)
```
SebastianElvis, line 89: Why do we need to defer the `EditValidator` operation to the end of the epoch, rather than executing it directly?

x/epoching/keeper/msg_server.go
```diff
@@ -16,13 +18,135 @@ type msgServer struct {
...
```
SebastianElvis, line 131: This does not seem necessary either as the protocol won't use these parameters in the middle of epochs. But it does not bother to update it upon epoch boundary for extra caution

x/epoching/keeper/keeper.go
```diff
@@ -20,7 +19,6 @@ type (
 		hooks        types.EpochingHooks
 		bk           types.BankKeeper
 		stk          types.StakingKeeper
```
SebastianElvis, line 21: Does it make sense to only pass `stk` in the `NewKeeper` function (without passing `stkMsgServer`), and instead use `stakingkeeper.NewMsgServerImpl(stakingKeeper)` to construct the `stkMsgServer` in tke `NewKeeper` function?

x/epoching/keeper/epoch_msg_queue.go
```diff
@@ -107,139 +108,150 @@ func (k Keeper) GetCurrentEpochMsgs(ctx context.Context) []*types.QueuedMessage
...
-	if err != nil {
-		return result, err
-	}
+	return sdk.WrapServiceResult(ctx, res, err)
+}
 
-	// release the cache
-	msCache.Write()
+func (k Keeper) runUnwrappedMsg(ctx sdk.Context, qMsg *types.QueuedMessage) (proto.Message, error) {
+	msgSvrCtx, msCache := cacheTxContext(ctx, qMsg.TxId, qMsg.MsgId, qMsg.BlockHeight)
```
SebastianElvis, line 120: How do you compare calling staking keeper functions directly v.s. using the msgSvr functions in the `runUnwrappedMsg` function?

The function for constructinig the cache tx makes me worried a bit as I'm not sure whether txid/msgid/blockheight are the only things we need there

x/epoching/types/msg.go
```diff
@@ -8,6 +8,7 @@ import (
 // ensure that these message types implement the sdk.Msg interface
 var (
 	_ sdk.Msg = &MsgWrappedDelegate{}
+	_ sdk.Msg = &MsgWrappedEditValidator{}
```
SebastianElvis, line 11: how abt wrapped staking param update?

SebastianElvis (commented):
Still parsing the whole thing, some initial comments

x/epoching/keeper/msg_server.go
RafilxTenfen: That is a good question, I also was thinking about it when writing it, but I thought since it could change the min self delegation amount and the commission was better to defer to the end of epoch

What do you think?

x/epoching/keeper/msg_server.go
RafilxTenfen: it might be good to have staking MsgUpdateParams done by the epoch as it modifies values referred to validator creation like the minimum commission allowed

x/epoching/keeper/keeper.go
RafilxTenfen: it could be, but then in the epoching keeper I would expect to receive the pointer to the staking module instead of the expected keeper staking interface

x/epoching/types/msg.go
RafilxTenfen: good catch, add in https://github.com/babylonlabs-io/babylon/pull/436/commits/3d78a36883d46477335c6a9785d2ab7241dea3c0

x/epoching/keeper/epoch_msg_queue.go
SebastianElvis: My bad, just realised that this function was same as the previous implementation and it's from baseapp... All good

app/app_test.go
```diff
@@ -94,3 +95,22 @@ func TestUpgradeStateOnGenesis(t *testing.T) {
...
+
+	app := NewBabylonAppWithCustomOptions(t, false, signer, SetupOptions{
+		Logger:             logger,
+		DB:                 db,
+		InvCheckPeriod:     0,
+		SkipUpgradeHeights: map[int64]bool{},
+		AppOpts:            TmpAppOptions(),
+	})
+
+	msg := &stktypes.MsgCreateValidator{}
```
SebastianElvis, line 112: Wdyt we test all messages under staking module rather than just one of them?

SebastianElvis (approved):
Great work!

app/app.go
```diff
@@ -588,6 +586,47 @@ func NewBabylonApp(
...
+	cfg, stkK := app.configurator, app.StakingKeeper
+	stkq := stakingkeeper.NewQuerier(stkK)
+
+	stakingtypes.RegisterQueryServer(cfg.QueryServer(), stkq)
+
+	ls := app.GetSubspace(stakingtypes.ModuleName)
+	m := stakingkeeper.NewMigrator(stkK, ls)
+	if err := cfg.RegisterMigration(stakingtypes.ModuleName, 1, m.Migrate1to2); err != nil {
```
KonradStaniec, line 616: ahh this makes me a bit worried, as if we bump cosmos-sdk and there will be new migration we may forget it 🤔 but I do not see easy way to improve on that. 

Maybe some additional docs to `epoching` module readme  or something like that (about those migrations being in app.go)

KonradStaniec (approved):
lgtm! great job figuring it out

Lazar955 (approved):

PR #445, opened 2025-01-27
Reject forks if the first header is already known

KonradStaniec:
Fixes: https://github.com/babylonlabs-io/pm/issues/159

We will reject BTC forks, if the first header is already known. 

This is consensus breaking change, but on our testnet it should be non-breaking as our vigilante reporter always reports only new btc headers prefix/


Lazar955 (approved):

RafilxTenfen (approved):
:rocket:

PR #466, opened 2025-02-02
chore: add test for block send of ubbn and still create btc delegation

RafilxTenfen:


RafilxTenfen:
Can we merge this? @babylonlabs-io/core-dev

GAtom22 (approved):
LGTM! Great work @RafilxTenfen!
One small nice-to-have would be to check the fee payer and the tx sender balances after the tx

PR #467, opened 2025-02-03
feat: BLS keystore improvement

### LINKED ISSUE #336: Separate BLS key from `priv_validator_key.json`

gitferry:
Closes #336


Makefile
```diff
@@ -86,6 +86,11 @@ else
   BUILD_TAGS += mainnet
 endif
 
+# Handles the inclusion of e2e upgrade in binary
```
RafilxTenfen, line 89: I think we can remove this part already, right?

RafilxTenfen: ohh, this is upgrade of bls keys, not software upgrade, my bad

app/signer/bls.go
```diff
@@ -0,0 +1,209 @@
...
+func GenBls(keyFilePath, passwordFilePath, password string) *Bls {
+	pv := NewBls(bls12381.GenPrivKey(), keyFilePath, passwordFilePath)
+	pv.Key.Save(password)
+	return pv
+}
+
+// LoadBls returns a Bls after loading the erc2335 type of structure
+// from the file and decrypt it using a password.
+func LoadBls(keyFilePath, passwordFilePath string) *Bls {
+	passwordBytes, err := os.ReadFile(passwordFilePath)
```
RafilxTenfen, line 79: it is possible to make it empty password? like no password

test/e2e/initialization/export.go
```diff
@@ -10,15 +12,16 @@ type ChainMeta struct {
...
+	Name          string            `json:"name"`
+	ConfigDir     string            `json:"configDir"`
+	Mnemonic      string            `json:"mnemonic"`
+	PublicAddress string            `json:"publicAddress"`
+	WalletName    string            `json:"walletName"`
+	PublicKey     []byte            `json:"publicKey"`
+	PrivateKey    []byte            `json:"privateKey"`
+	PeerId        string            `json:"peerId"`
+	IsValidator   bool              `json:"isValidator"`
+	CometPrivKey  cmtcrypto.PrivKey `json:"cometPrivKey"`
```
RafilxTenfen, line 24: having an interface here might be difficult to parse as json.Unmarshal at chain initialization https://github.com/babylonlabs-io/babylon/blob/bf31f69ba05caf513df605280a66c26ac0c3004f/test/e2e/initialization/chain/main.go#L43

We should probably use primitive types

x/checkpointing/types/types.go
```diff
@@ -22,6 +22,12 @@ const (
 	BitmapBits = txformat.BitMapLength * 8 // 104 bits for 104 validators at top
 )
 
+// BlsSigner is an interface for signing BLS messages
+type BlsSigner interface {
+	SignMsgWithBls(msg []byte) (bls12381.Signature, error)
+	GetBlsPubkey() (bls12381.PublicKey, error)
```
RafilxTenfen, line 28:
```suggestion
	BlsPubkey() (bls12381.PublicKey, error)
```

RafilxTenfen (commented):
A few comments, but in general looks good to me 

I was able to start a local chain and migrate bls, but how do I know if my node is using the correct key? 

Should we have a document with instructions how to migrate the bls keys?

wnjoon:
@RafilxTenfen 

This migration is aimed at separating the BLS keys, previously included in the existing `priv_validator_key.json`, into a separate file. This means that the keys used by the node will remain unchanged.

If you need documentation on the migration process, please let me know. Below is a summary:

## Summary

### Description

Previously, the `priv_validator_key.json` file, created locally during the chain initialization process, stored both the BLS keys and the FilePV of Comet.

Due to this mixed structure, ExtendVote became dependent on the local `priv_validator_key.json` file. As a result, it was unusable in environments where the `priv_validator_key.json` file was not stored locally, such as with a remote signer.

To address this issue, we introduced the migrate-bls-key command. This command extracts the BLS key from `priv_validator_key.json` and stores it in a separate file, ensuring that `priv_validator_key.json` retains only the FilePV content generated during Comet’s initialization process.

Since the BLS key is stored in a file formatted according to ERC-2335, a password is required during the migration process. This password is stored in a separate file.

### Requirements

First, we assume that the node is operating normally.

If you run a node for the first time with a new version, the `babylond init` command will create a `priv_validator_key.json` file that stores Comet’s FilePV, along with a separate file for the BLS key.

Additionally, the node must be running a previous version that does not use a separate file structure. If you are already using a newer version that creates a separate file structure, running the `migrate-bls-key` command will result in the following error:

```sh
failed to load previous version of priv_validator_key.json in ...
```

### Process of migration

The migration will proceed in the following order:

```
1. the separation feature would be inserted in a new rc release
2. validator stops the current node
3. validator download the new release
4. validator run the migration cmd
5. restart the node and voting should be resumed
```

### How to test migration-bls-key

#### 1. Start the node using the `babylond` version before changing.

```sh
$ ./babylond init my-node --chain-id test-chain-1
$ ./babylond keys add validator1 --keyring-backend test
$ ./babylond add-genesis-account $(./babylond keys show validator1 -a --keyring-backend test) 2000000000000ubbn,2000000000000stake
$ ./babylond gentx validator1 1500000000000stake --chain-id test-chain-1 --keyring-backend test --fees 400ubbn
$ ./babylond collect-gentxs
$ ./babylond create-bls-key 
$ ./babylond gen-helpers create-bls $(./babylond keys show validator1 -a --keyring-backend test)
# copy content of gen-bls-x.json into checkpointing.genesis_keys field in genesis.json
$ ./babylond start
```

#### 2. Stop the running node.
#### 3. Execute migration from the changed version of `babylond`.

Before calling migration, check if the `priv_validator_key.json` file storing the two types of keys exists inside `<home>/.babylond/config`.

```sh
$ ./babylond migrate-bls-key
```

#### 4. Enter the password you will use to encrypt the bls key into the erc2335 structure.
#### 5. Verify that the `bls_key.json` and `bls_password.txt` files are correctly created in `<home>/.babylond/config`.
#### 6. Verify that `priv_validator_key.json` contains only keys for comet.
#### 7. Restart the node using changed version of `babylond`.

```sh
$ ./babylond start
```

#### Error cases
- If `priv_validator_key.json` does not exist when calling migration
- If `priv_validator_key.json` does not satisfy the previous version of a structure containing comet and bls keys.

app/signer/bls.go
wnjoon: @RafilxTenfen 
Currently, the logic for generating the BLS key is as follows:

- If the `bls-password` flag is set but left empty, an error occurs.
- If the `bls-password` flag is not used, a password prompt appears, allowing the user to enter a password. At this point, an empty value ("") can be entered.

Are you suggesting that the `bls-password` flag should be usable while still allowing an empty value?

Makefile
wnjoon: @RafilxTenfen 

The `e2e_upgrade` build tag is used for the `e2e_upgrade_v1` test.

When this build tag is active, the previous version’s `priv_validator_key.json` file is automatically migrated to the updated structure, where the comet and BLS keys are separated. This ensures compatibility with the new key file format. [code](https://github.com/babylonlabs-io/babylon/blob/3df2c9f38ed7ee6296b72f4ede9c1fc5f51379ec/cmd/babylond/cmd/root.go#L272), 

This migration is necessary because containers for pre-upgrade versions are pulled from Docker Hub, Without this adjustment, errors occur due to the outdated key file structure not being applied in the previous version.

That is the reason why the build tag is named `e2e_upgrade`.

app/signer/bls.go
wnjoon: @RafilxTenfen 

Applied in [99b9d3d](https://github.com/babylonlabs-io/babylon/pull/471/commits/99b9d3d73ff1ae288f449676888f3cd138e5c064)

New PR created: #471

test/e2e/initialization/export.go
wnjoon: @RafilxTenfen 

Applied in [99b9d3d](https://github.com/babylonlabs-io/babylon/pull/471/commits/99b9d3d73ff1ae288f449676888f3cd138e5c064)

New PR created: #471

x/checkpointing/types/types.go
wnjoon: @RafilxTenfen 

Applied in [99b9d3d](https://github.com/babylonlabs-io/babylon/pull/471/commits/99b9d3d73ff1ae288f449676888f3cd138e5c064)

New PR created: #471

gitferry:
@RafilxTenfen I addressed conflicts with `main` mostly from #457. Could you review the changes?

gitferry:
I could not perform all the steps from old version `babylond` transitioned to this PR due to some breaking changes in between. For example, after setting up with `v1.0.0-rc.4`, I migrated the key and got errors with `babylond start`:
```
babylond start
failed to load latest version: version of store zoneconcierge mismatch root store's version; expected 1 got 0; new stores should be added using StoreUpgrades
```
@RafilxTenfen Do you have any clues?

KonradStaniec:
>I could not perform all the steps from old version babylond transitioned to this PR due to some breaking changes in between. For example, after setting up with v1.0.0-rc.4, I migrated the key and got errors with babylond start:

I think you will able to make the compatibility test only after back porting this pr to v1. As on `main` , zoneconcierge module exists while on `v1` it does not.

cmd/babylond/cmd/genhelpers/gentx.go
```diff
@@ -176,28 +176,27 @@ $ %s gentx my-key-name 1000000stake --home=/path/to/home/dir --keyring-backend=o
...
-			wrappedPV := privval.LoadWrappedFilePV(privValKeyFilePath, statePath)
-			if wrappedPV.Key.DelegatorAddress == "" {
-				cmd.PrintErrf("failed to read bls keys in %q. You probably forgot to run `babylond create-bls-key`, creating...\n", privValKeyFilePath)
-				wrappedPV, err = bls.CreateBlsKey(clientCtx.HomeDir, addr)
-				if err != nil {
-					return sdkerrors.Wrap(err, "failed to set bls keys into priv_validator_key.json")
-				}
-				cmd.PrintErrf("successfully created bls key in %q.\n", privValKeyFilePath)
+
+			ck, err := appsigner.LoadConsensusKey(clientCtx.HomeDir)
```
RafilxTenfen, line 180: Here we should basically check if the user already have the key setup, if it has there is no need to export, if there is no bls key set, export it

RafilxTenfen (approved):
LGTM, very nice docs @wnjoon :tada:
Only one comment during gentx where it might not even be necessary if the bls key is created during `init`

KonradStaniec (approved):

cmd/babylond/cmd/genhelpers/gentx.go
RafilxTenfen: so we should probably remove it here the call `appsigner.ExportGenBls(` 
@gitferry

RafilxTenfen: After offline discussing with @gitferry I understood 

`gen-bls-bbnvaloper1w60pd30puqlvma943fe5jcxsurlyfxzv3cxpvr.json` can be created only by `babylond gentx`

`bls_key.json` can be created by `bablyond init` or `babylond create-bls-key`

no need to remove `ExportGenBls` all good to merge

PR #473, opened 2025-02-04
fix status check from confirmed checkpoint

KonradStaniec:
fixes: https://github.com/babylonlabs-io/pm/issues/180


gitferry (approved):
Nice work!

PR #491, opened 2025-02-07
allow 4 decimal places in  slashing rateparamter

KonradStaniec:
closed: https://github.com/babylonlabs-io/pm/issues/182

Changes:
- allow slashing rate to have 4 decimal places
- add parameter validation to check whether minimal slashing output is not dust under BTC standard rules


gitferry (approved):
Nice fix!

PR #493, opened 2025-02-07
add rc.5 tesnet upgrade handler

KonradStaniec:
Adds testnet v1rc5 handler

Note: handler needs to live on main as later when we will be upgrading testnet to v2, the upgrade will need to be made from the latest upgrade on the testnet


RafilxTenfen (approved):
LGTM

PR #504, opened 2025-02-11
feat: Immediate BTC headers relay to consumer chains

gusin13:
Resolves - https://github.com/babylonlabs-io/pm/issues/192
ADR - https://github.com/babylonlabs-io/pm/pull/193

Note - 
Relevant changes are needed in babylon-contract, will start work on it soon.

### DIFF
--- a/x/zoneconcierge/keeper/ibc_packet_btc_headers.go
+++ b/x/zoneconcierge/keeper/ibc_packet_btc_headers.go
@@ -17,8 +17,17 @@ func (k Keeper) BroadcastBTCHeaders(ctx context.Context) {
 		return
 	}
 
-	// TODO: currently sending last w+1 headers but should fetch from BSN base header to tip
-	// This will be fixed once Babylon knows about BSN base header
+	// Current behavior:
+	// - If no headers sent: Returns last w+1 headers from tip
+	// - If headers previously sent:
+	//   - If last segment valid: Returns headers from last sent to tip
+	//   - If last segment invalid (reorg): Returns last w+1 headers from tip
+	//
+	// TODO: Should use BSN base BTC header as starting point:
+	// - If no headers sent: Return from BSN base to tip
+	// - If headers previously sent:
+	//   - If last segment valid: Return from last sent header to tip
+	//   - If last segment invalid (reorg): Return from BSN base to tip
 	headers := k.getHeadersToBroadcast(ctx)
 	if len(headers) == 0 {
 		k.Logger(sdkCtx).Info("no new BTC headers to broadcast")


x/zoneconcierge/keeper/hooks.go
```diff
@@ -38,7 +38,12 @@ func (h Hooks) AfterRawCheckpointSealed(ctx context.Context, epoch uint64) error
 
 // AfterRawCheckpointFinalized is triggered upon an epoch has been finalised
 func (h Hooks) AfterRawCheckpointFinalized(ctx context.Context, epoch uint64) error {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	headersToBroadcast := h.k.getHeadersToBroadcast(ctx)
+	if len(headersToBroadcast) == 0 {
+		h.k.Logger(sdkCtx).Info("no new BTC headers to broadcast", "epoch", epoch)
+		return nil
+	}
```
SebastianElvis, line 46: Why do we need this check?

In addition if `len(headersToBroadcast) == 0`, then the protocol will skip broadcasting BTC timestamps and update the last sent segment, which is not expected. Shall we remove it then?

x/zoneconcierge/keeper/ibc_packet_btc_headers.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"context"
+
+	btclctypes "github.com/babylonlabs-io/babylon/x/btclightclient/types"
+	"github.com/babylonlabs-io/babylon/x/zoneconcierge/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// BroadcastBTCHeaders sends an IBC packet of BTC headers to all open IBC channels to ZoneConcierge
+func (k Keeper) BroadcastBTCHeaders(ctx context.Context) {
```
SebastianElvis, line 12: Can we reuse `getHeadersToBroadcast` in this function, or merge these two functions to some extent?

SebastianElvis (commented):
Some initial comments, overall great work :+1:

x/zoneconcierge/keeper/hooks.go
SebastianElvis: > Both broadcasting and updating last segment would get skipped - code link

I think we should not skip these two if `len(headersToBroadcast) == 0`. The protocol still needs to broadcast this new BTC timestamp. It's possible that there is no new BTC header since last BTC timestamp, but the chain receives a new BTC checkpoint that is `w` deep, e.g., vigilante reporter is offline for a long time. 

But indeed `lastSegment` is not supposed to be `nil` in this case -- the supposed behaviour is that the chain does not broadcast any BTC header, but in the current impl it will broadcast `w` [headers](https://github.com/babylonlabs-io/babylon/blob/d0133d28e1e45db7a2678c9fd3e868594d439447/x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go#L168-L173). Maybe we need to fix this part

x/zoneconcierge/keeper/ibc_packet_btc_headers.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// 1. Check channels first - fail fast
+	openZCChannels := k.GetAllOpenZCChannels(ctx)
+	if len(openZCChannels) == 0 {
+		k.Logger(sdkCtx).Info("no open IBC channel with ZoneConcierge, skip broadcasting BTC headers")
+		return
+	}
+
+	// 2. Get headers to broadcast
+	headers := k.getBTCHeadersToSend(ctx, types.FullChainFetch)
```
SebastianElvis, line 22: This will be likely to bloat the IBC packet tbh, as the IBC packet will contain the entire BTC header chain.

Instead, it shall start from the base header of the BTC light client in the contract side. For now wdyt we still use `WDeepFetch` mode?

After we have proper initialisation of BTC light client in contract, we can revisit this part

SebastianElvis: > Shouldn't the base header in Babylon and consumer contract match?

Not really, because Babylon and BSN might not launch at the same time. If the BSN launches 1 years after Babylon then it can use a base header much later than Babylon

> Can we send multiple smaller IBC packets (in the same EndBlocker) instead of a big one?

Ideally, the BSN will use a base header that is slightly earlier to its launch date (e.g., 100 headers prior to the launch). Otherwise it adds overhead to both Babylon IBC traffic and the storage overhead of BSN

SebastianElvis: Yes sounds good on reusing the old [headersToBroadcast](https://github.com/babylonlabs-io/babylon/blob/07836b55989afef5400548d410cb93c0fb813792/x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go#L165) for this PR, and fix the contract init as a subsequent task

SebastianElvis (approved):
LGTM :+1:

x/zoneconcierge/keeper/ibc_packet_btc_headers.go
```diff
@@ -0,0 +1,41 @@
...
+func (k Keeper) BroadcastBTCHeaders(ctx context.Context) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+
+	openZCChannels := k.GetAllOpenZCChannels(ctx)
+	if len(openZCChannels) == 0 {
+		k.Logger(sdkCtx).Info("no open IBC channel with ZoneConcierge, skip broadcasting BTC headers")
+		return
+	}
+
+	// TODO: currently sending last w+1 headers but should fetch from BSN base header to tip
```
KonradStaniec, line 20: Is this something planned for the future and do we have task for it ?

KonradStaniec (approved):
Code lgtm! 👍 

Just one question / comment

x/zoneconcierge/keeper/ibc_packet_btc_headers.go
SebastianElvis: Reading the code again -- does `headers` always contain w+1 headers or only new headers since the last segment? My understanding is the latter but not former

Lazar955 (approved):

PR #513, opened 2025-02-12
Fix/datagen for consumers

maurolacy:
Generalises `datagen` utils / test helpers, so that they can be used from consumer networks that have a different address prefix.

### DIFF
--- a/testutil/datagen/account_balance.go
+++ b/testutil/datagen/account_balance.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	appparams "github.com/babylonlabs-io/babylon/app/params"
 	sec256k1 "github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
-	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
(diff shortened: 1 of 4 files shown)


testutil/datagen/account_balance.go
```diff
@@ -15,6 +18,34 @@ func GenRandomAccount() *authtypes.BaseAccount {
 	return acc
 }
 
+func GenRandomAccountWithPrefix(prefix string) *authtypes.BaseAccount {
+	senderPrivKey := sec256k1.GenPrivKey()
+	acc := MustNewBaseAccountWithPrefix(senderPrivKey.PubKey().Address().Bytes(), senderPrivKey.PubKey(), 0, 0, prefix)
+	return acc
+}
```
SebastianElvis, line 25: Can we use `sdk.MustBech32ifyAddressBytes` to convert a `sdk.AccAddress` to a string with a given prefix? This way we don't need the other functions introduced in this PR

testutil/datagen/btcstaking.go
```diff
@@ -121,7 +121,7 @@ func GenRandomBTCDelegation(
 	if err != nil {
 		return nil, err
 	}
-	staker := GenRandomAccount()
+	staker := GenRandomAccountWithPrefix("bbn") // Staker address is always Babylon's
```
SebastianElvis, line 124: We can use [Bech32PrefixAccAddr](https://github.com/babylonlabs-io/babylon/blob/c26151a04894976132cf8d343225298c5248083c/app/params/config.go#L20-L21) so that we don't need to hardcode

maurolacy: We can, as long as this is not chain specific. Will try it.

testutil/datagen/account_balance.go
maurolacy: Not sure. We need to convert to a `BaseAccount`.

SebastianElvis: I see. Would it be possible to get rid of the `BaseAccount` creation since we are only using `staker.Address` in the `staker` object ([ref](https://github.com/babylonlabs-io/babylon/pull/513/files#diff-131bdafa81adc9d991427ed518305f553a1077c162d190a0d923b0497f151ac6R168))?

maurolacy: We have `Address` (string) `AccAddress` ([]byte), and `BaseAccount` (among others...)

Question would be, how to get a random `AccAddress`, i.e. the random (but valid) stream of bytes that represents a pubkey, and can be converted to an address string.

maurolacy: Point is, that this `staker.Address` is _already_ the (bech32-formatted) address string with a given prefix. I can convert it to bytes and the bech32ify again with the new prefix, but... the current solution is better / clearer, IMO.

maurolacy: Oh, I can use this: `senderPrivKey.PubKey().Address().Bytes()` I guess.

SebastianElvis: > senderPrivKey.PubKey().Address().Bytes()

yes imo generating a random pk, deriving byte address, and then bech23ifying it with the prefix is most clean

maurolacy: Done.

SebastianElvis (approved):
Great work!

Maybe one more approval from @babylonlabs-io/core-dev would be great

PR #518, opened 2025-02-12
fix: add `btc reorg > k` check in end blocker 

RafilxTenfen:
- Update `x/btclightclient` hook to also report from which block header is rollingback to
 `AfterBTCRollBack(ctx context.Context, headerInfo *types.BTCHeaderInfo)` -> `AfterBTCRollBack(ctx context.Context, rollbackFrom, rollbackTo *types.BTCHeaderInfo)`
 - Implement btc light client hook in `x/btcstaking` to listen `AfterBTCRollBack` and register the largest BTC reorg
 - Add check in `x/btcstaking` `EndBlocker` that verifies if `largest_reorg > btc_confirmation_depth`


RafilxTenfen:
Tried to add an e2e test to simulate the halt of the chain, generating a BTC fork > depth, but it got too difficult...

x/btcstaking/keeper/hooks.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (k Keeper) Hooks() Hooks { return Hooks{k} }
+
+// AfterBTCRollBack updates the Largest BTC reorg if it is higher than the current one
+// to later verify in EndBlocker.
+func (h Hooks) AfterBTCRollBack(goCtx context.Context, rollbackFrom, rollbackTo *ltypes.BTCHeaderInfo) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// sanity checks for rollback values
+	if rollbackFrom == nil {
```
KonradStaniec, line 31: hmm I feel like all those checks i.e
- `if rollbackFrom == nil`
- `if rollbackTo == nil`
- `rollbackTo.Height >= rollbackFrom.Height`

Are kind of masking potential bugs in btclightclient. 

Wdyt about:
- documenting those invariants on light client side i.e in `func (k Keeper) AfterBTCRollBack(ctx context.Context, rollbackFrom, rollbackTo *types.BTCHeaderInfo) ` 
- here either assuming those holds or having separate function `checkInvariants` which panics if those are not fulfilled.

x/btcstaking/keeper/btc_reorg.go
```diff
@@ -0,0 +1,56 @@
...
+	largestReorg := k.MustGetLargestBtcReorg(ctx)
+	if largestReorg > p.BtcConfirmationDepth {
+		panic(fmt.Sprintf("Reorg %d is larger than BTC confirmation Depth %d", largestReorg, p.BtcConfirmationDepth))
+	}
+}
+
+// SetLargestBtcReorg sets the new largest BTC block reorg if it is higher than the current
+// value in the store.
+func (k *Keeper) SetLargestBtcReorg(ctx context.Context, newLargestBlockReorg uint32) error {
+	exists, err := k.LargestBtcReorgInBlocks.Has(ctx)
```
KonradStaniec, line 23: when does `LargestBtcReorgInBlocks.Has` can return error ? only if db is corrupted ? (if so maybe panic coud be considered here)

x/btcstaking/keeper/btc_reorg.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// HaltIfBtcReorgLargerThanConfirmationDepth safety mechanism to stop the chain in case there is an BTC reorg
+// higher than the BtcConfirmationDepth. In teory this should only happen if the babylon chain goes down for
+// a period longer than (2 * BtcConfirmationDepth * 10min).
+func (k *Keeper) HaltIfBtcReorgLargerThanConfirmationDepth(ctx context.Context) {
+	p := k.btccKeeper.GetParams(ctx)
+
+	largestReorg := k.MustGetLargestBtcReorg(ctx)
+	if largestReorg > p.BtcConfirmationDepth {
```
KonradStaniec, line 15: just to double check.

We are allowing stakes which are included in blocks at confirmation depth: https://github.com/babylonlabs-io/babylon/blob/v1.0.0-rc.5/x/btcstaking/keeper/inclusion_proof.go#L60, so maybe we should panic here if
```
largestReorg >= p.BtcConfirmationDepth
```
i.e main reason for this fail safe is that we won't progress forward with stake that potentially has been reorged.

Maybe we can add some more comprehensive test in `msg_server_test.go` which add some stake on depth boundary and trigger this fail safe to see if such stake would be affected ?

x/btcstaking/keeper/btc_reorg.go
```diff
@@ -0,0 +1,56 @@
...
+		panic(fmt.Sprintf("Reorg %d is larger than BTC confirmation Depth %d", largestReorg, p.BtcConfirmationDepth))
```
KonradStaniec, line 16: hmm I think this panic should be more descriptive i.e after this happens,  I imagine recovery procedure in which we will need to invalidate some stakes that were potentially reorged. Ideally we would have here:
- old btc light client tip info (height, hash)
- block to which we reorged to (height, hash)

Wdyt ? or this is overkil and if such reorg happen we will be able to easily get this data from the node ?

KonradStaniec (commented):
In general looks good 👍 only few discussion points

x/btcstaking/keeper/btc_reorg.go
RafilxTenfen: yes, if encoding fails 
That makes sense, I will also panic if Get returns an error in the cases where we verify if exists it first

x/btcstaking/keeper/btc_reorg.go
RafilxTenfen: I can add this information, this would mean we would have a new structure instead of just a primitive `uint32`  for `LargestBtcReorgInBlocks`, 
something like 
```proto
// LargestBtcReOrg stores the largest BTC reorg recorded
message LargestBtcReOrg {
    // BlockDiff is the difference of the block height of the BTC header Tip - the btc height
    // which it was rolled back
    uint32 block_diff = 1;
    // RollbackFrom is the latest BTC block header prior to rollback
    babylon.btclightclient.v1.BTCHeaderInfo rollback_from = 2;
    // RollbackTo is the BTC block header which we rollback to
    babylon.btclightclient.v1.BTCHeaderInfo rollback_to = 3;
}
```

We would be able to collect this information from past transactions, but it might be worth to do it to gather more evidence in the panic, what do you think? @KonradStaniec

KonradStaniec: sounds good ! imo having as much data as possible to recover after such even is something super important

x/btcstaking/keeper/msg_server_test.go
```diff
@@ -403,6 +404,156 @@ func TestProperVersionInDelegation(t *testing.T) {
...
+
+	// checks the query with a reorg set
+	respLargestReOrg, err := h.BTCStakingKeeper.LargestBtcReOrg(h.Ctx, &types.QueryLargestBtcReOrgRequest{})
+	h.NoError(err)
+	require.Equal(t, respLargestReOrg.BlockDiff, currLargestReorg.BlockDiff)
+	require.Equal(t, respLargestReOrg.RollbackFrom.HashHex, rBlockFrom.ToResponse().HashHex)
+	require.Equal(t, respLargestReOrg.RollbackTo.HashHex, rBlockTo.ToResponse().HashHex)
+
+	// BTC staking tx is still seen as active rolling back to a block where the confirmation depth is less than btcctParams.BtcConfirmationDepth
+	// Should we handle this differently?
```
RafilxTenfen, line 449: @KonradStaniec 

Imagine the following:
1. btc staking tx included block 20
2. btc tip is 30
3. Msg create btc delegation is sent + cov sigs
4. BTC delegation is ACTIVE
5. rollback 9 blocks (30 -> 21) 

what should be the status of that BTC delegation? In my mind would make sense to be as pending again
But I am not sure in the complications this might have in other modules

For any reorg larger or equal than `btcctParams.BtcConfirmationDepth` case by case will need to be analyzed and verified how it should be revoked

KonradStaniec: So in BTC rollback happens only if there is better chain presented and better chain means chain with more blocks. 

Therefore it this most probably after rollbacking `9` blocks, `10` or more different blocks will be inserted. This means that in case you presented the transaction included in block `20` will still be considered active. In other words, in recovery procedure we can only care about transactions in blocks which have been rollbacked.

KonradStaniec (approved):
Looks good to me ! lets wait for second approval given how critical it is.

Also wdyt about also adding `.md` file how the recovery procedure should look like when we hit this case ?

x/btcstaking/types/btcstaking.go
```diff
@@ -107,3 +108,12 @@ func GetOrderedCovenantSignatures(fpIdx int, covSigsList []*CovenantAdaptorSigna
 
 	return orderedCovSigs, nil
 }
+
+// NewLargestBtcReOrg creates a new Largest BTC reorg based on the rollback vars
+func NewLargestBtcReOrg(rollbackFrom, rollbackTo *btclightclienttypes.BTCHeaderInfo) LargestBtcReOrg {
+	return LargestBtcReOrg{
+		BlockDiff:    rollbackFrom.Height - rollbackTo.Height,
```
gitferry, line 115: Do we need to ensure that `rollbackFrom.Height > rollbackTo.Height`?

x/btcstaking/docs/btc-reorg.md
```diff
@@ -0,0 +1,150 @@
...
+the BTC delegations that were recorded until the first reorg block height.
+
+By example, if BTC is in block height 150 and Babylon chain has a
+`BtcConfirmationDepth` of 10 and it is submitted a valid BTC reorg of 10 blocks
+that rollbacks BTC to block height 140, every BTC delegation that was included
+in BTC block 140 or higher should be deleted as the inclusion proof will
+be different. BTC delegations included before BTC block height 140 will not
+be affected as the BTC rollbacks only happens if there is a better chain
+presented, a better chain means chain with more blocks, so any BTC delegation
+included in block 139 or prior would still be considered valid and active.
```
gitferry, line 19:
```suggestion
included in block 139 or prior would still be considered valid.
```

Not necessarily active

gitferry (approved):
Great work! Some questions:

x/btcstaking/keeper/btc_reorg.go
```diff
@@ -0,0 +1,73 @@
...
+func (k *Keeper) SetLargestBtcReorg(ctx context.Context, newLargestBlockReorg types.LargestBtcReOrg) error {
+	exists, err := k.LargestBtcReorg.Has(ctx)
+	if err != nil {
+		panic(fmt.Errorf("setting largest btc reorg failed encode in Has: %w", err))
+	}
+	if !exists {
+		return k.LargestBtcReorg.Set(ctx, newLargestBlockReorg)
+	}
+
+	currentLargestReorg, err := k.LargestBtcReorg.Get(ctx)
```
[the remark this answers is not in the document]
RafilxTenfen: but the it might be an error in the get even if it is `ErrNotFound` 
would you prefer for me to do an error check == `ErrNotFound`  and avoid a call to Has?

x/btcstaking/types/btcstaking.go
RafilxTenfen: we do check this in `CheckRollBackInvariants`

x/btcstaking/keeper/btc_reorg.go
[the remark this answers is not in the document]
RafilxTenfen: it could by example fail to decode, which then would lead to a panic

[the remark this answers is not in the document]
RafilxTenfen: I will merge this as it is, but happy to address further comments in another PR if needed

PR #519, opened 2025-02-12
chore(x/incentive): add missing data to genesis 

GAtom22:


proto/babylon/incentive/genesis.proto
```diff
@@ -3,10 +3,48 @@ package babylon.incentive;
...
+  // Type of stakeholder
+  StakeholderType stakeholder_type = 1;
+  // Address of the stakeholder
+  string address = 2 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
+  // The reward gauge object
+  RewardGauge reward_gauge = 3;
+}
+
+// StakeholderType represents the different types of stakeholders.
+enum StakeholderType {
```
RafilxTenfen, line 42: this is a dangerous part :sweat_smile:  we should definitely do it! 
Just needs to verify if a migration is needed in the upgrade handlers
https://github.com/babylonlabs-io/babylon/blob/ab5129460681b4682f44ef66bdc764ba7060f7ff/app/upgrades/README.md#L1

GAtom22: yeah, good point!
I'll test if this breaks. In such case, we'll need to add the corresponding migration.

GAtom22: tested this and works OK

RafilxTenfen: how did you did this test?

x/incentive/keeper/genesis.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/babylonlabs-io/babylon/x/incentive/types"
+	"github.com/cosmos/cosmos-sdk/runtime"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// InitGenesis performs stateful validations and initializes the keeper state from a provided initial genesis state.
+func (k Keeper) InitGenesis(ctx context.Context, gs types.GenesisState) error {
+	for _, entry := range gs.BtcStakingGauges {
+		// check that height is less than current height
+		sdkCtx := sdk.UnwrapSDKContext(ctx)
```
RafilxTenfen, line 17: question: does the context have correct block height at `InitGenesis`? Like does it loads from the genstate the last block height it stopped as well? 

If yes, nice check!

x/incentive/keeper/genesis_test.go
RafilxTenfen, line 1: :rocket:

x/incentive/types/genesis.go
```diff
@@ -1,14 +1,157 @@
...
+
+		if err := entry.Validate(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Helper function to sort gauges and addresses to get a deterministic
```
RafilxTenfen, line 140: Nice!

GAtom22:
https://github.com/Mergifyio backport release/v1.x

x/incentive/keeper/genesis.go
GAtom22: yes it does! However, with the current binary this will fail anyways due to a check in the `InitEpoch` called in the epoching `InitGenesis`

```
panic: InitEpoch can be invoked only at genesis. Invoked at height 6
```

proto/babylon/incentive/genesis.proto
GAtom22: I started a local chain with the current changes with some reward gauge seeds on the genesis. Then switched to the bin on main without these changes and queried the reward gauges. Also just added a [test here](https://github.com/babylonlabs-io/babylon/blob/d6de2fc6406a0664cbae6331a48ee79eb36cad1c/x/incentive/keeper/reward_gauge_test.go#L44) to make sure it is non-breaking. Let me know if we need a more thorough test

RafilxTenfen: awesome! that makes sense
Thanks!

RafilxTenfen (approved):
LGTM :tada:

proto/babylon/incentive/genesis.proto
KonradStaniec: I think it is not breaking as we are stil using the same method to convert enum to bytes i.e:
```
func (st StakeholderType) Bytes() []byte {
	return []byte{byte(st)}
}
```
If we would switch to proto serialization then most probably it would be breaking change.

KonradStaniec (approved):
lgtm ! great improvement

Lazar955 (approved):
Nice work!

PR #527, opened 2025-02-17
fix: Add bls signer to start command with password options

Vvaradinov:
- Adds `--bls-password` and `--bls-no-password` to start command.
- Default behaviour is to provide a password before node starts for bls signer.

### DIFF
--- a/cmd/babylond/cmd/cmd_test.go
+++ b/cmd/babylond/cmd/cmd_test.go
@@ -2,14 +2,23 @@ package cmd_test
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
 	"github.com/babylonlabs-io/babylon/app"
+	"github.com/babylonlabs-io/babylon/app/signer"
 	"github.com/babylonlabs-io/babylon/cmd/babylond/cmd"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/crypto/hd"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	svrcmd "github.com/cosmos/cosmos-sdk/server/cmd"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/x/genutil/client/cli"
+	stakingcli "github.com/cosmos/cosmos-sdk/x/staking/client/cli"
 )
 
 func TestInitCmd(t *testing.T) {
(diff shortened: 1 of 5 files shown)


RafilxTenfen (approved):
LGTM, but we should add some unit test to check the init command writes the bls

cmd/babylond/cmd/root.go
```diff
@@ -308,9 +310,24 @@ func newApp(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts serverty
...
+				password = appsigner.NewBlsPassword()
+			}
+		}
+
+		var err error
+		blsSigner, err = appsigner.CreateBlsSigner(homeDir, password)
+		if err != nil {
+			panic(fmt.Errorf("failed to create new BLS signer: %w", err))
+		}
 	}
```
gitferry, line 331: I think we probably can wrap this chunk into `LoadOrGenBlsKey` like [here](https://github.com/cosmos/cosmos-sdk/blob/4e74a30430c8c2bcb9f760d7df287b2388a48702/server/start.go#L381)

cmd/babylond/cmd/root.go
```diff
@@ -241,6 +241,8 @@ func addModuleInitFlags(startCmd *cobra.Command) {
 
 	startCmd.Flags().String(flags.FlagKeyringBackend, flags.DefaultKeyringBackend, "Select keyring's backend (os|file|kwallet|pass|test)")
 	startCmd.Flags().String(flags.FlagChainID, "", "genesis file chain-id, if left blank will be randomly created")
+	startCmd.Flags().Bool(flagNoBlsPassword, false, "Generate BLS key without password protection (suitable for RPC nodes)")
```
gitferry, line 244: Would it improve UX if we set the default value to `true`? Then RPC node runners don't need to be aware of BLS stuff

gitferry (commented):
Great work! Some minor comments. Agreed that we should have a test for this like below:
- run `babylond start` without bls key existence
- ensure `babylond start` can be started normally and bls key is created

cmd/babylond/cmd/root.go
Vvaradinov: Yeah setting to true would be better.

gitferry (approved):
Lgtm!

gitferry:
Let's do a local running to ensure the cmd works before running. You can find a good guide here https://github.com/babylonlabs-io/networks/blob/main/bbn-test-5/babylon-node/README.md

Vvaradinov:
@gitferry I added a unit test for the `start` command, not sure if the scenario makes sense but if the validator inits the chain with BLS key and then deletes it, the `start` command now re-creates it.

gitferry:
@Vvaradinov it makes sense. Imagine an RPC node runer migrate the node to a new site without the BLS key and stuff (she might not be aware of the BLS key) and run `babylond start`.

cmd/babylond/cmd/cmd_test.go
```diff
@@ -23,3 +32,122 @@ func TestInitCmd(t *testing.T) {
...
+		sdk.FullFundraiserPath,
+		keyring.DefaultBIP39Passphrase,
+		hd.Secp256k1,
+	)
+	require.NoError(t, err)
+
+	addr, err := keyInfo.GetAddress()
+	require.NoError(t, err)
+
+	prepareCmd := cmd.NewRootCmd()
```
gitferry, line 76: not sure why we need this and below? the testing scenario is for rpc node runner, right?

Vvaradinov: Yeah but we still need a valid chain genesis and a running chain to test the `start` command otherwise it fails if the initial setup is not correct.

PR #528, opened 2025-02-17
docs: Staking Registration Eligibility

samricotta:
Document explaining Staking Registration Eligibility inclusive of timeline


vitsalis (approved):
LGTM. We should also backport and please double check that the image corresponds to the text.

PR #530, opened 2025-02-17
chore(checkpointing): add ConflictingCheckpointReceived flag

GAtom22:
Closes: https://github.com/babylonlabs-io/pm/issues/196


proto/babylon/checkpointing/v1/genesis.proto
```diff
@@ -10,6 +10,9 @@ option go_package = "github.com/babylonlabs-io/babylon/x/checkpointing/types";
 message GenesisState {
   // genesis_keys defines the public keys for the genesis validators
   repeated GenesisKey genesis_keys = 1;
+  // conflicting_checkpoint_received is a flag used to halt the chain
+  // when receiving a conflicting checkpoint
+  bool conflicting_checkpoint_received = 2;
```
KonradStaniec, line 15: I wonder whether there is  a point in exporting this in genesis i.e if you restart chain from genesis sate in which this is `true` it will immediately halt after block `1` . Wdyt ?

KonradStaniec (commented):
Just one minor question / comment

proto/babylon/checkpointing/v1/genesis.proto
GAtom22: Yes, with the current logic, will work as you mention.
OK, will remove it from the genesis state then

KonradStaniec (approved):

PR #537, opened 2025-02-19
chora(btcstaking): add validations to EditFinalityProvider

GAtom22:
Closes: https://github.com/babylonlabs-io/pm/issues/190


gitferry (commented):
While I was looking into how Cosmos-SDK did this, I found one surprise is that when create a validator, a commissionRate info is also required, which includes `CommissionMaxRate` and `CommissionMaxChange` other than in the params. Do we want to keep it consistent with Cosmos-SDK? I think the benefit is that we can avoid adding a new param to the staking module but we need to change the message structure of creating a finality provider (i.e., adding `CommissionRate`). Wdyt? cc @KonradStaniec

GAtom22:
thanks for the review @gitferry! yes, I find it better to include the [`Commission` ](https://github.com/cosmos/cosmos-sdk/blob/a01cb3ba66ac66efa9d30534cbba5e17d55c3182/x/staking/types/staking.pb.go#L214)and [`CommissionRates`](https://github.com/cosmos/cosmos-sdk/blob/a01cb3ba66ac66efa9d30534cbba5e17d55c3182/x/staking/types/staking.pb.go#L171) types. It is cleaner cause the `Commission` type includes there the update time. But it is true that is a more disruptive change. Let's see what the team thinks about it

GAtom22:
> While I was looking into how Cosmos-SDK did this, I found one surprise is that when create a validator, a commissionRate info is also required, which includes `CommissionMaxRate` and `CommissionMaxChange` other than in the params. Do we want to keep it consistent with Cosmos-SDK? I think the benefit is that we can avoid adding a new param to the staking module but we need to change the message structure of creating a finality provider (i.e., adding `CommissionRate`). Wdyt?
> 
> Seems this is a big decision @babylonlabs-io/core-dev bump vis

Updated the PR with these changes on the `MsgCreateFinalityProvider`

gitferry (commented):
Overall it looks good! One major comment is that we probably should avoid using Cosmos SDK's type for commission but we should use our own type and duplicate the logics to follow Cosmos' convension.

Also request @KonradStaniec to have a look as well given that this PR involves an upgrade

proto/babylon/btcstaking/v1/tx.proto
```diff
@@ -56,6 +54,8 @@ message MsgCreateFinalityProvider {
   // consumer_id is the ID of the consumer
   // If it's empty, it's assumed to be Babylon's chain id
   string consumer_id = 6;
+  // commission is the finality provider commission information
+  cosmos.staking.v1beta1.CommissionRates commission = 7 [(gogoproto.nullable) = false, (amino.dont_omitempty) = true];
```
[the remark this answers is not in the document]
GAtom22: done!

x/btcstaking/types/msg.go
```diff
@@ -50,8 +65,13 @@ func (m *MsgCreateFinalityProvider) ValidateBasic() error {
...
 func (m *MsgEditFinalityProvider) ValidateBasic() error {
-	if m.Commission == nil {
-		return fmt.Errorf("empty commission")
+	if m.Commission != nil {
+		if m.Commission.IsNegative() {
+			return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "commission rate must be between 0 and 1 (inclusive). Got negative value")
+		}
+		if m.Commission.GT(math.LegacyOneDec()) {
+			return ErrCommissionGTMaxRate
+		}
```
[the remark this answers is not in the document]
GAtom22: In the case of the `MsgEditFinalityProvider` the `Commission` is a `math.LegacyDec`, similarly to the `MsgEditValidator` in the staking module of [the sdk](https://github.com/cosmos/cosmos-sdk/blob/e77156858f14e2725144cd81c5e39c7872db6d3c/proto/cosmos/staking/v1beta1/tx.proto#L93)

x/btcstaking/types/msg.go
```diff
@@ -22,8 +25,20 @@ var (
...
+		return fmt.Errorf("empty commission rate")
+	}
+
+	if m.Commission.MaxRate.IsNil() {
+		return fmt.Errorf("empty commission max rate")
+	}
+
+	if m.Commission.MaxChangeRate.IsNil() {
+		return fmt.Errorf("empty commission max change rate")
+	}
```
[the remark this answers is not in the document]
GAtom22: done

x/btcstaking/types/msg.go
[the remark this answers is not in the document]
GAtom22: we could rename this to `CommissionRate` to avoid the confussion with the other method

app/upgrades/v1rc7/testnet/upgrade.go
```diff
@@ -0,0 +1,101 @@
...
+			ctx := sdk.UnwrapSDKContext(context)
+
+			logger := ctx.Logger().With("upgrade", UpgradeName)
+			logger.Info("migrating finality providers...")
+			if err := MigrateFinalityProviders(ctx, keepers.BTCStakingKeeper); err != nil {
+				return nil, fmt.Errorf("failed migrate finality providers: %w", err)
+			}
+			logger.Info("finality providers migration done!")
+
+			migrations, err := mm.RunMigrations(ctx, cfg, fromVM)
```
KonradStaniec, line 43: I think it is customary to add ` mm.RunMigrations(ctx, cfg, fromVM)` as first thing in the handler. It does not make a difference it this particular case (we knot there is not migartion here), but eitherway lets add it as first thing.

app/upgrades/v1rc7/testnet/upgrade.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/babylonlabs-io/babylon/app/upgrades"
+	btcstakingkeeper "github.com/babylonlabs-io/babylon/x/btcstaking/keeper"
+	btcstakingtypes "github.com/babylonlabs-io/babylon/x/btcstaking/types"
+)
+
+const (
+	UpgradeName = "v1rc7"
+)
+
+func CreateUpgrade() upgrades.Upgrade {
```
KonradStaniec, line 23: I think you still need to add the call to `v1rc7.CreateUpgrade()` in the file `inlucde_upgrade_testnet.go` to make it available in binary.

app/upgrades/v1rc7/testnet/upgrade.go
```diff
@@ -0,0 +1,101 @@
...
+				Addr:                 fp.Addr,
+				Description:          fp.Description,
+				Commission:           fp.Commission,
+				BtcPk:                fp.BtcPk,
+				Pop:                  fp.Pop,
+				SlashedBabylonHeight: fp.SlashedBabylonHeight,
+				SlashedBtcHeight:     fp.SlashedBtcHeight,
+				Jailed:               fp.Jailed,
+				HighestVotedHeight:   fp.HighestVotedHeight,
+				ConsumerId:           ctx.ChainID(),
```
KonradStaniec, line 83: hmm this field will no be available in `release/v1.x` branch (as this is future feature) but this will need to be fixed during backport.

Here this field should be taken from `fp.ConsumerId` though it seems query is missing it 🤔 

Maybe, this upgrade won't affect the phase-3 in this shape so maybe for now lets:
- leave comments about this here
- fix this situation during backport, we will need to do it either way, as on `release/v1.x` branch this field just not exist

GAtom22: @KonradStaniec got it. should we include the `ConsumerId` on the query response for the code on `main`? I can add that change if we want to include this field

app/upgrades/v1rc7/testnet/upgrade.go
GAtom22: yes, thanks for reminding me this

app/include_upgrade_testnet.go
```diff
@@ -22,6 +22,6 @@ func init() {
 			TokensDistributionStr:     testnet.TokensDistributionStr,
 			AllowedStakingTxHashesStr: testnet.AllowedStakingTxHashesStr,
 		}, testnet.TestnetParamUpgrade),
-		v1rc5.CreateUpgrade(),
+		v1rc7.CreateUpgrade(),
```
KonradStaniec, line 25: I think we can just add `v1rc7.CreateUpgrade()` to this list instead of replace

KonradStaniec (approved):
one last nit :) and lgtm

app/upgrades/v1rc7/testnet/upgrade.go
KonradStaniec: hmm we can do this as separate pr to `main` 👍

PR #542, opened 2025-02-19
chore(x/incentive): add missing data to genesis

GAtom22:
This is a follow up of https://github.com/babylonlabs-io/babylon/pull/519 to add some missing fields in the genesis state


x/incentive/keeper/genesis.go
```diff
@@ -49,6 +51,80 @@ func (k Keeper) InitGenesis(ctx context.Context, gs types.GenesisState) error {
...
+		acc := k.accountKeeper.GetAccount(ctx, fpAddr)
+		if acc == nil {
+			return fmt.Errorf("finality provider account with address %s does not exist", entry.FinalityProviderAddress)
+		}
+		delAddr := sdk.MustAccAddressFromBech32(entry.DelegatorAddress)
+		acc = k.accountKeeper.GetAccount(ctx, delAddr)
+		if acc == nil {
+			return fmt.Errorf("delegator account with address %s does not exist", entry.DelegatorAddress)
+		}
+		if err := k.BTCDelegationRewardsTracker.Set(
```
RafilxTenfen, line 105: can we use the function `setBTCDelegationRewardsTracker`

x/incentive/keeper/genesis.go
```diff
@@ -49,6 +51,80 @@ func (k Keeper) InitGenesis(ctx context.Context, gs types.GenesisState) error {
...
+			*entry.Rewards,
+		); err != nil {
+			return err
+		}
+	}
+
+	for _, entry := range gs.BtcDelegationRewardsTrackers {
+		// check that fp and delegator accounts exists
+		fpAddr := sdk.MustAccAddressFromBech32(entry.FinalityProviderAddress)
+		acc := k.accountKeeper.GetAccount(ctx, fpAddr)
```
RafilxTenfen, line 96: nice check :tada:

x/incentive/keeper/genesis.go
```diff
@@ -49,6 +51,80 @@ func (k Keeper) InitGenesis(ctx context.Context, gs types.GenesisState) error {
...
+	}
+
+	for _, entry := range gs.FinalityProvidersHistoricalRewards {
+		// check that fp address exists
+		fpAddr := sdk.MustAccAddressFromBech32(entry.Address)
+		acc := k.accountKeeper.GetAccount(ctx, fpAddr)
+		if acc == nil {
+			return fmt.Errorf("finality provider account with address %s does not exist", entry.Address)
+		}
+		if err := k.FinalityProviderHistoricalRewards.Set(
```
RafilxTenfen, line 84: same setFinalityProviderHistoricalRewards

x/incentive/keeper/genesis.go
```diff
@@ -49,6 +51,80 @@ func (k Keeper) InitGenesis(ctx context.Context, gs types.GenesisState) error {
...
+	}
+
+	for _, entry := range gs.FinalityProvidersCurrentRewards {
+		// check that fp address exists
+		fpAddr := sdk.MustAccAddressFromBech32(entry.Address)
+		acc := k.accountKeeper.GetAccount(ctx, fpAddr)
+		if acc == nil {
+			return fmt.Errorf("finality provider account with address %s does not exist", entry.Address)
+		}
+		if err := k.FinalityProviderCurrentRewards.Set(ctx, fpAddr.Bytes(), *entry.Rewards); err != nil {
```
RafilxTenfen, line 72: lets avoid going to the collections property `k.FinalityProviderCurrentRewards` and centralize the access to the collections by using `setFinalityProviderCurrentRewards`

In fact, these collections almost should be private, what do you think?

x/incentive/keeper/genesis.go
```diff
@@ -164,3 +271,170 @@ func (k Keeper) withdrawAddresses(ctx context.Context) ([]types.WithdrawAddressE
...
+		}
+		entries = append(entries, entry)
+	}
+
+	return entries, nil
+}
+
+// btcDelegatorsToFps loads all BTC delegators to finality providers relationships stored.
+// This function has high resource consumption and should be only used on export genesis.
+func (k Keeper) btcDelegatorsToFps(ctx context.Context) ([]types.BTCDelegatorToFpEntry, error) {
```
RafilxTenfen, line 409: this part was tricky, very good job figuring it out by yourself :tada:

RafilxTenfen (commented):
a few comments/suggestion but great work

Sorry for the delay on the review

x/incentive/keeper/genesis.go
GAtom22: We can use that function. Initially I chose this approach because it separates the storing of the `BTCDelegationRewardsTracker` and the `BTCDelegatorToFP` entries. But considering we're checking the addresses maps in the `Validate()` we can use that function

```
// setBTCDelegationRewardsTracker sets a new structure in the store, it fails and returns an error if the rwd fails to marshal.
func (k Keeper) setBTCDelegationRewardsTracker(ctx context.Context, fp, del sdk.AccAddress, rwd types.BTCDelegationRewardsTracker) error {
	k.setBTCDelegatorToFP(ctx, del, fp)
	return k.BTCDelegationRewardsTracker.Set(ctx, collections.Join(fp.Bytes(), del.Bytes()), rwd)
}
```

RafilxTenfen (approved):
LGTM :tada:

x/incentive/keeper/genesis.go
GAtom22: done

PR #543, opened 2025-02-20
fix: set last segment only after successful broadcasts to all consumers

gusin13:
Currently the last segment is set even if any failures in IBC packet creation/sending. 
This leads to inconsistent state in consumers as the last segment is shared b/w BTC timestamp and BTC header IBC packets.


SebastianElvis (commented):
Still thinking hard on the error handling here. Throwing questions there to drive the offline discussion

PR #552, opened 2025-02-21
fix: reject staking messages involving slashed FPs

### LINKED ISSUE #550: btcstaking: reject staking msgs involving slashed FPs
It's possible that upon covenant signatures or staking tx inclusion proofs, the FP that the BTC delegation stakes to is slashed. In this situation Babylon should reject such messages.

SebastianElvis:
Resolves #550

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ invocation
 
 ### Bug fixes
 
+- [#552](https://github.com/babylonlabs-io/babylon/pull/552) fix: fix: reject staking messages involving slashed FPs
 - [#525](https://github.com/babylonlabs-io/babylon/pull/525) fix: add back `NewIBCHeaderDecorator` post handler
 
 ## v1.0.0-rc6
--- a/x/btcstaking/keeper/msg_server.go
+++ b/x/btcstaking/keeper/msg_server.go
@@ -225,10 +225,8 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 	// - at least 1 one of them is a Babylon finality provider,
 	// - are not slashed, and
 	// - their registered epochs are finalised
-	// and then check whether the BTC stake is restaked to FPs of consumers
-	// TODO: ensure the BTC delegation does not restake to too many finality providers
-	// (pending concrete design)
-	restakedToConsumers, err := ms.validateRestakedFPs(ctx, parsedMsg.FinalityProviderKeys.PublicKeysBbnFormat)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ invocation
 
 ### Bug fixes
 
+- [#552](https://github.com/babylonlabs-io/babylon/pull/552) fix: fix: reject staking messages involving slashed FPs
```
RafilxTenfen, line 64:
```suggestion
- [#552](https://github.com/babylonlabs-io/babylon/pull/552) fix: reject staking messages involving slashed FPs
```

x/btcstaking/keeper/msg_server.go
```diff
@@ -225,10 +225,8 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 	// - at least 1 one of them is a Babylon finality provider,
 	// - are not slashed, and
 	// - their registered epochs are finalised
-	// and then check whether the BTC stake is restaked to FPs of consumers
-	// TODO: ensure the BTC delegation does not restake to too many finality providers
-	// (pending concrete design)
```
maurolacy, line 230: It would be good to track this in an issue (if it's not already).

maurolacy (approved):
LGTM.

SebastianElvis:
closed as per offline discussion

> I think this may have a big impact on how covenant committe program will work i.e currently covenant members just see pending delegations and send signatures. The loop is basically:
> get all pending delegations that require my signatures
> send signatures
> This change modifies this behaviour in subtle way. Now in case of slashing, there will be delegations which will be forever pending. Imo this is troublesome as pending status represent delegations that are pending covenant signatures and now there is subset of delegations which are pending and Babylon queries will return them as pending, but Babylon actively rejects signatures.
> Imo other way to fix this would be to either:
> accept signatures / proofs, but do not generate voting power event in case of activation of delegation which is delegated to slashed fp
> introduce new state slashed which will be switched in all delegation when fp is slashed, and based decisions on this status

> in fact, rejecting AddBTCDelegationInclusionProof leads to some similar issue for vigilante -- it will keep retrying [here](https://github.com/babylonlabs-io/vigilante/blob/77236dac05673da2d3db9f42f9b78e8efae847e7/btcstaking-tracker/stakingeventwatcher/stakingeventwatcher.go#L762-L766) for each verified-but-slashed BTC delegation, and everytime when it polls verified BTC delegations those verified-but-slashed BTC delegations will pop up

PR #554, opened 2025-02-21
chore: Improve vote extension logs

### LINKED ISSUE #553: Missing validator address in vote extension logs
When invalid vote extension or invalid proposal is found, the validator address should be in the log for better diagnosis

gitferry:
Closes #553

### DIFF
--- a/x/checkpointing/vote_ext.go
+++ b/x/checkpointing/vote_ext.go
@@ -71,8 +71,8 @@ func (h *VoteExtensionHandler) ExtendVote() sdk.ExtendVoteHandler {
 		blsSig, err := k.SignBLS(epoch.EpochNumber, req.Hash)
 		if err != nil {
 			// NOTE: this indicates misconfiguration of the BLS key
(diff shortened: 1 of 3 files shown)


x/checkpointing/vote_ext.go
```diff
@@ -116,57 +116,61 @@ func (h *VoteExtensionHandler) VerifyVoteExtension() sdk.VerifyVoteExtensionHand
 			return resAccept, nil
 		}
 
+		extensionSigner := sdk.ValAddress(req.ValidatorAddress).String()
```
RafilxTenfen, line 119: why not continue as an error if it fails during `VerifyVoteExtension`?

RafilxTenfen (approved):

x/checkpointing/vote_ext.go
gitferry: My understanding is that Error log means something wrong in your node. But in this case, it just means that some votes are invalid, not something wrong in your node. Imagine that there are bad validators continuously sending invalid votes, people will see a large amount of error logs, which is scary

PR #556, opened 2025-02-21
chore(zoneconcierge): panic on consensus critical errors

gusin13:
Certain errors in broadcasting/generating IBC packet should be consensus critical and we should panic.

### DIFF
--- a/x/zoneconcierge/keeper/ibc_packet.go
+++ b/x/zoneconcierge/keeper/ibc_packet.go
@@ -48,14 +48,12 @@ func (k Keeper) SendIBCPacket(ctx context.Context, channel channeltypes.Identifi
 		timeoutTime, // if the packet is not relayed after this time, then the packet will be time out
 		k.cdc.MustMarshal(packetData),
 	)
-
-	// send packet
 	if err != nil {
-		// Failed/timeout packet should not make the system crash
 		k.Logger(sdkCtx).Error(fmt.Sprintf("failed to send IBC packet (sequence number: %d) to channel %v port %s: %v", seq, destinationChannel, destinationPort, err))
-	} else {
-		k.Logger(sdkCtx).Info(fmt.Sprintf("successfully sent IBC packet (sequence number: %d) to channel %v port %s", seq, destinationChannel, destinationPort))
+		return err
 	}
+
+	k.Logger(sdkCtx).Info(fmt.Sprintf("successfully sent IBC packet (sequence number: %d) to channel %v port %s", seq, destinationChannel, destinationPort))
 
 	// metrics stuff
 	labels := []metrics.Label{


x/zoneconcierge/abci.go
```diff
@@ -20,7 +20,13 @@ func BeginBlocker(ctx context.Context, k keeper.Keeper) error {
...
 
-	k.BroadcastBTCHeaders(ctx)
-	k.BroadcastBTCStakingConsumerEvents(ctx)
+	if err := k.BroadcastBTCHeaders(ctx); err != nil {
+		panic(err)
+	}
+
+	if err := k.BroadcastBTCStakingConsumerEvents(ctx); err != nil {
+		panic(err)
+	}
```
SebastianElvis, line 29: Probably have some comments on why panic is needed here (e.g., err here means consensus is breaking)?

x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
```diff
@@ -137,12 +117,20 @@ func (k Keeper) createBTCTimestamp(
...
+	if err == nil {
+		// if there is a CZ header checkpointed in this finalised epoch,
+		// add this CZ header and corresponding proofs to the BTC timestamp
+		epochOfHeader := epochChainInfo.ChainInfo.LatestHeader.BabylonEpoch
+		if epochOfHeader == epochNum {
+			btcTimestamp.Header = epochChainInfo.ChainInfo.LatestHeader
+			btcTimestamp.Proof.ProofCzHeaderInEpoch = epochChainInfo.ProofHeaderInEpoch
+		}
+	} else {
+		k.Logger(sdkCtx).Warn("failed to get epochChainInfo", "consumerID", consumerID, "epochNum", epochNum, "error", err)
```
SebastianElvis, line 133:
```suggestion
		k.Logger(sdkCtx).Debug("no epochChainInfo for this consumer at this epoch", "consumerID", consumerID, "epochNum", epochNum, "error", err)
```

SebastianElvis (approved):
Great work!

x/zoneconcierge/abci.go
```diff
@@ -20,7 +20,13 @@ func BeginBlocker(ctx context.Context, k keeper.Keeper) error {
...
 
-	k.BroadcastBTCHeaders(ctx)
-	k.BroadcastBTCStakingConsumerEvents(ctx)
+	// error in generating IBC packet data or sending packets is consensus-critical
+	if err := k.BroadcastBTCHeaders(ctx); err != nil {
+		panic(err)
+	}
+	if err := k.BroadcastBTCStakingConsumerEvents(ctx); err != nil {
+		panic(err)
+	}
```
maurolacy, line 29: I think this is better than ignoring errors, but wouldn't panicking in the end blocker cause a chain stop? I mean, imagine if one of the BSNs is down for whatever reason, that will cause the Babylon chain to stop.

Not sure this is the best option. Perhaps logging the error here and continuing makes more sense, even if it'll lead to an inconsistent state between Babylon and the BSN.

x/zoneconcierge/keeper/hooks.go
```diff
@@ -41,7 +41,9 @@ func (h Hooks) AfterRawCheckpointFinalized(ctx context.Context, epoch uint64) er
 	headersToBroadcast := h.k.getHeadersToBroadcast(ctx)
 
 	// send BTC timestamp to all open channels with ZoneConcierge
-	h.k.BroadcastBTCTimestamps(ctx, epoch, headersToBroadcast)
+	if err := h.k.BroadcastBTCTimestamps(ctx, epoch, headersToBroadcast); err != nil {
+		panic(err)
```
maurolacy, line 45: Same here, what are the consequences of a panic at this level?

x/zoneconcierge/abci.go
maurolacy: Thanks for clarifying!

maurolacy (approved):
LGTM.

PR #558, opened 2025-02-22
chore: fix bls pubkey format in bls_key.json

wnjoon:
This PR has a simple but necessary change to prevent user confusion when migrating the BLS key from an older version of `priv_validator_key.json`, which contained both Comet and BLS keys.

Before this PR, the BLS public key was stored in `bls_key.json` in hexadecimal format. While this field was never actually used—since the BLS pubkey is derived from the private key rather than being read directly from a file—users may still want to verify that their key remains unchanged after the migration.

To address this, this PR updates the BLS pubkey format in `bls_key.json` to match the base64 format used in older versions of `priv_validator_key.json`.

In the previous version of `priv_validator_key.json`:

```json
{
  ...
  "bls_pub_key": "tiy+DpOVfQeqeJZsZuQmnshaX1QESo1/GTzNR1Pafo1obwcNQTEhpTVwJi6GupFMEoh0xixGbtt3yVEff9ZXFegKoOXMjC24Le4+X98JOK0XTODaJ9zRNzjzyq/cuWPs",
  ...
}
```

The contents of `bls_key.json` after running `migrate-bls-key,` as modified in this PR:

```json
{
  ...
  "pubkey": "tiy+DpOVfQeqeJZsZuQmnshaX1QESo1/GTzNR1Pafo1obwcNQTEhpTVwJi6GupFMEoh0xixGbtt3yVEff9ZXFegKoOXMjC24Le4+X98JOK0XTODaJ9zRNzjzyq/cuWPs",
}
```


gitferry (approved):

gitferry:
@wnjoon Ah I see this field is never used. Looks good to me! Thanks for the fix. Could you resolve conflicts before I merge?

wnjoon:
@gitferry I just resolved.

PR #566, opened 2025-02-24
fix(epoching): remove float values in BeforeValidatorSlashed hook

GAtom22:
Closes: https://github.com/babylonlabs-io/pm/issues/247

### DIFF
--- a/x/epoching/keeper/hooks.go
+++ b/x/epoching/keeper/hooks.go
@@ -51,12 +51,9 @@ func (h Hooks) BeforeValidatorSlashed(ctx context.Context, valAddr sdk.ValAddres
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	// Define thresholds as integer ratios to avoid doing arithmetic
 	// operations with float values that may cause nondeterminism due to rounding
-	thresholds := []struct {
-		numerator   int64
-		denominator int64
-	}{
-		{1, 3}, // 1/3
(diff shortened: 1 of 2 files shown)


x/epoching/keeper/hooks.go
```diff
@@ -68,7 +76,16 @@ func (h Hooks) BeforeValidatorSlashed(ctx context.Context, valAddr sdk.ValAddres
 
 	for _, threshold := range thresholds {
 		// if a certain threshold voting power is slashed in a single epoch, emit event and trigger hook
-		if float64(slashedVotingPower) < float64(totalVotingPower)*threshold && float64(totalVotingPower)*threshold <= float64(slashedVotingPower+thisVotingPower) {
+		// Use integer multiplication instead of floating point comparison
+		// totalVotingPower is multiplied by the threshold (totalVotingPower * threshold) and then compared against slashedVotingPower
+		// and (slashedVotingPower + thisVotingPower). So, by multiplying totalVotingPower * threshold.numerator
+		// and slashedVotingPower & (slashedVotingPower + thisVotingPower) by the threshold.denominator
+		// we get the same comparison.
+		left := slashedVotingPower * threshold.denominator
```
KonradStaniec, line 84: given that we are dealing with decimals maybe lets use `LegacyDec` type for computations (imo 2 place precision will be enough here)

KonradStaniec (approved):

PR #572, opened 2025-02-25
backport: chore(btcstaking): update `MsgCreateFinalityProvider` and add validations to `EditFinalityProvider` #537

GAtom22:


GAtom22:
@KonradStaniec This change is API breaking. Should we create the corresponding tasks to update the processes/libraries that use these messages?

gitferry:
> This change is API breaking. Should we create the corresponding tasks to update the processes/libraries that use these messages?

Good point. I think finality provider is affected by this when creating finality provider. Could you create an issue and tackle it?

@KonradStaniec Do you think other repos need the bump?

gitferry (approved):

KonradStaniec:
good point ! we will definitely need a bump on finality provider repo. Other repos should be fine, though imo when releasing `rc.7` we should bump babylon on all repos be sure of that, it seems this release will be quite breaking.

KonradStaniec (approved):

PR #577, opened 2025-02-26
docs: BTC eligibility Document - Staking TX

samricotta:
Two questions on this:
- is it detailed enough
- Should we still include serialisation format for `OP_RETURN` data?

### DIFF
--- a/docs/btc-transactions-validity.md
+++ b/docs/btc-transactions-validity.md
@@ -14,9 +14,10 @@ There are three special transaction types
 
 ### Staking Transaction
 
-A staking transaction is different to the other transactions. It creates a UTXO
-and can only be spent by specific paths. The paths are a predefined spending
-rules enforced by a one of the spending scripts [here] (#spending-scripts).
+A staking transaction is different to the other transactions. It creates
+exactly one UTXO which commits to and can only be spent by specific paths.
+The paths are a predefined spending rules enforced by a one of the spending
+scripts [here](#spending-scripts). All other UTXOs are ignored by Babylon
 
 Requirements:
 
@@ -28,13 +29,6 @@ Requirements:
     - **Merkle Root**: Inside `Q`, commits to spending rules (timelock,
         unbonding, slashing)
     - **Internal Public Key**: Used to derive `Q`
-3. **`OP_RETURN` Output**:  Contains Babylon metadata (tag, `staker_pk`,
-    `finality_provider_pk`, `staking_time`, `version`, `global_parameters.tag`)
-
-The following are **not yet** needed.
-    - Merkle Proof
-    - Spending Script
-    - Schnorr signatures
 
 #### Data Required for Staking Output
 
@@ -70,16 +64,15 @@ generates the following:
 - An `OP_RETURN` output, storing the staking metadata
 
 ```go
-func BuildV0IdentifiableStakingOutputsAndTx(
- tag []byte,
+func BuildStakingInfo(
  stakerKey *btcec.PublicKey,
- fpKey *btcec.PublicKey,
+ fpKeys []*btcec.PublicKey,
  covenantKeys []*btcec.PublicKey,
  covenantQuorum uint32,
  stakingTime uint16,
  stakingAmount btcutil.Amount,
  net *chaincfg.Params,
-) (*IdentifiableStakingInfo, *wire.MsgTx, error)
+) (*StakingInfo, error) {
 ```
 
 To construct the full staking transaction, the user must provide the following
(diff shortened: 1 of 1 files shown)


docs/btc-transactions-validity.md
```diff
@@ -1,23 +1,179 @@
...
-## Slashing
+1. **Inputs**: References previous UTXOs, provides BTC
+2. **Taproot Output (`staking_output`)**: The vault that locks BTC. Commits
+    to a script tree composed of the three scripts (timelock, unbondng and slashing.)
+    - **Taproot Public Key**: Created in the Taproot output, commits to `m`
+        (Merkle root)
+    - **Merkle Root**: Inside `Q`, commits to spending rules (timelock,
+        unbonding, slashing)
+    - **Internal Public Key**: Used to derive `Q`
+3. **`OP_RETURN` Output**:  Contains Babylon metadata (tag, `staker_pk`,
```
KonradStaniec, line 31: I wonder wheter we need to focus on op_return at all here. It was only necessary in phase-1 and it won't used in phase-2 at all.

docs/btc-transactions-validity.md
```diff
@@ -1,23 +1,179 @@
...
-### Sending a TX
+- Staking Transaction
+- Unbonding Transaction
+- Slashing Transaction
 
-## Unbonding
+### Staking Transaction
 
-### Sending a TX
+A staking transaction is different to the other transactions. It creates a UTXO
```
KonradStaniec, line 17: I think the critical piece of infromation there here is that staking transactions must contain exactly one UTXO which commits to Babylon designed scripts. 

All other UTXOs are ignored by Babylon

docs/btc-transactions-validity.md
```diff
@@ -1,23 +1,179 @@
...
+rules enforced by a one of the spending scripts [here] (#spending-scripts).
 
-## Spending taproot outputs
+Requirements:
 
-## Slashing
+1. **Inputs**: References previous UTXOs, provides BTC
+2. **Taproot Output (`staking_output`)**: The vault that locks BTC. Commits
+    to a script tree composed of the three scripts (timelock, unbondng and slashing.)
+    - **Taproot Public Key**: Created in the Taproot output, commits to `m`
```
KonradStaniec, line 26: Imo it would be clearer if this whole info:
```
Taproot Public Key: Created in the Taproot output, commits to m (Merkle root)
Merkle Root: Inside Q, commits to spending rules (timelock, unbonding, slashing)
Internal Public Key: Used to derive Q
```
was presented in separate sub-section similiarty to what is done in current impl doc:
```
Taproot outputs are outputs whose locking script is an elliptic curve point Q created as follows:

Q = P + hash(P||m)G
where:

P is the internal public key
m is the root of a Merkle tree whose leaves consist of a version number and a script
For Bitcoin Staking transactions, the internal public key is chosen as:

P = lift_x(0x50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0)
```

docs/btc-transactions-validity.md
```diff
@@ -1,23 +1,179 @@
...
+ Tag                       []byte
+ Version                   byte
+ StakerPublicKey           *XonlyPubKey
+ FinalityProviderPublicKey *XonlyPubKey
+ StakingTime               uint16
+}
+
+```
+
+<!-- should we still include serialisation format for OP_RETURN data? -->
```
KonradStaniec, line 134: I think I would leave out the op_return at all from this docs as it is inteneded for phase-2 and onwards.

docs/btc-transactions-validity.md
```diff
@@ -1,23 +1,179 @@
...
+    - **Taproot Public Key**: Created in the Taproot output, commits to `m`
+        (Merkle root)
+    - **Merkle Root**: Inside `Q`, commits to spending rules (timelock,
+        unbonding, slashing)
+    - **Internal Public Key**: Used to derive `Q`
+3. **`OP_RETURN` Output**:  Contains Babylon metadata (tag, `staker_pk`,
+    `finality_provider_pk`, `staking_time`, `version`, `global_parameters.tag`)
 
-### Sending a TX
+The following are **not yet** needed.
```
KonradStaniec, line 34: I think I would delete this list. We do not know what inputs are used to create the staking transaction. It is possible the input to the staking transaction is previous staking output, then user would need to have those 😅

docs/btc-transactions-validity.md
```diff
@@ -1,23 +1,179 @@
...
+The Babylon staking library exposes the
+[BuildV0IdentifiableStakingOutputsAndTx](https://github.com/babylonlabs-io/babylon/blob/main/btcstaking/identifiable_staking.go?plain=1#L231)
+function, which is used to construct the staking transaction. This function
+generates the following:
+
+- A Taproot UTXO (locks BTC using `Q`)
+- An `OP_RETURN` output, storing the staking metadata
+
+```go
+func BuildV0IdentifiableStakingOutputsAndTx(
```
KonradStaniec, line 73: I would use functions without the op_return i.e
```go
func BuildStakingInfo(
	stakerKey *btcec.PublicKey,
	fpKeys []*btcec.PublicKey,
	covenantKeys []*btcec.PublicKey,
	covenantQuorum uint32,
	stakingTime uint16,
	stakingAmount btcutil.Amount,
	net *chaincfg.Params,
) (*StakingInfo, error) {
```

PR #579, opened 2025-02-26
fix: slashed fp inclusion proof

RafilxTenfen:
- Add a new check if the FP was slashed in `ProcessAllPowerDistUpdateEvents`
- Func `NewFinalityProviderDistInfo` was also modified to load the properties `IsJailed` and `IsSlashed` from the FP that was in DB (only this check was already enough to fix the issue), but to keep consistency with slashed event processing in the `ProcessAllPowerDistUpdateEvents` the slashed finality provider should not even be included in the `newDc`

This check was to avoid activating a slashed FP 

The slashed FP could be activated if a BTC delegation without inclusion proof was created before the FP was slashed and having the inclusion proof sent after the slashing


x/finality/keeper/power_dist_change_test.go
```diff
@@ -2,21 +2,268 @@ package keeper_test
...
+		events := btcStkK.GetAllPowerDistUpdateEvents(ctx, lastBTCTipHeight, tip.Height)
+		require.Equal(t, len(events), 1)
+
+		AddBtcBlock(t, r, app, ctx)
+		ctx = ProduceBlock(t, r, app, ctx)
+		vpDstCacheAfterInclusionProofOfSlashedFp := finalityK.GetVotingPowerDistCache(ctx, uint64(ctx.HeaderInfo().Height-1))
+		activeFps = vpDstCacheAfterInclusionProofOfSlashedFp.GetActiveFinalityProviderSet()
+
+		// last check to verify that the voting power distribution cache didn't changed after including proof of an BTC delegation that contains a slashed finality provider
+		require.Equal(t, len(activeFps), int(createdFps-1))
```
RafilxTenfen, line 203: important test check

RafilxTenfen:
Since this change affects the voting power distribution cache, I will wait for review&approval of all of you 
@KonradStaniec @gitferry @SebastianElvis

x/finality/keeper/power_dist_change.go
```diff
@@ -346,6 +348,13 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 			// This is a consumer FP rather than Babylon FP, skip it
 			continue
 		}
+		// if the fp is slashed it shouldn't be included in the newDc
+		if newFP.IsSlashed() {
```
RafilxTenfen, line 352: check that fixes the issue

gitferry (approved):
Nice fix!

x/finality/types/power_table.go
```diff
@@ -140,12 +139,16 @@ func (dc *VotingPowerDistCache) GetInactiveFinalityProviderSet() map[string]*Fin
...
+// NewFinalityProviderDistInfo loads the FinalityProviderDistInfo based on the fp data
 func NewFinalityProviderDistInfo(fp *bstypes.FinalityProvider) *FinalityProviderDistInfo {
 	return &FinalityProviderDistInfo{
 		BtcPk:          fp.BtcPk,
 		Addr:           sdk.MustAccAddressFromBech32(fp.Addr),
 		Commission:     fp.Commission,
 		TotalBondedSat: 0,
+		IsJailed:       fp.Jailed,
+		IsSlashed:      fp.IsSlashed(),
+		IsTimestamped:  false, // not possible to fill info without tip height
```
KonradStaniec, line 151: hmm this assumes the caller will fill this field approprietly. Maybe lets add it in docstring of this function so nobody will forget that ?

KonradStaniec (approved):

x/finality/types/power_table.go
RafilxTenfen: good idea

x/finality/keeper/power_dist_change.go
SebastianElvis: I'm not sure if this fixes the issue due to the cache. Suppose `newFP` is saved when it's not slashed yet. Then at a later point it gets slashed. However, since at the time of saving `newFP` is not slashed, `newFP.IsSlashed()` will still be false. Did I miss anything?

PR #585, opened 2025-02-27
fix: Proposal vote extensions' byte limit

gitferry:
Closes https://github.com/babylonlabs-io/pm/issues/250


RafilxTenfen (approved):
utACK

x/checkpointing/prepare/transactions.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	return PrepareProposalTxs{
+		MaxBytes:  uint64(req.MaxTxBytes),
+		UsedBytes: 0,
+	}, nil
+}
+
+// SetCheckpointTx sets the tx used for checkpoint
+func (t *PrepareProposalTxs) SetCheckpointTx(tx []byte) error {
```
KonradStaniec, line 39: maybe lets change name to `SerOrReplaceCheckpointTx` ?

x/checkpointing/prepare/transactions.go
```diff
@@ -0,0 +1,105 @@
...
+	if err := t.UpdateUsedBytes(oldBytes, newBytes); err != nil {
+		return err
+	}
+	t.CheckpointTx = tx
+	return nil
+}
+
+// AddOtherTxs adds txs to the "other" tx category and return early
+// if all the remaining bytes are used up
+func (t *PrepareProposalTxs) AddOtherTxs(allTxs [][]byte) error {
```
KonradStaniec, line 51: Maybe `ReplaceOtherTxs` ? `Add` suggests that it would extend exsiting ones but in the first line we are doing 
```go
t.OtherTxs = make([][]byte, 0, len(allTxs))
```
which mean all other txs will be cleared

x/checkpointing/prepare/transactions.go
```diff
@@ -0,0 +1,105 @@
...
+	if err := t.UpdateUsedBytes(0, bytesToAdd); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// UpdateUsedBytes updates the used bytes field. This returns an error if the num used bytes
+// exceeds the max byte limit.
+func (t *PrepareProposalTxs) UpdateUsedBytes(
```
KonradStaniec, line 73: it seems this should be private, as using this without corresponding transactions operations would leave `PrepareProposalTxs` struct in bad state

KonradStaniec (approved):
Only few minor nits, LGTM!

PR #599, opened 2025-02-28
Verify staker sig for unbonding tx

KonradStaniec:
- Adds verification of staker signature for unbonding transactions, when handling unbondings.
- To verify signature of arbitrary transaction which spend taproot output in it required to provide all funding outputs.

Note: ADR will be added after implementation to not block required changes on vigilante reporter


x/btcstaking/keeper/unbonding_transaction_sig_validation.go
```diff
@@ -0,0 +1,188 @@
...
+	if err := txscript.VerifyTaprootLeafCommitment(
+		controlBlock,
+		stakingWitnessProgram,
+		witnessScript,
+	); err != nil {
+		return fmt.Errorf("failed to verify taproot leaf commitment in witness: %w", err)
+	}
+
+	// Staker key is always first in the script, therefore signature will be last.
+	// In this true regardless of the path used to spend the staking output (timelock, unbonding, slashing)
```
gitferry, line 148:
```suggestion
	// It is true regardless of the path used to spend the staking output (timelock, unbonding, slashing)
```

x/btcstaking/keeper/msg_server.go
```diff
@@ -658,19 +711,8 @@ func (ms msgServer) BTCUndelegate(goCtx context.Context, req *types.MsgBTCUndele
...
-		if err != nil {
-			// panic as we already verified the staking tx hash in the beginning
-			panic(fmt.Errorf("failed to parse staking tx hash from existing delegation with hash %s: %w", req.StakingTxHash, err))
-		}
-
-		if !containsInput(stakeSpendingTx, stakingTxHash, btcDel.StakingOutputIdx) {
-			return nil, types.ErrInvalidBTCUndelegateReq.Wrap("stake spending tx does not spend staking output")
-		}
-
+		// spend staking tx is not the registered unbonding tx, we need to save it in the database
```
gitferry, line 714: Do we need an alert for this case?

x/btcstaking/keeper/unbonding_transaction_sig_validation.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	controlBlock, err := txscript.ParseControlBlock(
+		stakeSpendWitness[len(stakeSpendWitness)-1],
+	)
+	if err != nil {
+		return fmt.Errorf("failed to parse control block in witness: %w", err)
+	}
+
+	stakingWitnessProgram := stakingOutput.PkScript[2:]
```
gitferry, line 131: Can we explain why the staking witness is after index `1`?

gitferry (approved):
Great work! Some nits:

x/btcstaking/keeper/msg_server.go
KonradStaniec: hmm probably, we emit event about with all necessary data to identify transaction so:
1. we can have alert (maybe a warning one)
2. after that manual interverion would be needed to check saved transaction

PR #625, opened 2025-03-04
chore(ante): add tx gas limit decorator

GAtom22:


app/ante/gas_limit.go
```diff
@@ -0,0 +1,41 @@
+package ante
+
+import (
+	errorsmod "cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// defaultMaxGasWantedPerTx is the maximum gas allowed for a single transaction
+const defaultMaxGasWantedPerTx = 50_000_000
```
GAtom22, line 10: Do we want this to be configurable via the config file?
Asking because that's how Osmosis handles this. It is mostly a local mempool config in their case

app/ante/gas_limit.go
```diff
@@ -0,0 +1,41 @@
...
+
+// GasLimitDecorator will check if the gas required by a transaction
+// is less than the defined MaxGasWantedPerTx
+type GasLimitDecorator struct {
+	maxGasWantedPerTx uint64
```
GAtom22, line 15: Do we want this to be configurable via the config file?
Asking because that's how Osmosis handles this. It is mostly a local mempool config in their case

app/ante/gas_limit.go
```diff
@@ -0,0 +1,41 @@
...
+func (gld GasLimitDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+	feeTx, ok := tx.(sdk.FeeTx)
+	if !ok {
+		return ctx, errorsmod.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
+	}
+
+	// Ensure that the provided gas is less than the maximum gas per tx,
+	// if this is a CheckTx. This is only for local mempool purposes, and thus
+	// is only ran on check tx.
+	if ctx.IsCheckTx() && !simulate {
```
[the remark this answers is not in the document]
KonradStaniec: hmm what other chains with this mechanism do ? 

I imagine, if we would like to have in other modes then the maxium must be either param or hardcoded as it is consensus related, if we would like to have it only in `CheckTxMode` then it can be config option.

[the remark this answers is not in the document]
GAtom22: @KonradStaniec yes, exactly! In the case of Osmosis, is a local mempool configuration. But maybe in our case, we want to enforce this and not depend on validator configuration.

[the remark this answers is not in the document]
GAtom22: If we go with the approach to run this logic in `DeliverTxMode`, probably would be better to have the `maxGasWantedPerTx` as a param, so it can be updated if necessary via gov proposal. Otherwise, if we need to change the limit, we'll need a coordinated upgrade

app/ante/gas_limit.go
RafilxTenfen: Just read the other thread, agree with param value

app/ante/gas_limit.go
```diff
@@ -0,0 +1,71 @@
...
+		if feeTx.GetGas() > gld.maxGasWantedPerTx {
+			msg := "Too much gas wanted: %d, maximum is %d"
+			return ctx, errorsmod.Wrapf(sdkerrors.ErrOutOfGas, msg, feeTx.GetGas(), gld.maxGasWantedPerTx)
+		}
+	}
+
+	return next(ctx, tx, simulate)
+}
+
+func parseMaxGasWantedPerTx(opts servertypes.AppOptions) uint64 {
```
RafilxTenfen, line 61: will it be modified to be a parameter?

GAtom22: for now it will remain as a config param on the `app.toml`

RafilxTenfen (approved):
utACK :+1:

KonradStaniec (approved):

PR #631, opened 2025-03-05
ignore expire events when there is no covenant quorum

KonradStaniec:
- We should ignore delegation event in case of lack of covenant quorum. If there was not covenant quorum it means delegaiton was never activated and sats were never counted towards fp


gitferry (approved):
Nice fix!

PR #644, opened 2025-03-06
chore(mempool): add priority nonce mempool and tx priority logic

GAtom22:


KonradStaniec (commented):
Ideally we would have some kind of integration / higher level test (maybe in proposal_test.go ? ) which:
- puts several transactions in the mempool
- creates proposal block
- asserts that transactions in the block are order from protocol transactions to other transactions

GAtom22:
> Ideally we would have some kind of integration / higher level test (maybe in proposal_test.go ? ) which:
> 
> * puts several transactions in the mempool
> * creates proposal block
> * asserts that transactions in the block are order from protocol transactions to other transactions

Yes, working on some tests for this

app/ante/priority.go
```diff
@@ -0,0 +1,70 @@
...
+	ftypes "github.com/babylonlabs-io/babylon/x/finality/types"
+)
+
+const (
+	// RegularTxMaxPriority is the max priority for transactions with regular messages.
+	// Thus, the reserved priority range for protocol liveness-related messages is (RegularTxMaxPriority, MaxInt64]
+	RegularTxMaxPriority = math.MaxInt64 - 1000
+	// LivenessTxPriority is the priority for protocol liveness-related messages.
+	// For the moment, the priority is the same for all of these messages
+	LivenessTxPriority = RegularTxMaxPriority + 100
```
GAtom22, line 21: Do we want to add priority tiers?
'cause, for example, the `MsgCreateBTCDelegation` takes about [1.2M gas](https://testnet.babylon.explorers.guru/transaction/4A718959970785344789E03E81FAC0D178303610F9D0644EEF7FA5129101EE98?height=482328). We could set up a lower tier for the delegation creation to avoid censoring the other liveness-related txs that are cheaper (> 200k) when having many `MsgCreateBTCDelegation`

app/ante/priority.go
```diff
@@ -0,0 +1,70 @@
...
+
+func isLivenessTx(tx sdk.Tx) bool {
+	for _, msg := range tx.GetMsgs() {
+		switch msg.(type) {
+		case *btclctypes.MsgInsertHeaders, // BTC light client
+			// BTC timestamping
+			*btcctypes.MsgInsertBTCSpvProof,
+			// BTC staking
+			*bstypes.MsgCreateFinalityProvider,
+			*bstypes.MsgCreateBTCDelegation,
```
KonradStaniec, line 54: I do not think this should be prioritised  tbh

I think good rule of thumb which messages should have priority, is to prfioritize the ones which are currently being refunded (i.e there is call `IndexRefundableMsg` after message execution)

Maybe you can create small confluence page with all the messages that should be prioritized and then we can discuss it on the slack in async way ?

app/ante/priority.go
RafilxTenfen: we could have tiers for the ones that are really needed to the chain to work properly Top to Bottom priority:
- *btcctypes.MsgInsertBTCSpvProof,
- *btclighttypes.MsgInsertHeaders,
- *ftypes.MsgCommitPubRandList,
- *ftypes.MsgAddFinalitySig,

And maybe also add *ftypes.MsgSelectiveSlashingEvidence at the end 

The other types to create FPs/BTC delegations, add covenant signatures, undelegate can wait to be included

app/ante/priority.go
GAtom22: doc [here](https://babylonlabs.atlassian.net/wiki/spaces/BABYLON/pages/140149031/Babylon+Liveness-Related+Transactions). Will post it in slack as well

app/ante/priority.go
```diff
@@ -0,0 +1,64 @@
...
+			// BTC timestamping
+			*btcctypes.MsgInsertBTCSpvProof,
+			// BTC staking
+			*bstypes.MsgAddCovenantSigs,
+			*bstypes.MsgBTCUndelegate,
+			*bstypes.MsgSelectiveSlashingEvidence,
+			*bstypes.MsgAddBTCDelegationInclusionProof,
+			// BTC staking finality
+			*ftypes.MsgAddFinalitySig:
+			return true
```
KonradStaniec, line 58: We probably need to iterate over all messages and makes sure that all of them are priority, as otherwise somebody could try to prioritize some other message in the same tx

GAtom22: OK, will add this check

KonradStaniec (approved):
LGTM !

Given importance of this pr, lets wait for second review before merging.

app/params/config.go
```diff
@@ -48,6 +49,10 @@ var (
 func init() {
 	SetAddressPrefixes()
 	RegisterDenoms()
+
+	// Set DefaultMaxTx = 0 so the application uses the PriorityNonceMempool by default
+	// instead of the NoOpMempool
+	mempool.DefaultMaxTx = 0
```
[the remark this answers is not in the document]
GAtom22: it is for the `DefaultBabylonAppConfig`, but the default [on the flag ](https://github.com/cosmos/cosmos-sdk/blob/eb1a8e88a4ddf77bc2fe235fc07c57016b7386f0/server/start.go#L998-L999)will remain `-1`. Is no big deal, we can remove this

Lazar955 (approved):
looks good overall, nice work!

gitferry (approved):
Great work!

RafilxTenfen (approved):
:tada:

PR #648, opened 2025-03-06
feat: add query to iterate over all parameters in btcstaking

RafilxTenfen:


proto/babylon/btcstaking/v1/query.proto
```diff
@@ -345,3 +351,20 @@ message QueryLargestBtcReOrgResponse {
   // RollbackTo is the BTC block header which we rollback to
   babylon.btclightclient.v1.BTCHeaderInfoResponse rollback_to = 3;
 }
+
+
+// QueryParamsVersionsRequest is the request type for the
+// Query/ParamsVersions RPC method.
+message QueryParamsVersionsRequest {
+  // pagination defines an optional pagination for the request.
```
GAtom22, line 359: Nit, add annotations to avoid generating code that will not be used, e.g.:

```suggestion
  option (gogoproto.equal) = false;
  option (gogoproto.goproto_getters) = false;
  
  // pagination defines an optional pagination for the request.
```

proto/babylon/btcstaking/v1/query.proto
```diff
@@ -345,3 +351,20 @@ message QueryLargestBtcReOrgResponse {
...
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
+}
+
+// QueryParamsVersionsResponse stores all the params with versions.
+message QueryParamsVersionsResponse {
+  // params holds all the params with version from this module.
```
GAtom22, line 365: ditto here
```suggestion
  option (gogoproto.equal) = false;
  option (gogoproto.goproto_getters) = false;
  
  // params holds all the params with version from this module.
```

GAtom22 (commented):
LGTM! Great work @RafilxTenfen!!

Left a small nit.
One question, do we want to add this as a cli command also? In such case, that is missing

RafilxTenfen:
> One question, do we want to add this as a cli command also? In such case, that is missing
Good idea, added the query to CLI and addressed the pr comments 

@GAtom22 Can you kindly review again?

x/btcstaking/client/cli/query_params.go
```diff
@@ -35,3 +35,29 @@ func CmdQueryParams() *cobra.Command {
...
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientQueryContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			queryClient := types.NewQueryClient(clientCtx)
+
+			res, err := queryClient.ParamsVersions(cmd.Context(), &types.QueryParamsVersionsRequest{})
```
GAtom22, line 51: Pagination missing on request. You can get it from the flags

```suggestion
                        pageReq, err := client.ReadPageRequest(cmd.Flags())
			if err != nil {
				return err
			}
			res, err := queryClient.ParamsVersions(cmd.Context(), &types.QueryParamsVersionsRequest{
			     Pagination: pageReq,
			})
```

RafilxTenfen: good catch

GAtom22 (approved):
LGTM! :rocket:

PR #649, opened 2025-03-06
chore: backport #648

RafilxTenfen:
add query to list all the params in btcstaking


proto/babylon/btcstaking/v1/query.proto
```diff
@@ -315,3 +323,26 @@ message QueryLargestBtcReOrgResponse {
...
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
+}
+
+// QueryParamsVersionsResponse stores all the params with versions.
+message QueryParamsVersionsResponse {
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+
+  // params holds all the params with version from this module.
+  repeated StoredParams params = 1 [ (gogoproto.nullable) = false ];
```
KonradStaniec, line 344: ideally we would not expose type used for db storage, only `Params` struct as:
- every other params query contains only `Params` struct
- now every change to `StoredParams` is both api and state breaking
Though given this is backport, it is not blocking 👍

KonradStaniec (approved):

proto/babylon/btcstaking/v1/query.proto
RafilxTenfen: damm, you are correct 
I will make a PR to fix in main and backport to release

RafilxTenfen: fix at #651

PR #651, opened 2025-03-07
fix: proto query response versioned params

RafilxTenfen:
in https://github.com/babylonlabs-io/babylon/pull/648 was being returned the direct db structure instead of parsing it to a response


RafilxTenfen:
too long :/ and at this point would be API breaking change

PR #655, opened 2025-03-09
chore: add `ParseV0StakingTxWithoutTag`

RafilxTenfen:
Helps btc-staker to get parse the tx without verifying the op return tag


Lazar955 (approved):
utACK

PR #658, opened 2025-03-10
crypto: check if Z==1 in ToBTCPK

SebastianElvis:
Resolves https://github.com/babylonlabs-io/pm/issues/282

### DIFF
--- a/crypto/schnorr-adaptor-signature/keys_test.go
+++ b/crypto/schnorr-adaptor-signature/keys_test.go
@@ -25,7 +25,8 @@ func FuzzKeyGen(f *testing.F) {
 
 		// ensure that the corresponding btcPK and btcSK
 		// constitute a key pair
-		btcPK := encKey.ToBTCPK()
+		btcPK, err := encKey.ToBTCPK()
+		require.NoError(t, err)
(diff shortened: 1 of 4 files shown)


crypto/schnorr-adaptor-signature/keys_test.go
```diff
@@ -25,7 +25,8 @@ func FuzzKeyGen(f *testing.F) {
 
 		// ensure that the corresponding btcPK and btcSK
 		// constitute a key pair
-		btcPK := encKey.ToBTCPK()
+		btcPK, err := encKey.ToBTCPK()
+		require.NoError(t, err)
```
gitferry, line 29: maybe add a case where the `encKey` is one and assert error?

gitferry (approved):

Lazar955 (approved):

KonradStaniec (approved):

crypto/schnorr-adaptor-signature/keys_test.go
SebastianElvis: Probably not necessary, as per offline discussion constructor functions of `encKey` enforce Z==1, and this check is only for precaution in case future modifications of the code break this assumption

PR #660, opened 2025-03-10
add function to recover pub key from sig

KonradStaniec:
fixes:
- https://github.com/babylonlabs-io/pm/issues/298
- https://github.com/babylonlabs-io/pm/issues/297


Lazar955 (approved):
lgtm!

PR #671, opened 2025-03-12
crypto: align adaptor sig impl with Blockstream spec

SebastianElvis:
Resolves https://github.com/babylonlabs-io/pm/issues/279
Resolves https://github.com/babylonlabs-io/pm/issues/281

This PR implements the adaptor signature following the Blockstream spec @kwantam. In particular, this PR

- aligns all structs/functions to the spec
- tests the impl against the test vectors in the spec


crypto/schnorr-adaptor-signature/types.go
```diff
@@ -0,0 +1,173 @@
...
+
+const (
+	ModNScalarSize       = 32
+	FieldValSize         = 32
+	JacobianPointSize    = 33
+	AdaptorSignatureSize = JacobianPointSize + ModNScalarSize
+)
+
+type AdaptorSignature struct {
+	R0 btcec.JacobianPoint
```
KonradStaniec, line 19: I wonder wheter it would better to represent this value as `btcec.PublicKey` which already have a lot of helpers like:
- `SerializeCompressed`
- or `btcec.ParsePubKey`

crypto/schnorr-adaptor-signature/types.go
```diff
@@ -0,0 +1,173 @@
...
+	var yVal btcec.FieldVal
+	isOdd := asigBytes[0] == 0x03
+	if success := btcec.DecompressY(&xVal, isOdd, &yVal); !success {
+		return nil, fmt.Errorf("failed to decompress R0.Y")
+	}
+
+	// Create the Jacobian point
+	var zVal btcec.FieldVal
+	zVal.SetInt(1)
+	r0 := btcec.MakeJacobianPoint(&xVal, &yVal, &zVal)
```
KonradStaniec, line 67: shouldn't we check that resulting point is on `secp256k1` curve ? 

I think the easiest way to the parsing here would be to use `btcec.ParseJacobian` that does all those checks

crypto/schnorr-adaptor-signature/sign_utils.go
```diff
@@ -9,168 +9,248 @@ import (
...
-	// get R = R'+T
-	var R btcec.JacobianPoint
-	btcec.AddNonConst(RHat, t, &R)
-	// negate k and R if R.y is odd
-	affineRWithEvenY, needNegation := intoPointWithEvenY(&R)
-	R = *affineRWithEvenY
-	if needNegation {
+	var R0 btcec.JacobianPoint
+	btcec.AddNonConst(R, &tAffine, &R0)
+	R0.ToAffine()
```
KonradStaniec, line 39: shouldn't we check R0 is point at infinity - https://github.com/ZhePang/Python_Specification_for_Schnorr_Adaptor/blob/main/reference.py#L182

crypto/schnorr-adaptor-signature/sign_utils.go
```diff
@@ -9,168 +9,248 @@ import (
...
-	RHat.ToAffine()
-
-	// P = lift_x(int(pk))
+	// Step 1: Let P = lift_x(pubkey)
 	pubKey, err := schnorr.ParsePubKey(pubKeyBytes)
 	if err != nil {
 		return err
 	}
-	// Fail if P is not a point on the curve
 	if !pubKey.IsOnCurve() {
```
KonradStaniec, line 99: `schnorr.ParsePubKey(pubKeyBytes)` does `IsOnCurve` check

crypto/schnorr-adaptor-signature/sign_utils.go
```diff
@@ -9,168 +9,248 @@ import (
...
+	if (T.X.IsZero() && T.Y.IsZero()) || T.Z.IsZero() {
+		return fmt.Errorf("T point is at infinity")
 	}
-	// ensure R' is same as the expected R' = s'*G - e*P by checking both x and y coordinates
-	if !expRHat.X.Equals(&RHat.X) || !expRHat.Y.Equals(&RHat.Y) {
-		return fmt.Errorf("expected R' = s'*G - e*P is different from the actual R'")
+	T.ToAffine()
+
+	// Step 7: Verify T matches encryption key
+	T.ToAffine()
```
KonradStaniec, line 154: duplicated `T.ToAffine()` conversion

crypto/schnorr-adaptor-signature/sign_utils.go
```diff
@@ -9,168 +9,248 @@ import (
...
+	}
+
+	// Check if s0 or s is zero or exceeds curve order
+	if s0.IsZero() || s.IsZero() {
+		return nil, fmt.Errorf("s values must be non-zero")
+	}
+
+	// Get R0 point from adaptor signature to check its Y parity
+	R0 := psig.R0
+	R0.ToAffine()
```
KonradStaniec, line 224: we have this `R0.ToAffine()` methods scattered around... 

Can we define an invariant for `AdaptorSignature` struct that it always contains point in affine cooridnates and check it during struct construction ?

crypto/schnorr-adaptor-signature/sign_utils.go
```diff
@@ -9,168 +9,248 @@ import (
...
-// point with even y value, and returns a bool value on whether the
-// negation is performed.
-// The bool value will be used for decrypting an adaptor signature
-// to a Schnorr signature.
-func intoPointWithEvenY(point *btcec.JacobianPoint) (*btcec.JacobianPoint, bool) {
-	affinePoint := point
-	affinePoint.ToAffine()
+// decrypt decrypts a pre-signature using a decryption key.
+// This implements the Decrypt algorithm as defined in the spec.
+func decrypt(psig *AdaptorSignature, dk *btcec.ModNScalar) (*schnorr.Signature, error) {
```
KonradStaniec, line 164: is this function in original python spec https://github.com/ZhePang/Python_Specification_for_Schnorr_Adaptor/blob/main/reference.py ? ( I do not see it there, maybe I am missing something)

KonradStaniec (commented):
some initial comments

crypto/schnorr-adaptor-signature/types.go
SebastianElvis: In theory yes, but this might conflate between keys and underlying points representation, and deviate a bit from the spec. So I recommend we use point here and keep it aligned with spec. Wdyt?

crypto/schnorr-adaptor-signature/types.go
SebastianElvis: Nice, indeed we can just do `ParseJacobian` here

crypto/schnorr-adaptor-signature/sign_utils.go
SebastianElvis: Hmm actually this invariant is already the case https://github.com/babylonlabs-io/babylon/blob/51d86025be02ac7d4821f2abba4e0e84bb5abddc/crypto/schnorr-adaptor-signature/sign_utils.go#L44-L45. We can remove L224, and add some docstring on this invariant

crypto/schnorr-adaptor-signature/sign_utils.go
SebastianElvis: This is `schnorr_adapt` function. Will add docstring

crypto/schnorr-adaptor-signature/types.go
KonradStaniec: hmm makes sense 👍

KonradStaniec (approved):
lgtm ! though lets wait for second pair of eyes before merging.

gitferry:
sorry for being late here. I will take a look tmr as seems a lot to consume

crypto/schnorr-adaptor-signature/types.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+const (
+	ModNScalarSize       = 32
+	FieldValSize         = 32
+	JacobianPointSize    = 33
+	AdaptorSignatureSize = JacobianPointSize + ModNScalarSize
+)
+
+type AdaptorSignature struct {
```
gitferry, line 18: so the difference of adaptor signature format in block stream spec is missing one `boolean` to indicate whether it needs negation? So the conclusion is that we handle this in vigilante slasher without breaking change to Babylon?

gitferry (approved):
Great work!

crypto/schnorr-adaptor-signature/types.go
SebastianElvis: > so the difference of adaptor signature format in block stream spec is missing one boolean to indicate whether it needs negation? 

Yes

> So the conclusion is that we handle this in vigilante slasher without breaking change to Babylon?

That's one approach. The other approach (as recommened by Cubist) is to generalise the `NewAdaptorSignatureFromBytes` function such that it handles two formats of the adaptor signatures. I'm working on this

crypto/schnorr-adaptor-signature/sig.go
```diff
@@ -1,271 +1,132 @@
...
-		nonce := btcec.NonceRFC6979(
-			privKeyBytes[:], hashForNonce, customBabylonRFC6979ExtraDataV0[:], nil, iteration,
-		)
+		// Step 5-6: Generate random bytes for nonce generation
+		// genRandForNonce does the following:
+		// - Generates t as byte-wise XOR of bytes(d) and tagged_hash("SchnorrAdaptor/aux", a)
+		// - Generates rand = tagged_hash("SchnorrAdaptor/nonce", t || T || P || m)
+		var skBytes [chainhash.HashSize]byte
+		skScalar.PutBytes(&skBytes)
+		randForNonce := genRandForNonce(skBytes, auxData, encKeyBytes, pkBytes, msgHash)
```
kwantam, line 112: zeroize skBytes after use

crypto/schnorr-adaptor-signature/sig.go
```diff
@@ -1,271 +1,132 @@
...
+
+		// Step 7: Generate nonce `k' = int(rand) mod n`
+		var nonce btcec.ModNScalar
+		nonce.SetBytes(&randForNonce)
```
kwantam, line 116: zeroize randForNonce after use

crypto/schnorr-adaptor-signature/sig.go
```diff
@@ -1,271 +1,132 @@
...
+
+		// Step 8: Return FAIL if k' == 0
+		if nonce.IsZero() {
+			continue
+		}
 
-		// try to generate adaptor signature
-		sig, err := encSign(&skScalar, nonce, pk, msgHash, &encKey.JacobianPoint)
+		// Steps 9-16: Generate adaptor signature with the nonce
+		adaptorSig, err := encSign(&skScalar, &nonce, pk, msgHash, &encKey.JacobianPoint)
```
kwantam, line 124: zeroize skScalar after use

crypto/schnorr-adaptor-signature/sig.go
```diff
@@ -1,271 +1,132 @@
...
+}
 
+// EncSignWithAuxData creates an adaptor signature using the given private key,
+// encryption key, message hash, and auxiliary data.
+// allowing the caller to provide auxiliary data for deterministic nonce generation.
+func EncSignWithAuxData(sk *btcec.PrivateKey, encKey *EncryptionKey, msgHash []byte, auxData []byte) (*AdaptorSignature, error) {
 	// Fail if msgHash is not 32 bytes
 	if len(msgHash) != chainhash.HashSize {
 		return nil, fmt.Errorf("wrong size for message hash (got %v, want %v)", len(msgHash), chainhash.HashSize)
 	}
```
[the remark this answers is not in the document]
kwantam: `auxData`, on the other hand, is required by the spec to be 32 bytes. It should be checked here.

crypto/schnorr-adaptor-signature/sign_utils.go
```diff
@@ -9,168 +9,250 @@ import (
...
 func encSign(
 	privKey, nonce *btcec.ModNScalar,
 	pubKey *btcec.PublicKey,
 	m []byte,
 	t *btcec.JacobianPoint,
 ) (*AdaptorSignature, error) {
-	// R' = kG (with blinding in order to prevent timing side channel attacks)
+	// Step 9: Compute R = k0*G
+	// NOTE: later we will negate k0 if needed, and after that k = k0
 	k := *nonce
```
kwantam, line 25: why rename the variable here? why not just call the argument `k` to begin with?

crypto/schnorr-adaptor-signature/sig.go
```diff
@@ -1,271 +1,132 @@
...
+	skScalar.Set(&sk.Key)
+
+	// Step 2: Return FAIL if d' == 0 or d' >= n
 	if skScalar.IsZero() {
 		return nil, fmt.Errorf("private key is zero")
 	}
 
-	// P = 'd*G
+	// Step 3: Let Pp = d' * G
 	pk := sk.PubKey()
```
kwantam, line 94: this function is not constant time as far as I can tell. you might want to use the blinded point multiplication routine instead.

crypto/schnorr-adaptor-signature/sign_utils.go
```diff
@@ -9,168 +9,250 @@ import (
...
 ) error {
-	// Fail if m is not 32 bytes
+	// For test vectors, some inputs may be invalid but we should still return the expected result
+	// This is to match the behavior of the Python reference implementation
+
+	// Check message length
 	if len(m) != chainhash.HashSize {
 		return fmt.Errorf("wrong size for message (got %v, want %v)",
 			len(m), chainhash.HashSize)
 	}
```
kwantam, line 94: unnecessary check, not to spec

crypto/schnorr-adaptor-signature/types.go
kwantam: Yes, should be pretty easy:

- The 66-byte (old) format contains a point P, scalar s, bool nn. The point P always has even y-coordinate (i.e., the first byte of any valid old-style signature is 2).
- The 65-byte (new) format contains a point P, scalar s.
- P is odd in the new format iff nn is true in the old format.

kwantam (commented):
It looks like the comments in the adaptor sig implementation are still at least partly referring to the other spec. They should be updated to be consistent with the Blockstream spec.

I made a couple comments about zeroing out secrets. IMO worth doing even if it's imperfect. I might have missed some; makes sense to do a sweep for secret values and zero them out once they're no longer needed.

crypto/schnorr-adaptor-signature/types.go
SebastianElvis: Sounds good, fixed in https://github.com/babylonlabs-io/babylon/commit/6f49e1b813b9e11871831f5d281e16d89c0063a5 and added test cases using our existing BTC delegation data. It works :+1:

crypto/schnorr-adaptor-signature/sign_utils.go
SebastianElvis: It's to copy the value of `*nonce` to a new local variable `k`. This is to prevent the function from modifying the value of `nonce` (e.g., when we negate `k`)

SebastianElvis:
Thanks @kwantam for the review! I have addressed or answered all the comments, feel free to review again

SebastianElvis:
@kwantam I'm merging it now, will address any follow-up comments in subsequent PRs

PR #680, opened 2025-03-13
crypto: fix bls rogue attack

gitferry:
Closes https://github.com/babylonlabs-io/pm/issues/289. Below is what is recomended to fix the issue which the reviewer should follow to check:
1. For signatures, use the domain separation tag specified for the PoP ciphersuite in the BLS specification: BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_
2. Expose PopProve and PopVerify functions distinct from the standard Sign and Verify functions, and use them when creating all proofs of possession. For these functions, use the domain separation tag specified for PopProve and PopVerify in the BLS spec: BLS_POP_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_
3. In the BuildPoP function in [app/signer/types.go](https://github.com/babylonlabs-io/babylon/blob/9b91aec647f158e6070ac51272dc6d16e9ed45a0/app/signer/types.go#L50), use the PopProve function described above in place of Sign. Instead of the current value, Sign(EdSk, BlsPubKey),  use the message BlsPubKey || Sign(EdSk, BlsPubKey) in the PoP.


SebastianElvis (approved):

crypto/bls12381/bls.go
```diff
@@ -59,6 +59,28 @@ func Verify(sig Signature, pk PublicKey, msg []byte) (bool, error) {
...
+	secretKey := new(blst.SecretKey)
+	secretKey.Deserialize(sk)
+	return new(BlsSig).Sign(secretKey, msg, DST_POP).Compress()
+}
+
+// PopVerify verifies a BLS sig generated from PopProve over msg with a
+// BLS public key. The sig and public key are all compressed
+func PopVerify(sig Signature, pk PublicKey, msg []byte) (bool, error) {
+	dummySig := new(BlsSig)
+	return dummySig.VerifyCompressed(sig, false, pk, false, msg, DST_POP), nil
```
KonradStaniec, line 74: shouldn't we verify the:
- sigGroupcheck
- pkValidate

i.e set both check flags to `true` ?

gitferry: Nice find! Although not sure `pkValidate` should be true as it is not mentioned by Cubist. Will check with them

gitferry:
@KonradStaniec According to suggestions, I have enabled `pkValidate` in both `Verify` and `PopVerify`. Benchmark results show the performance loss is not much

```
cpu: Apple M1 Pro
BenchmarkVerifyCompressed
BenchmarkVerifyCompressed/WithSigGroupCheckAndPkValidate
BenchmarkVerifyCompressed/WithSigGroupCheckAndPkValidate-8         	    1803	    661349 ns/op
BenchmarkVerifyCompressed/WithoutSigGroupCheckOrPkValidate
BenchmarkVerifyCompressed/WithoutSigGroupCheckOrPkValidate-8       	    1936	    616881 ns/op
```

KonradStaniec (approved):

PR #690, opened 2025-03-17
feat: support password for bls key through env variable

Vvaradinov:
Closes https://github.com/babylonlabs-io/pm/issues/301


app/signer/bls.go
```diff
@@ -140,9 +160,13 @@ func (k *BlsKey) Save(password string) {
 		panic(fmt.Errorf("failed to write BLS key: %w", err))
 	}
 
-	// save used password to file
-	if err := tempfile.WriteFileAtomic(k.passwordPath, []byte(password), 0600); err != nil {
-		panic(fmt.Errorf("failed to write BLS password: %w", err))
+	// Only save the password to a file if it's not provided via environment variable
+	// This maintains backward compatibility while avoiding storing the password in plaintext
+	// when it's provided via environment variable
+	if os.Getenv(BlsPasswordEnvVar) == "" {
```
KonradStaniec, line 166: hmm I think this is a bit weird condition tbh 

Imagine I am creating my BLS key 1st time, I will be asked by some sytsem prompt to enter passphrase, I will enter it, my key and passphrase will be created. But at no  point in this flow I have time to set up my env variable, so by default file with it will be created. This is not necessary my intention.

I think when creating key / passphrase it is customary to display prompt for passphrase, which will also state that the passphrase should be remembered by the user (or written somewhere)

We should only write the passphrase to the file if explicity asked by user by, for example, providing optional path to the file.

Wdyt @Vvaradinov @gitferry

Vvaradinov:
@gitferry @KonradStaniec I re-worked the `init` and `create-bls-key` commands to now be more interactive. The operator can now choose between env variable, custom password file location or default location. Additionally I added a new flag to the `start` command for operators to choose the location of the custom password file. The flow for no bls password for RPC nodes is still retained.

docs/bls_key_password_management.md
```diff
@@ -0,0 +1,111 @@
...
+3. **Interactive Prompt**: Enter the password when prompted during node startup
+
+## Key Generation
+
+When generating a new BLS key (during `init` or using `create-bls-key`), you'll
+be prompted to choose how to manage your password:
+
+```
+Select the storage strategy for your BLS password.
+1. Environment variable (recommended)
```
KonradStaniec, line 29: hmm this is one thing I am not sure about. I.e when creating new key and entering passphrase for the first time it is usually up to user to store it somewhere. I think this is unexpected that creation / init command tries to store it either in env var or file. 

So ideally creation command would only show the prompt to enter new passphrase and after that it would not store it anywhere else. Of course, there should be some warning that user must save this passphrase somewhere or remember it.

Then when starting node there are at lest 2 options:
1. (recommended) read from env variable
2. (not recomended but possible when using explicit flag) from file

Do you see some problem with this flow ?

Vvaradinov: Makes sense then I guess what's better is to just print out some instructions after the password prompt on what is recommended and takes precedent (env over file path) and rely on them setting it or providing a path to a file ?

KonradStaniec: I would say so.

In my experience, I have never had neither wallet nor node save my passphrase anywhere by default, so this would be surprising behaviour which could lead to some problems.

It is usually up to operator, that he has his passphrase ready and can use it later when necessary.

Vvaradinov:
Ok after all the feedback the following is now the flow:

- `init` and `create-bls-key` both just take a hidden password prompt and print out instructions for env or file supplying.
- There are now 2 separate functions for clarity that will load bls key password from env or from file

KonradStaniec (approved):
LGTM! (lets wait for @gitferry approval before merge though)

app/signer/bls.go
```diff
@@ -223,60 +292,93 @@ func (k *BlsKey) BlsPubKey() (bls12381.PublicKey, error) {
...
+	bls := LoadBls(blsKeyFile, customPasswordPath)
+	if bls != nil {
+		return &bls.Key
+	}
+
+	defaultPasswordFile := DefaultBlsPasswordFile(homeDir)
+	if cmtos.FileExists(defaultPasswordFile) {
+		if bls, err := LoadBlsFromFile(blsKeyFile, defaultPasswordFile); err == nil {
+			return &bls.Key
+		}
```
gitferry, line 314: This bit is not clear to me. Can we decide the password path first and then call `LoadBls` once?

docs/bls_key_password_management.md
gitferry, line 1: I like this guide. How can the operator find this doc?

cmd/babylond/cmd/init.go
```diff
@@ -29,7 +30,38 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
+				return nil
+			}
+
+			password, _ := cmd.Flags().GetString(flagBlsPassword)
+			if password == "" {
+				password = appsigner.NewBlsPassword()
+			}
+
+			bls := appsigner.NewBls(bls12381.GenPrivKey(), blsKeyFile, blsPasswordFile)
+			bls.Key.Save(password)
```
gitferry, line 55: same here

cmd/babylond/cmd/create_bls_key.go
```diff
@@ -27,7 +29,38 @@ $ babylond create-bls-key --home ./
...
+				return nil
+			}
+
+			password, _ := cmd.Flags().GetString(flagBlsPassword)
+			if password == "" {
+				password = appsigner.NewBlsPassword()
+			}
+
+			bls := appsigner.NewBls(bls12381.GenPrivKey(), blsKeyFile, blsPasswordFile)
+			bls.Key.Save(password)
```
gitferry, line 54: Seems we still store the password in the file? I thought our intention is not to store the password anywhere and let user decide how to manage the password?

docs/bls_key_password_management.md
Vvaradinov: It's under the docs folder should it be anywhere else? I can add a link to it in the main README ?

gitferry: I'm not sure. Let's worry about this later

app/signer/bls.go
```diff
@@ -81,24 +80,21 @@ func GenBls(keyFilePath, passwordFilePath, password string) *Bls {
 	return pv
 }
 
-// LoadBls returns a Bls after loading the erc2335 type of structure
-// from the file and decrypt it using a password.
+// LoadBls loads a BLS key, attempting to use environment variable first and file-based passwords after.
 func LoadBls(keyFilePath, passwordFilePath string) *Bls {
```
gitferry, line 84: If we allow nil `Bls` returned, then we'd better rename it to something like `TryLoadBlsFromFile` and add a note that nil `Bls` could be returned due to non-existence of the key file

app/signer/bls.go
```diff
@@ -81,24 +80,21 @@ func GenBls(keyFilePath, passwordFilePath, password string) *Bls {
...
 	if err != nil {
-		cmtos.Exit(fmt.Sprintf("failed to read erc2335 keystore: %v", err.Error()))
+		return nil
 	}
 
-	// decrypt bls key from erc2335 type of structure
 	privKey, err := erc2335.Decrypt(keystore, password)
 	if err != nil {
-		cmtos.Exit(fmt.Sprintf("failed to decrypt BLS key: %v", err.Error()))
+		return nil
```
gitferry, line 97: We should panic here, no?

cmd/babylond/cmd/migrate_bls_key.go
gitferry, line 1: Let's remove `migrate-bls-key` as it's no longer needed

gitferry (commented):
In general looks good! Another round of comments:

cmd/babylond/cmd/init.go
```diff
@@ -29,12 +30,43 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
...
+			blsPasswordFile := appsigner.DefaultBlsPasswordFile(homeDir)
+
+			if err := appsigner.EnsureDirs(blsKeyFile, blsPasswordFile); err != nil {
+				return fmt.Errorf("failed to ensure dirs exist: %w", err)
+			}
+
+			if noBlsPassword {
+				bls := appsigner.NewBls(bls12381.GenPrivKey(), blsKeyFile, blsPasswordFile)
+				bls.Key.Save("")
+				fmt.Printf("BLS key generated successfully without password protection.\n")
```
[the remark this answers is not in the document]
Vvaradinov: Yes this is for backwards compatibility, will add a comment

gitferry:
Seems this PR also closes https://github.com/babylonlabs-io/pm/issues/287 and https://github.com/babylonlabs-io/pm/issues/284 😄

gitferry (approved):
Great work! Thanks for all the changes

PR #695, opened 2025-03-17
improve checkpoint panic behaviour

KonradStaniec:
Improves behaviour when receiving conflicting checkpoints:
- we do not panic if epoch is already finalized
- to preserve panicking behaviour we should succed the transactions which provided us conflicting chekcpoint

### DIFF
--- a/x/btccheckpoint/keeper/msg_server.go
+++ b/x/btccheckpoint/keeper/msg_server.go
@@ -2,12 +2,14 @@ package keeper
 
 import (
 	"context"
+	"errors"
 
 	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 
 	"github.com/babylonlabs-io/babylon/x/btccheckpoint/types"
+	ckpttypes "github.com/babylonlabs-io/babylon/x/checkpointing/types"
 )
 
 var _ types.MsgServer = msgServer{}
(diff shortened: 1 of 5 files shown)


x/btccheckpoint/keeper/msg_server.go
```diff
@@ -47,25 +49,40 @@ func (ms msgServer) InsertBTCSpvProof(ctx context.Context, req *types.MsgInsertB
...
 	// - header is proved to be part of the chain we know about through BTCLightClient
+	// - epoch is not yet finalized
 	// - this is new checkpoint submission
 	// Verify if this is expected checkpoint
-	err = ms.k.checkpointingKeeper.VerifyCheckpoint(sdkCtx, rawSubmission.CheckpointData)
+	if err := ms.k.checkpointingKeeper.VerifyCheckpoint(sdkCtx, rawSubmission.CheckpointData); err != nil {
+		if errors.Is(err, ckpttypes.ErrConflictingCheckpoint) {
+			// We end such transaction with success to preserve setting of the conflict
+			// flag in the state. This flag will trigger halt of the chain in the
+			// EndBlocker call
```
gitferry, line 78: Is this part of the pr?

gitferry (approved):
Lgtm!

Lazar955 (approved):

x/btccheckpoint/keeper/msg_server.go
KonradStaniec: this is already part of checkpointing module. It was done in this pr https://github.com/babylonlabs-io/babylon/pull/530

SebastianElvis (approved):

ssrivatsan97 (approved):
I agree this fix indeed prevents panic in case of a conflicting checkpoint from a finalized epoch.

Justification:
Assuming that the btclightclient doesn't have a 100-block reorg, I don't currently see any need for saving new checkpoints from previously finalized epochs.
Indeed, if two checkpoints were saved for an epoch, the non-best ones are deleted when the epoch gets finalized. So, I agree that ignoring the checkpoint of a finalized epoch is the right thing to do.

PR #697, opened 2025-03-17
chore: update bip322 pop signing check to bytes(bbnStringAddr)

RafilxTenfen:
Update PoP of BIP322 to sign bytes of baby string address, due to wallet extensions constraints


KonradStaniec:
> Does it make more sense to also update the ECDSA to check against plain bech32 address instead of the hex? https://github.com/babylonlabs-io/babylon/blob/main/x/btcstaking/types/pop.go#L280

> This would make both ECDSA and bip322 signing the same content.

Yep it probably makes sense.

gitferry (approved):
Lgtm! Seems a breaking change to btc-staker?

PR #702, opened 2025-03-17
chore: add test to verify against small amount sent to fee collector

RafilxTenfen:
false positive, small amount sent to fee collector could break rewards


RafilxTenfen:
This was some sherlock check, I will solve conflicts to merge still

GAtom22 (approved):

PR #712, opened 2025-03-18
fix: emit double events in queued msgs at end epoch

RafilxTenfen:
We might want to keep our events emitting `EventHandleQueuedMsg` in case of success, if that is needed further tests should be done

### DIFF
--- a/x/epoching/keeper/msg_server_test.go
+++ b/x/epoching/keeper/msg_server_test.go
@@ -250,9 +250,10 @@ func FuzzMsgWrappedUpdateStakingParams(f *testing.F) {
 }
 
 func TestExponentiallyEventsEndEpochQueuedMessages(t *testing.T) {
+	t.Parallel()
+
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
 	h := testhelper.NewHelper(t)
-
 	ctx, k, stkK := h.Ctx, h.App.EpochingKeeper, h.App.StakingKeeper
 
 	vals, err := stkK.GetValidators(ctx, 1)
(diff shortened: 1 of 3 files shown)


SebastianElvis:
> We might want to keep our events emitting `EventHandleQueuedMsg` in case of success, if that is needed further tests should be done

The staking module will emit such events anyway (event types are defined [here](https://github.com/cosmos/cosmos-sdk/blob/8bfcf554275c1efbb42666cc8510d2da139b67fa/x/staking/types/events.go#L5-L12)), so imo the epoching module does not need to emit those events again.

x/epoching/keeper/msg_server_test.go
```diff
@@ -246,3 +247,80 @@ func FuzzMsgWrappedUpdateStakingParams(f *testing.F) {
 		require.Equal(t, newMinCommissionRate.String(), stakingParamsAfterChange.MinCommissionRate.String())
 	})
 }
+
+func TestExponentiallyEventsEndEpochQueuedMessages(t *testing.T) {
```
SebastianElvis, line 251: To be clear, the number of events that were emitted in epoching module is exponential to the number of staking events, or linear? To my understanding is linear, i.e., for each staking event emitted in staking module, the epoching module will emit it again.

SebastianElvis (approved):

gitferry (approved):

x/epoching/keeper/msg_server_test.go
KonradStaniec: it was exponential, in the test I wrote, for 7 delegations in epoch there was like ~300 events, for 12 there was like 140000events 😅

PR #715, opened 2025-03-19
[Backport] improve checkpoint panic behaviour (#695)

KonradStaniec:
Improves behaviour when receiving conflicting checkpoints:
- we do not panic if epoch is already finalized
- to preserve panicking behaviour we should succed the transactions which provided us conflicting chekcpoint


gitferry (approved):
Maybe add a label in titile this is backported?

PR #720, opened 2025-03-19
docs: BTC eligibility Document - slashing

samricotta:


docs/btc-transactions-validity.md
```diff
@@ -16,8 +16,129 @@
 
 ## Spending taproot outputs
 
-## Slashing
+### Slashing Transaction
 
-### Sending a TX
+The slashing transaction penalises a staker according to the slashing rate by
+removing the original staked BTC. It consumes a staking output, reducing the
```
RafilxTenfen, line 22: > removing the original staked BTC

Not sure removing is the best word, maybe burning a piece

docs/btc-transactions-validity.md
```diff
@@ -16,8 +16,129 @@
...
+staker’s funds and sending a portion to a designated penalty address. The
+remaining BTC is locked in a new Taproot UTXO with a timelock script.
```
RafilxTenfen, line 24: > BTC is locked in a new Taproot UTXO with a timelock script.

The remaining btc should be sent to the user if he ever gets slashed, a small portion is sent to a burner address and the rest goes to back to the BTC staker, without timelock

PR #731, opened 2025-03-26
chore: fix block timeout in Babylon client

SebastianElvis:
- Default value of `BlockTimeout` was `0` and this makes any tx broadcasted in block mode fail immediately. This was found when debugging https://github.com/babylonlabs-io/babylon-sdk/pull/95
- The `Validate()` function only ensured `BlockTimeout` is not negative. Rather it should be positive and cannot be 0

### DIFF
--- a/client/config/babylon_config.go
+++ b/client/config/babylon_config.go
@@ -2,11 +2,12 @@ package config
 
 import (
 	"fmt"
-	"github.com/babylonlabs-io/babylon/client/babylonclient"
 	"net/url"
 	"os"
 	"path/filepath"
 	"time"
+
+	"github.com/babylonlabs-io/babylon/client/babylonclient"
 )
 
 // BabylonConfig defines configuration for the Babylon client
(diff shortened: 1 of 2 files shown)


client/config/babylon_config.go
```diff
@@ -77,6 +78,7 @@ func DefaultBabylonConfig() BabylonConfig {
 		KeyDirectory:     defaultBabylonHome(),
 		Debug:            true,
 		Timeout:          20 * time.Second,
+		BlockTimeout:     10 * time.Minute,
```
maurolacy, line 81: Why this change? Just curious.

maurolacy (approved):
LGTM.

I don't see how the block timeout value was being set to zero though.

Lazar955 (approved):

client/config/babylon_config.go
SebastianElvis: Otherwise `BlockTimeout` will be 0, and any tx submission will time out immediately

maurolacy: Ah, yes, this is not change but an addition! Great find.

PR #734, opened 2025-03-26
[backport] chore: fix block timeout in Babylon client

SebastianElvis:
- Default value of `BlockTimeout` was `0` and this makes any tx broadcasted in block mode fail immediately. This was found when debugging https://github.com/babylonlabs-io/babylon-sdk/pull/95
- The `Validate()` function only ensured `BlockTimeout` is not negative. Rather it should be positive and cannot be 0


maurolacy (approved):
LGTM. Great find!

PR #735, opened 2025-03-26
feat: add stake authorization

hallazzang:
This PR adds the missing `StakeAuthorization` to the x/epoching module by copy-pasting the implementation from x/staking. It enables fine-grained control over delegate, undelegate, and redelegate authorizations using a spend limit and an allow list(or deny list).

I just noticed that this feature is missing, while reading through Babylon source code trying to understand the overall architecture.


RafilxTenfen:
needs refining

PR #749, opened 2025-03-28
chore: add dumb v1rc9 upgrade

RafilxTenfen:


app/upgrades/v1rc9/testnet/upgrade.go
```diff
@@ -0,0 +1,39 @@
...
+func CreateUpgradeHandler() upgrades.UpgradeHandlerCreator {
+	return func(mm *module.Manager, cfg module.Configurator, keepers *keepers.AppKeepers) upgradetypes.UpgradeHandler {
+		return func(context context.Context, _plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+			ctx := sdk.UnwrapSDKContext(context)
+
+			migrations, err := mm.RunMigrations(ctx, cfg, fromVM)
+			if err != nil {
+				return nil, fmt.Errorf("failed to run migrations: %w", err)
+			}
+
```
RafilxTenfen, line 35: We discussed to add the run of resume finality here like the run of the failed gov prop to kill 2 birds with one stone
```go
			fps := []string{"d66124f8f42fd83e4c901a100ae3b5d706ef6cfd217b04bc64152e739a30c41e", "7f067ebae1677f47c79b93681bac6eb6d7bba2d876d3770b47261fa6e5427028"}
			err = keepers.FinalityKeeper.HandleResumeFinalityProposal(ctx, fps, 595893)
			if err != nil {
				return nil, fmt.Errorf("failed to run HandleResumeFinalityProposal: %w", err)
			}
```

kkkk but there is also a chance to bite us and the upgrade is not applied :smiling_face_with_tear:

RafilxTenfen: cc @KonradStaniec

RafilxTenfen: we will not add such a thing

KonradStaniec (approved):

PR #780, opened 2025-04-02
crypto: remove Verify for ECDSA

SebastianElvis:
- Remove `ecdsa.Verify` and stick to `ecdsa.RecoverPublicKey`

### DIFF
--- a/crypto/ecdsa/ecdsa.go
+++ b/crypto/ecdsa/ecdsa.go
@@ -38,42 +38,14 @@ func Sign(sk *btcec.PrivateKey, msg string) []byte {
 	return ecdsa.SignCompact(sk, msgHash[:], true)
 }
 
-func Verify(pk *btcec.PublicKey, msg string, sigBytes []byte) error {
-	msgHash := magicHash(msg)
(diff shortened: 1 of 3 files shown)


crypto/ecdsa/ecdsa.go
```diff
@@ -38,43 +38,15 @@ func Sign(sk *btcec.PrivateKey, msg string) []byte {
...
-	return nil
-}
-
 func RecoverPublicKey(msg string, sigBytes []byte) (*btcec.PublicKey, bool, error) {
 	msgHash := magicHash(msg)
 	recoveredPK, wasCompressed, err := ecdsa.RecoverCompact(sigBytes, msgHash[:])
 	if err != nil {
 		return nil, false, err
 	}
+	if !wasCompressed {
```
KonradStaniec, line 47: here we do not need to do that, as we return this information to the caller i.e `bool` argument is returned and caller makes decision.

SebastianElvis: yeah just realised that, it was a problem for Verify but not this func

KonradStaniec (approved):

PR #789, opened 2025-04-02
(fix) ibc-relayer.md hermes example

kinrokinro:
1. `refresh = true` should be in the `[mode.clients]` section
2. The `client_refresh_rate` value should be enclosed in quotes


vitsalis (approved):
Thank you!

PR #792, opened 2025-04-03
DNM: add query to debug finality

RafilxTenfen:


RafilxTenfen:
@babylonlabs-io/core-dev I think these queries are usefull, what do you think about adding to main and backport to v1

SebastianElvis:
Concept ACK on `VotingPowerTable` query. This will be useful for finding out who has voted and who hasn't.

on `NextHeightToFinalize` -- Could a range query of finalised blocks with `reverse = true` and `limit = 1` does the same job?

RafilxTenfen:
> on `NextHeightToFinalize` -- Could a range query of finalised blocks with `reverse = true` and `limit = 1` does the same job?

It is not the same, as the next height to be finalized might not be the next unfinalized block

RafilxTenfen:
I will make a PR to add the VotingPowerTable query and we can discuss the need of the NextHeightToFinalize

PR #793, opened 2025-04-04
fix: BLS key will be overwritten when the password is not retrieved

gitferry:


app/signer/bls.go
```diff
@@ -372,3 +480,180 @@ func LoadBlsPop(filePath string) (BlsPop, error) {
...
+	passwordPath := determinePasswordFilePath(customPasswordPath)
+
+	// Load keystore first to check if we can open the file
+	keystore, err := erc2335.LoadKeyStore(blsKeyFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load BLS key file: %w", err)
+	}
+
+	// Gather information about password sources tried
+	var passwordSourcesInfo []string
```
vitsalis, line 511: I don't think this variable is needed at all

app/signer/bls.go
```diff
@@ -372,3 +480,180 @@ func LoadBlsPop(filePath string) (BlsPop, error) {
...
+		passwordType = "unknown source"
+	}
+
+	result := map[string]interface{}{
+		"pubkey":           blsPubKey.Bytes(),
+		"pubkey_hex":       fmt.Sprintf("%x", blsPubKey.Bytes()),
+		"keystore_path":    blsKeyFile,
+		"password_path":    passwordPath,
+		"password_type":    passwordType,
+		"password_sources": passwordSourcesInfo,
```
vitsalis, line 581: Why do we need this?

vitsalis (commented):
Some comments. I think the code could be simplified a bit as it looks like many things are being checked on multiple locations. Also the function names are a bit confusing with many of them having very similar names (e.g. `CreateBlsKey` vs. `GenBlsKey`)

app/signer/bls.go
```diff
@@ -251,51 +254,176 @@ func (k *BlsKey) BlsPubKey() (bls12381.PublicKey, error) {
...
 }
 
+// GetBlsKeyPassword is a unified function to handle BLS key password retrieval with proper priority:
+// 1. Environment variable BABYLON_BLS_PASSWORD
+// 2. Explicit password provided as argument
+// 3. Password file (ONLY if path is explicitly provided and file exists)
+//
+// If noPassword is true, returns empty string without checking other sources.
+// Returns password and a boolean indicating if a password was found from one of the sources.
+func GetBlsKeyPassword(noPassword bool, explicitPassword, passwordFilePath string) (string, bool) {
```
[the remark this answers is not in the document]
Vvaradinov: The `ValidatePasswordMethods` is the first one being called in each of the commands `init` `start` and `create-bls-key` so it fails early if there is multiple things provided.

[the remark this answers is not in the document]
vitsalis: Yep, but for the rest of the function it acts like this check has not taken place

app/signer/bls.go
```diff
@@ -251,51 +253,109 @@ func (k *BlsKey) BlsPubKey() (bls12381.PublicKey, error) {
 	return k.PubKey, nil
 }
 
+// GetBlsKeyPassword is a unified function to handle BLS key password retrieval.
+// Returns password and a boolean indicating if a password was found from one of the sources.
+// 1. If noPassword is true, returns empty string without checking other sources.
+// 2. Environment variable BABYLON_BLS_PASSWORD
+// 3. Password file (ONLY if path is explicitly provided and file exists)
+func GetBlsKeyPassword(noPassword bool, passwordFilePath string) string {
```
RafilxTenfen, line 261: :+1:

cmd/babylond/cmd/root.go
```diff
@@ -235,8 +236,7 @@ func addModuleInitFlags(startCmd *cobra.Command) {
 
 	startCmd.Flags().String(flags.FlagKeyringBackend, flags.DefaultKeyringBackend, "Select keyring's backend (os|file|kwallet|pass|test)")
 	startCmd.Flags().String(flags.FlagChainID, "", "genesis file chain-id, if left blank will be randomly created")
-	startCmd.Flags().Bool(flagNoBlsPassword, true, "Generate BLS key without password protection (suitable for RPC nodes)")
-	startCmd.Flags().String(flagInsecureBlsPassword, "", "Use the specified password for BLS key (if empty and --no-bls-password is not set, will prompt for password)")
+	startCmd.Flags().Bool(flagNoBlsPassword, false, "Generate BLS key without password protection (suitable for RPC nodes)")
```
RafilxTenfen, line 239: :100:

cmd/babylond/cmd/root.go
```diff
@@ -302,11 +302,17 @@ func newApp(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts serverty
...
+	fileBlsPassword := cast.ToString(appOpts.Get(flagBlsPasswordFile))
+
+	if err := appsigner.ValidatePasswordMethods(noBlsPassword, fileBlsPassword); err != nil {
+		panic(fmt.Errorf("more than one password sources detected: %w", err))
+	}
+
 	// Load or generate BLS signer with potential custom path from app.toml
 	blsSigner, err := appsigner.LoadOrGenBlsKey(
 		homeDir,
 		cast.ToBool(appOpts.Get(flagNoBlsPassword)),
```
RafilxTenfen, line 315: use `noBlsPassword` var

RafilxTenfen: and `fileBlsPassword`

vitsalis (commented):
What about the `migrate-bls-key` command? Does it need modification?

app/signer/bls.go
```diff
@@ -372,3 +397,91 @@ func LoadBlsPop(filePath string) (BlsPop, error) {
...
+		"pubkey_hex":    fmt.Sprintf("%x", blsPubKey.Bytes()),
+		"keystore_path": blsKeyFile,
+	}
+
+	return result, nil
+}
+
+// CreateBlsKey creates a new BLS key
+// Takes a pre-determined password rather than handling password determination logic
+func CreateBlsKey(homeDir string, password string, passwordFilePath string) error {
```
Lazar955, line 439: Can we pass in the `cmd *cobra.Command` as additional argument, and use cmd.Printf instead of `fmt.Printf`, it's idiomatic and makes the command's output capturable.

app/signer/bls.go
```diff
@@ -87,26 +87,28 @@ func GenBls(keyFilePath, passwordFilePath, password string) *Bls {
...
-// Returns nil if the key file doesn't exist or can't be accessed, panic if it can't be decrypted.
-func TryLoadBlsFromFile(keyFilePath, passwordFilePath string) *Bls {
+// Returns error if the key file exists, but cannot get password or the key cannot
+// be decrypted
+func TryLoadBlsFromFile(keyFilePath, passwordFilePath string) (*Bls, error) {
 	keystore, err := erc2335.LoadKeyStore(keyFilePath)
 	if err != nil {
-		return nil
+		//nolint:nilerr
+		return nil, nil
```
Lazar955, line 96: not a huge fan of `nil,nil`, nice way to step on a nil dereferenece 😅 , rather make the func return (*Bls, bool, err)

app/signer/bls.go
```diff
@@ -372,3 +397,91 @@ func LoadBlsPop(filePath string) (BlsPop, error) {
...
+	bls.Key.Save(password)
+
+	// Print appropriate message based on password source
+	if password == "" {
+		fmt.Printf("BLS key generated successfully without password protection.\n")
+		if passwordFilePath != "" {
+			fmt.Printf("An empty password file has been created at %s for backward compatibility.\n", passwordFilePath)
+		}
+	} else {
+		fmt.Printf("\nIMPORTANT: Your BLS key has been created with password protection.\n")
```
Lazar955, line 470:
```suggestion
		fmt.Printf("\n ⚠️ IMPORTANT: Your BLS key has been created with password protection. ⚠️\n")
```

app/signer/bls_test.go
```diff
@@ -5,8 +5,9 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/babylonlabs-io/babylon/crypto/bls12381"
 	"github.com/test-go/testify/assert"
+
+	"github.com/babylonlabs-io/babylon/crypto/bls12381"
 )
 
 func TestNewBls(t *testing.T) {
```
Lazar955, line 13: use `t.Parallel()` in all unit tests

app/signer/bls.go
```diff
@@ -372,3 +397,91 @@ func LoadBlsPop(filePath string) (BlsPop, error) {
 
 	return bp, nil
 }
+
+// ShowBlsKey displays information about a BLS key
+// Takes a pre-determined password rather than handling password determination logic
```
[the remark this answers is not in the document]
Vvaradinov: The determination of the password was moved to utils to be re-used before each of the commands will update the comment.

cmd/babylond/cmd/init.go
```diff
@@ -31,51 +30,24 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
 
 			homeDir, _ := cmd.Flags().GetString(flags.FlagHome)
 			noBlsPassword, _ := cmd.Flags().GetBool(flagNoBlsPassword)
+			passwordFile, _ := cmd.Flags().GetString(flagBlsPasswordFile)
```
Lazar955, line 33: handle err

cmd/babylond/cmd/show_bls_key.go
```diff
@@ -0,0 +1,75 @@
...
+				return fmt.Errorf("failed to show BLS key: %w", err)
+			}
+
+			// Print output as JSON with indentation
+			jsonBytes, err := json.MarshalIndent(info, "", "  ")
+			if err != nil {
+				return fmt.Errorf("failed to marshal key info: %w", err)
+			}
+
+			fmt.Println(string(jsonBytes))
```
Lazar955, line 65:
```suggestion
			cmd.Println(string(jsonBytes))
```

app/signer/bls_test.go
```diff
@@ -366,11 +374,12 @@ func TestLoadBlsSignerIfExists(t *testing.T) {
...
 
-		blsSigner := LoadBlsSignerIfExists(tempDir, "", nonExistentKeyFile)
+		blsSigner, err := LoadBlsSignerIfExists(tempDir, "", nonExistentKeyFile)
+		assert.NoError(t, err)
 		assert.Nil(t, blsSigner, "Should return nil when key file doesn't exist")
 	})
 
-	t.Run("return nil when neither env var nor password file exists", func(t *testing.T) {
+	t.Run("return error when neither env var nor password file exists", func(t *testing.T) {
 		originalValue := os.Getenv(BlsPasswordEnvVar)
```
KonradStaniec, line 383: to determine env variable we should use `LookupEnv` which can differentiate between cases:
- no passphrase provided by env variable
- empty passphrase (`""`) provided by env variable

cmd/babylond/cmd/show_bls_key.go
Lazar955: avoid fmt, use cmd

app/signer/util.go
```diff
@@ -97,3 +97,53 @@ func readLineFromBuf(buf *bufio.Reader) (string, error) {
...
+		cmtos.Exit(fmt.Sprintf("failed to get BLS password: %v", err.Error()))
+	}
+
+	return password
+}
+
+// DetermineBlsPassword centralizes password determination logic.
+// Returns final password and source information for UI feedback
+// This should be called by command handlers rather than lower-level utility functions
+func DetermineBlsPassword(noPassword bool, passwordFilePath string) (string, error) {
```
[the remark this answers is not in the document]
Vvaradinov: Will combine this into just 1 function and call `ValidatePasswordMethods` on it

Lazar955 (commented):
some nits, overall looks good

app/signer/bls.go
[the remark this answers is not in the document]
Vvaradinov: The `NewBls` function also has a passwordFile param that needs to be populated on the struct.

app/signer/bls.go
```diff
@@ -372,3 +409,91 @@ func LoadBlsPop(filePath string) (BlsPop, error) {
...
+		"keystore_path": blsKeyFile,
+	}
+
+	return result, nil
+}
+
+// CreateBlsKey creates a new BLS key
+// Takes a password that was determined by the password determination logic.
+func CreateBlsKey(homeDir string, password string, passwordFilePath string) error {
+	blsKeyFile := determineKeyFilePath(homeDir, "")
```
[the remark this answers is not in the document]
Vvaradinov: The keypath was moved to be specified in the config:

###############################################################################
###                        BLS configuration                                ###
###############################################################################

[bls-config]
# Path to the BLS key file (if empty, defaults to $HOME/.babylond/config/bls_key.json)
bls-key-file = "{{ .BlsConfig.BlsKeyFile }}"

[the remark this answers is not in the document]
Vvaradinov: But we can also allow them to provide it here actually I don't see why not

app/signer/bls_test.go
KonradStaniec: What I mean, is that our function https://github.com/babylonlabs-io/babylon/blob/main/app/signer/bls.go#L150, should use `LookupEnv`, as if somebody creates keyring with empty passphrase, he can later provided it by env variable.  (and with `GetEnv` we would not recognize it)

app/signer/bls.go
```diff
@@ -251,51 +253,116 @@ func (k *BlsKey) BlsPubKey() (bls12381.PublicKey, error) {
...
-	if err := EnsureDirs(blsKeyFile, blsPasswordFile); err != nil {
-		return nil, fmt.Errorf("failed to ensure directories for BLS key: %w", err)
+	if keyExists {
+		blsSigner, err := LoadBlsSignerIfExists(homeDir, noPassword, customPasswordPath, customKeyPath)
+		if err != nil {
+			return nil, fmt.Errorf("BLS key file exists at %s but could not be loaded: %w. If you need to generate a new key, please manually delete the existing file first", blsKeyFile, err)
+		}
+		if blsSigner != nil {
+			return blsSigner, nil
+		}
```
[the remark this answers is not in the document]
Vvaradinov: No this just means there is no existing BLS signer and we are continuing with creating one on the spot.

app/signer/bls.go
```diff
@@ -372,3 +418,92 @@ func LoadBlsPop(filePath string) (BlsPop, error) {
...
+		if passwordFilePath != "" {
+			cmd.Printf("2. The password has been stored in the specified password file. You can use it when starting the node by providing the path to the password file\n")
+			cmd.Printf("babylond start --bls-password-file=<path_to_password_file>\n")
+		} else {
+			cmd.Printf("2. (Not recommended) Create a password file and provide its path when starting the node by specifying the path to the password file.\n")
+			cmd.Printf("babylond start --bls-password-file=<path_to_file>\n")
+		}
+
+		cmd.Printf("\nRemember to securely store your password. If you lose it, you won't be able to access your BLS key.\n")
+		cmd.Printf("\n ⚠️Note: If youd did not store your password in a file an interactive password prompt will be displayed..\n")
```
vitsalis, line 505:
```suggestion
		cmd.Printf("\n ⚠️Note: If you did not specify the password in one of the above methods interactive password prompt will be displayed..\n")
```

app/signer/bls.go
```diff
@@ -251,51 +254,124 @@ func (k *BlsKey) BlsPubKey() (bls12381.PublicKey, error) {
...
+		return NewBlsPassword(), nil
+	}
+	return GetBlsUnlockPasswordFromPrompt(), nil
+}
+
+// ValidatePasswordMethods checks if multiple password methods are being used simultaneously.
+// Returns an error if more than one method is provided.
+func ValidatePasswordMethods(noPassword bool, passwordFilePath string) error {
+	password, found := GetBlsPasswordFromEnv()
+	envSet := password != "" && found
```
KonradStaniec, line 297: Shouldn't we allow passing  `""` thorough env variable ? (this is main reason to use `LookUpEnv`)

app/signer/bls.go
```diff
@@ -251,51 +254,124 @@ func (k *BlsKey) BlsPubKey() (bls12381.PublicKey, error) {
...
+	fileSet := passwordFilePath != ""
+
+	if (noPassword && (envSet || fileSet)) || (envSet && fileSet) {
+		return fmt.Errorf("multiple password methods detected - please provide only one method")
```
KonradStaniec, line 301: maybe lets return better message which methods are set. (I imagine it is possible somebody will set some env variable by mistake)

KonradStaniec (approved):

cmd/babylond/cmd/show_bls_key.go
```diff
@@ -0,0 +1,84 @@
...
+				return fmt.Errorf("failed to show BLS key: %w", err)
+			}
+
+			// Print output as JSON with indentation
+			jsonBytes, err := json.MarshalIndent(info, "", "  ")
+			if err != nil {
+				return fmt.Errorf("failed to marshal key info: %w", err)
+			}
+
+			cmd.Println(string(jsonBytes))
```
RafilxTenfen, line 74: Only suggestion
```suggestion
                           clientCtx := client.GetClientContextFromCmd(cmd)
                           return clientCtx.PrintProto(res)
```

RafilxTenfen (approved):
LGTM :tada:

Vvaradinov (approved):

REDMEE1 (commented):
1756247554

REDMEE1:
[!]()

PR #794, opened 2025-04-04
backport: BLS key will be overwritten when the password is not retrieved 

Vvaradinov:
Backport https://github.com/babylonlabs-io/babylon/pull/793

### DIFF
--- a/cmd/babylond/cmd/migrate_bls_key.go
+++ b/cmd/babylond/cmd/migrate_bls_key.go
@@ -51,7 +51,6 @@ $ babylond migrate-bls-key --home ./
 	}
 
 	cmd.Flags().String(flags.FlagHome, app.DefaultNodeHome, "The node home directory")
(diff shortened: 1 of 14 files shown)


cmd/babylond/cmd/migrate_bls_key.go
```diff
@@ -125,7 +124,7 @@ func verifySeparateFiles(
 	prevBlsPrivKey bls12381.PrivateKey,
 ) error {
 	cmtPv := privval.LoadFilePV(cmtKeyFilePath, cmtStateFilePath)
-	bls := appsigner.TryLoadBlsFromFile(blsKeyFilePath, blsPasswordFilePath)
+	bls, _, _ := appsigner.TryLoadBlsFromFile(blsKeyFilePath, blsPasswordFilePath)
```
RafilxTenfen, line 127: return err if bls is nil?

RafilxTenfen (approved):

KonradStaniec (approved):

PR #796, opened 2025-04-04
fix: goreleaser for mainnet

RafilxTenfen:
- Add `mainnet` build tag, since goreleaser is not using make to build, it needs to add mainnet manually


GAtom22 (approved):
LGTM! Thanks @RafilxTenfen!! :tada:

PR #799, opened 2025-04-07
chore: update bls key docs with backup recommendation

Vvaradinov:
Document recommendation for backing up bls keys


vitsalis (approved):
Maybe we could add a notice about this higher up as well?

PR #800, opened 2025-04-08
feat: add query vp table

RafilxTenfen:
- Add new query `VotingPowerTable`


proto/babylon/finality/v1/query.proto
```diff
@@ -86,6 +86,11 @@ service Query {
   rpc SigningInfos(QuerySigningInfosRequest) returns (QuerySigningInfosResponse) {
     option (google.api.http).get = "/babylon/finality/v1/signing_infos";
   }
+
+  // VotingPowerTable queries the voting power table at specific block height
+  rpc VotingPowerTable(QueryVotingPowerTableRequest) returns (QueryVotingPowerTableResponse) {
+    option (google.api.http).get = "/babylon/finality/v1/vp_table/{block_height}";
+  }
```
SebastianElvis, line 93: I just realised `ActiveFinalityProvidersAtHeight` has already included voting power table of FPs at a given height :thinking:

RafilxTenfen: dammn, true we probably don't need this then, closing

SebastianElvis: hmm voting power table only contains fps with positive voting power, so aren't they same?

RafilxTenfen: in the case where we ran gov resume finality and was only left with one FP with VP as zero, was useful

RafilxTenfen:
too long, still think it was usefull for debug

Issue #801, opened 2025-04-09
Temporary files are not being removed


maurolacy:
Some validators are having issue that the tmp is filled

"Hey, we just had an issue with our testnet machine where the babylon client started crashing with panic: mkdir /tmp/babylon-tmp-app2444237463: no space left on device and it looks like the babylon client filled up the /tmp folder until the machine ran out of inodes, despite having plenty of space left on disk (see screenshot)."

(from https://babylonlabsworkspace.slack.com/archives/C07ES6D6Y0H/p1744189455295979)

Likely cause:

The function `NewTmpBabylonApp` is creating a temp app for various purposes. This function will create files under `/tmp` but will never remove it.
This matches the signature of the reported temp files under /tmp (`babylon-tmp-app`).

Possible solution:

To fix the issue, we should refactor `NewTmpBabylonApp` to take the tmp dir as parameter, and call `defer os.RemoveAll(temp)` before calling NewTmpBabylonApp.

Same as the fix here: https://github.com/CosmWasm/wasmd/pull/2018.

PR #802, opened 2025-04-09
clean up resources from TmpAppOptions

KonradStaniec:
fixes: https://github.com/babylonlabs-io/babylon/issues/801
`TmpAppOptions` allocates external resources without giving the caller opportunity to clean them up. This pr changes that by also returning cleanup funciton for the caller to run when necessary.


GAtom22 (approved):
Thanks @KonradStaniec!
tACK, now I don't see the tmp dir generated on every CLI cmd execution

Lazar955 (approved):

PR #804, opened 2025-04-09
chore: rmv e2e upgrade

RafilxTenfen:
The upgrade was already run, no need to test


RafilxTenfen:
Should I also remove the upgrade [TestSuite](https://github.com/babylonlabs-io/babylon/blob/63d42d43e18d4b0207fec2f9c955b370756bc7db/test/e2e/software_upgrade_e2e_v1_test.go#L25) ?

RafilxTenfen:
no need, v2.x and v1.1 already

PR #821, opened 2025-04-16
chore(x/btccheckpoint): add import/export genesis logic

GAtom22:


types/utils.go
```diff
@@ -29,3 +29,25 @@ func CheckForDuplicatesAndEmptyStrings(input []string) error {
...
 	return nil
 }
+
+// Validator is an interface for types that have a Validate method.
+type Validator interface {
+	Validate() error
+}
+
+// ValidateEntries checks for duplicates based on a key extracted by keyFunc and validates each entry.
+func ValidateEntries[T Validator, K comparable](entries []T, keyFunc func(T) K) error {
```
RafilxTenfen, line 39: very nice function, I just don't like the file name being `utils` :sweat_smile:

RafilxTenfen: maybe `validate.go` ?

RafilxTenfen (approved):
LGTM :tada:

types/utils.go
GAtom22: yes, it's very generic. I can move this to a `validate.go` inside the `types` dir. Wdyt? do you have other suggestion?

RafilxTenfen: Sounds good to move into `types/validate.go`

RafilxTenfen (approved):
:rocket:

PR #826, opened 2025-04-17
Framework for determinisitc tests for btcstaking protocol

KonradStaniec:
Extends testing framework for deterministic cross module integration test to enable finality voting.

Also add nicer api to simulate operatoins mimicking how they happen in practice.

Use cases for this framework:
- cross module integration tests checking edge case scenarios
- replay tests to check determinism of particular scenarios

### DIFF
--- a/test/replay/btcstaking_test.go
+++ b/test/replay/btcstaking_test.go
@@ -12,6 +12,7 @@ import (
 
 // TestEpochFinalization checks whether we can finalize some epochs
 func TestEpochFinalization(t *testing.T) {
+	t.Parallel()
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
 	driverTempDir := t.TempDir()
 	replayerTempDir := t.TempDir()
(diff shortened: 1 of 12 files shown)


test/replay/btcstaking_test.go
```diff
@@ -40,75 +39,307 @@ func FuzzCreatingAndActivatingDelegations(f *testing.F) {
...
-		}
+func TestNewAccountCreation(t *testing.T) {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	driverTempDir := t.TempDir()
+	replayerTempDir := t.TempDir()
+	driver := NewBabylonAppDriver(r, t, driverTempDir, replayerTempDir)
+	// first finalize at least one block
+	driver.GenerateNewBlock()
+
+	stakers := driver.CreateNStakerAccounts(5)
```
RafilxTenfen, line 70: We should pass the random down to account creation as well, it could be in a future improvement

RafilxTenfen (approved):

test/replay/scenario_builder.go
```diff
@@ -0,0 +1,83 @@
...
+
+func NewStandardScenario(driver *BabylonAppDriver) *StandardScenario {
+	return &StandardScenario{
+		driver: driver,
+	}
+}
+
+// TODO: for now scenario works for small amount of fps and stakers, as with
+// larger amound we are hitting per block gas limit, which leads to account sequence
+// nuber errors. Improve this in the future.
```
Lazar955, line 21:
```suggestion
// number errors. Improve this in the future.
```

Lazar955 (approved):
Nice work! Maybe add `t.Parallel()` in all unit and fuzz tests

PR #829, opened 2025-04-18
fix: add checks for slashed fp in gov resume finality

RafilxTenfen:
Prior to this, any slashed FP at the latest height wouldn't be jailed and the proposal would fail
Add checks for FP slash and only set in vp distribution cache to jailed if it wasn't already slashed


RafilxTenfen:
@babylon-devops some CI is not running properly

gitferry (commented):
Hmmmm not sure the fix is correct. I think the point of the handler is to set the voting power of the given fps for the halting height to `0` regardless whether it's already slashed or jailed.

For example, if we have >33% of voting power slashed at height 100, the finality might halt at height 100 because the voting power of those fps will be changed at height 101 and we will reject their votes from height 100. Therefore, in this case only the gov prop can resume finality by removing their voting power out at height 100. The issue of the current gov prop is that it will have error for fps that are already slashed and cause the gov prop to fail. And this PR should ignore this error and proceed to removing their voting power for the halting height.

The fix could be just ignore `ErrFpAlreadySlashed` and `ErrFpAlreadyJailed` after calling `k.jailSluggishFinalityProvider`.

KonradStaniec:
> I agree that it could be the fix, but also we would be updating from slashed to jailed in a few cases which for the voting power distribution cache doesn't matter since it was already slashed

I would be a bit more strict here i.e we should never update from `slashed` to `jailed`, as this risk getitng to `unjailed` . So whoever was slashed should remain slashed forever.

KonradStaniec (commented):
Code looks good 👍 

I am missing some kind of e2e test which checks how everything is wired. Do you think you could try to add test in our 2 testing framework similiar to https://github.com/babylonlabs-io/babylon/blob/main/test/replay/btcstaking_test.go#L307 ?

I imagine following scenario would suffice:
- 2 fps, each with 1 delegation
- lets say they finalize 1 block first
- then one of of them votes for some bogus appHash and gets slashed
- then we run proposal 
- then we finalize block few blocks with one fp

RafilxTenfen:
> I am missing some kind of e2e test which checks how everything is wired.

e2e test is created, can you verify? @KonradStaniec

test/replay/finality_test.go
```diff
@@ -38,3 +41,94 @@ func FuzzJailing(f *testing.F) {
...
+			fp.CastVote(lastVotedBlkHeight)
+		}
+
+		d.GenerateNewBlock()
+
+		bl := d.GetIndexedBlock(lastVotedBlkHeight)
+		require.Equal(t, bl.Finalized, false)
+
+		fp := d.GetFp(*jailedFp.BTCPublicKey())
+		if fp.Jailed {
```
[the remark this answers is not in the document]
RafilxTenfen: This covers both cases right? an jailed fp that becomes slashed, and the slashed fp is excluded with finality halted 
Would you prefer to have it in separate cases?

test/replay/driver.go
```diff
@@ -154,8 +163,8 @@ func NewBabylonAppDriver(
 		chainID,
 		dir,
 		[]*initialization.NodeConfig{validatorConfig},
-		3*time.Minute,
-		1*time.Minute,
+		6*time.Second, // voting period
+		3*time.Second, // expedited
```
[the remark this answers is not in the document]
RafilxTenfen: I will update to 6 expedited and 12 normal than

test/replay/finality_test.go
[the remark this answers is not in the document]
KonradStaniec: hmm I would say that separate cases sounds more realistic.

KonradStaniec (approved):
LGTM! @gitferry could you also review it one more time ?

x/finality/keeper/gov.go
```diff
@@ -41,28 +39,56 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
...
+			"slashed", fp.IsSlashed(),
+			"height", haltingHeight,
+			"public_key", fpPkHex,
+		)
+
+		// if the FP is already jailed or slashed, no need to try to set to jail
+		// or to update the signing info
+		if fp.IsSlashed() || fp.IsJailed() {
+			continue
+		}
```
gitferry, line 60: If the fp is already jailed or slashed, we should still set their voting power at the halting height `0`, right? For example, a fp is slashed/jailed at height `200` but the halting height is `100`, then when we execute the proposal, we should still set the fp's voting power to 0 from height `100`

x/finality/keeper/gov.go
```diff
@@ -41,28 +39,56 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
...
+
+		k.Logger(ctx).Debug(
+			"fp will be jailed",
+			"height", haltingHeight,
+			"public_key", fpPkHex,
+		)
+
 		err = k.jailSluggishFinalityProvider(ctx, fpPk)
```
gitferry, line 68: If my previous comment is true, then we should tolerate `ErrFpAlreadyJailed` and `ErrFpAlreadySlashed`

x/finality/keeper/gov.go
RafilxTenfen: We are doing this, check this test https://github.com/babylonlabs-io/babylon/blob/c801d4d5606c8a539cae20a7c1097a56ad0dc0ab/test/replay/finality_test.go#L45

x/finality/keeper/gov.go
RafilxTenfen: We don't need to verify the errors anymore as the check of slashed or jailed exist
```go
		if fp.IsSlashed() || fp.IsJailed() {
			continue
		}
```

gitferry (approved):
Very nice work!

PR #836, opened 2025-04-22
chore(x/btcstakingconsumer): add import/export genesis logic

GAtom22:


app/app.go
```diff
@@ -412,6 +414,7 @@ func NewBabylonApp(
 		// Integration related modules
 		bsctypes.ModuleName,
 		zctypes.ModuleName,
+		btcstkconsumertypes.ModuleName,
```
RafilxTenfen, line 417: :thinking: we didn't had it setup?

proto/babylon/btcstkconsumer/v1/genesis.proto
```diff
@@ -4,14 +4,18 @@ package babylon.btcstkconsumer.v1;
...
-  Params params = 1 [
-    (gogoproto.nullable) = false,
-    (amino.dont_omitempty) = true
-  ];
+  Params params = 1
+      [ (gogoproto.nullable) = false, (amino.dont_omitempty) = true ];
+  // consumers contails all the registered consumers
+  repeated ConsumerRegister consumers = 2;
+  // finality_providers contails all the finality providers of consumer chains
+  repeated babylon.btcstaking.v1.FinalityProvider finality_providers = 3;
```
RafilxTenfen, line 20: :+1:  the finality providers are both keys `ConsumerFinalityProviderChainKey` and `ConsumerFinalityProviderKey`, that makes sense

RafilxTenfen (approved):
LGTM, when we finalize all the modules, would be nice to have an e2e test to export / import genesis data kind of like [TestAppImportExport](https://github.com/cosmos/cosmos-sdk/blob/698627016d64dbf33bea01837364c5a6e55a1ee3/simapp/sim_test.go#L107) from the sdk

app/app.go
SebastianElvis: Hmm I think this is because the `btcstkconsumer` module does not have any logic in Begin/EndBlocker. But doesn't bother adding them here

x/btcstkconsumer/types/genesis.go
```diff
@@ -10,5 +15,49 @@ func DefaultGenesis() *GenesisState {
...
+			return fmt.Errorf("finality provider consumer is not registered. Consumer id : %s, BTC pk: %s", fp.ConsumerId, fp.BtcPk.MarshalHex())
+		}
+	}
+
 	return gs.Params.Validate()
 }
+
+// validateConsumers validates the consumers
+// and returns a map with the consumer ids
+func validateConsumers(consumers []*ConsumerRegister) (map[string]bool, error) {
```
SebastianElvis, line 38: Perhaps 

```suggestion
func getConsumerMap(consumers []*ConsumerRegister) (map[string]struct{}, error) {
```

SebastianElvis (approved):
Thanks!

SebastianElvis:
Btw there is a GH issue https://github.com/babylonlabs-io/babylon/issues/426 for genesis import/export logics of btcstkconsumer and zoneconcierge

PR #837, opened 2025-04-23
chore: bump babylon version to v2 and add upgrade handler

Vvaradinov:
In preparation for the `v2` upgrade.


app/upgrades/v2/upgrades.go
```diff
@@ -0,0 +1,35 @@
...
+var Upgrade = upgrades.Upgrade{
+	UpgradeName:          UpgradeName,
+	CreateUpgradeHandler: CreateUpgradeHandler,
+	StoreUpgrades: store.StoreUpgrades{
+		Added:   []string{},
+		Deleted: []string{},
+	},
+}
+
+func CreateUpgradeHandler(mm *module.Manager, configurator module.Configurator, keepers *keepers.AppKeepers) upgradetypes.UpgradeHandler {
```
RafilxTenfen, line 25: for this upgrade we don't have to add the `x/btcstkconsumer` than, right?

test/e2e/btc_staking_e2e_test.go
```diff
@@ -625,7 +625,7 @@ func (s *BTCStakingTestSuite) Test8BTCDelegationFeeGrantTyped() {
 
 	// tries to create a send transaction putting the freegranter as feepayer, it should FAIL
 	// since we only gave grant for BTC delegation msgs.
-	// TODO: Uncomment the next lines when issue: https://github.com/babylonlabs-io/babylon/issues/693
+	// TODO: Uncomment the next lines when issue: https://github.com/babylonlabs-io/babylon/v2/issues/693
```
RafilxTenfen, line 628:
```suggestion
	// TODO: Uncomment the next lines when issue: https://github.com/babylonlabs-io/babylon/issues/693
```

KonradStaniec (approved):

RafilxTenfen (changes requested):
running `make proto-all` generates files which are not being removed, might need to update `proto/scripts/protocgen.sh`

KonradStaniec:
hmm good catch with protos @RafilxTenfen.

It seems with those changes all protos are generated within `v2`. I would say we would like to increment node version to v2 but protos should stay within v1 paths. (similarly to osmosis https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/tokenfactory/v1beta1/tx.proto)

KonradStaniec:
oh I think every proto needs regeneration (`make proto-all`) due to change in paths for custom types we have. Example:
```
message BTCHeaderInfo {
  bytes header = 1
      [ (gogoproto.customtype) =
            "github.com/babylonlabs-io/babylon/v2/types.BTCHeaderBytes" ];
  bytes hash = 2
      [ (gogoproto.customtype) =
            "github.com/babylonlabs-io/babylon/v2/types.BTCHeaderHashBytes" ];
  uint32 height = 3;
  bytes work = 4
      [ (gogoproto.customtype) = "cosmossdk.io/math.Uint" ];
}
```

now contains `github.com/babylonlabs-io/babylon/v2/types.BTCHeaderHashBytes` 


This is less than ideal imo, though I do not see easy way out of this 🤔  . I imagine this is not breaking as `Marshal` and `Unmarshal` methods of custom types did not change.

Vvaradinov:
@RafilxTenfen updated the protocgen script. Will add a tesst for v1 to v2 in a separate PR.

Vvaradinov:
@KonradStaniec protos are now regenerated correctly with the script.

x/btccheckpoint/types/genesis.pb.go
```diff
@@ -218,33 +218,33 @@ func init() {
 }
 
 var fileDescriptor_9776220697c13f63 = []byte{
```
RafilxTenfen, line 220: this file descriptor change does it matter for struct parsing? like will v1 babylon structures be able to be parsed correctly with this change `Babylon v2 `

RafilxTenfen (approved):
LGTM, great job :rocket: 

> Reminder, from now on, backports to v1 will have different imports :exploding_head:

x/btccheckpoint/types/genesis.pb.go
Vvaradinov: It shouldn't matter, it's just reflecting the addition of `v2` in the paths.

PR #840, opened 2025-04-24
chore(x/checkpointing): Add import/export genesis logic

GAtom22:


proto/babylon/checkpointing/v1/genesis.proto
```diff
@@ -3,13 +3,20 @@ package babylon.checkpointing.v1;
...
 option go_package = "github.com/babylonlabs-io/babylon/v2/x/checkpointing/types";
 
 // GenesisState defines the checkpointing module's genesis state.
 message GenesisState {
   // genesis_keys defines the public keys for the genesis validators
   repeated GenesisKey genesis_keys = 1;
+  // validator sets per epoch
+  repeated ValidatorSetEntry validator_sets = 2;
+  // checkpoints are all the raw checkpoints with meta
+  repeated RawCheckpointWithMeta checkpoints = 3;
```
RafilxTenfen, line 17: what about `ConflictingCheckpoint` ?

x/checkpointing/keeper/genesis_bls.go
RafilxTenfen, line 1: :100:

proto/babylon/checkpointing/v1/genesis.proto
GAtom22: in case it's set to true, [will halt the chain](https://github.com/babylonlabs-io/babylon/blob/1a3c7cb1b81380411e5ed158d6877e455f50a2cb/x/checkpointing/abci.go#L32). Not sure it makes sense to add it

RafilxTenfen: kkkk good point :sweat_smile:

RafilxTenfen (approved):
LGTM :tada:

PR #846, opened 2025-04-28
feat: Add tokenfactory module

Vvaradinov:
Wire up tokenfactory module in `app.go`. Seems no other migration is needed in the upgrade.


go.mod
```diff
@@ -272,11 +272,14 @@ require (
...
 replace (
+
+	// core v0.12 was tagged wrong
+	cosmossdk.io/core => cosmossdk.io/core v0.11.0
 	// use cosmos fork of keyring
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
-
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
+	github.com/strangelove-ventures/tokenfactory => github.com/babylonlabs-io/tokenfactory v0.50.3-wasmvm2
```
RafilxTenfen, line 282: Why don't we use our import directly in the code ?

SebastianElvis: i think we decided to have our fork of it so that we have control over its next versions

KonradStaniec (commented):
hmm this is weird upgrade is not needed i.e we are adding whole new keeper and module, I imagine we should at least add store in upgrade.


So our v2 upgrade would look like:

```go
const UpgradeName = "v2"

var Upgrade = upgrades.Upgrade{
	UpgradeName:          UpgradeName,
		Added: []string{
			tokenfactorytypes.StoreKey,
		},
		Deleted: []string{},
	},
}

```

Vvaradinov:
@KonradStaniec ahh my bad missed that one but otherwise seems like no other changes in upgrade need to be made. Was referencing Osmosis v9 [upgrade](https://github.com/osmosis-labs/osmosis/blob/e14ace31b7ba46be3d519966fb8563127534b245/app/upgrades/v9/upgrades.go)

app/keepers/keepers.go
```diff
@@ -657,6 +685,7 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
 	// whole usage of params module
 	paramsKeeper.Subspace(ibcexported.ModuleName)
 	paramsKeeper.Subspace(ibctransfertypes.ModuleName)
+	paramsKeeper.Subspace(tokenfactorytypes.ModuleName)
```
KonradStaniec, line 688: do you think this line is needed ? It seems the current version of token factory stores the params inside the module itself - https://github.com/babylonlabs-io/tokenfactory/blob/main/x/tokenfactory/keeper/params.go ?

app/keepers/keepers.go
```diff
@@ -84,8 +84,18 @@ import (
 	"github.com/babylonlabs-io/babylon/v2/x/zoneconcierge"
 	zckeeper "github.com/babylonlabs-io/babylon/v2/x/zoneconcierge/keeper"
 	zctypes "github.com/babylonlabs-io/babylon/v2/x/zoneconcierge/types"
+	tokenfactorykeeper "github.com/strangelove-ventures/tokenfactory/x/tokenfactory/keeper"
+	tokenfactorytypes "github.com/strangelove-ventures/tokenfactory/x/tokenfactory/types"
 )
 
+var tokenFactoryCapabilities = []string{
```
KonradStaniec, line 91: maybe lets add some doc explaining why those capabilites were chosen ? if those are defaults from some other docs we can add link to this doc here

go.mod
RafilxTenfen: yeap, but why not use our import in cases
```go
tokenfactorykeeper "github.com/strangelove-ventures/tokenfactory/x/tokenfactory/keeper"
	tokenfactorytypes "github.com/strangelove-ventures/tokenfactory/x/tokenfactory/types"
```

Vvaradinov: It still needs to reference the main repo included in the require.

app/keepers/keepers.go
KonradStaniec: ahh weird 🤔 then lets leave it 👍

KonradStaniec (approved):

RafilxTenfen (approved):
👍

Issue #158, opened 2024-10-10
Clean up Babylon make file


KonradStaniec:
Current make file -  https://github.com/babylonlabs-io/babylon/blob/main/Makefile, is an ancient copy of either Osmosis or Gaia chain. Most of the commands do not work and are undocumented.

This issue is about cleaning it up:
- leave only useful commands and scripts
- documents the

PR #851, opened 2025-04-29
[backport] chore: upgrade the make file: release and build (#841)

samricotta:
closes partly: https://github.com/babylonlabs-io/babylon/issues/158

Make file needed a bit of a makeover and some organisation


scripts/makefile/protobuf.mk
```diff
@@ -0,0 +1,40 @@
+
+###############################################################################
+###                                Protobuf                                 ###
+###############################################################################
+
+protoVer=0.14.0
```
KonradStaniec, line 6: this is duplicated in 3 make files. 

Maybe as future improvement we can have version defined in one separate file and then used in those make files ? 

I worry, in the future we will upgrade those versions and forgot to do it in all places.

KonradStaniec (approved):

PR #860, opened 2025-04-30
fix: ensure active fps have signing info after proposal

gitferry:
It is possible that some inactive fps become active after executing the finality resumption proposal. However, current implementation does not initiate signing info for those fps, which would cause panicking when handling liveness.

This PR fixed the issue by ensuring every active finality provider has signing info in the finality resumption proposal.

### DIFF
--- a/x/finality/keeper/gov.go
+++ b/x/finality/keeper/gov.go
@@ -31,13 +31,13 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
 	}
 
 	// jail the given finality providers
-	fpPks := make([]*bbntypes.BIP340PubKey, 0, len(fpPksHex))
+	fpPksToJail := make(map[string]struct{}, len(fpPksHex))
 	for _, fpPkHex := range fpPksHex {
 		fpPk, err := bbntypes.NewBIP340PubKeyFromHex(fpPkHex)
 		if err != nil {
 			return fmt.Errorf("invalid finality provider public key %s: %w", fpPkHex, err)
 		}
-		fpPks = append(fpPks, fpPk)
+		fpPksToJail[fpPkHex] = struct{}{}
 
 		_, voted := voters[fpPkHex]
 		if voted {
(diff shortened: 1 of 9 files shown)


x/finality/keeper/gov.go
```diff
@@ -117,5 +126,29 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
 		k.SetVotingPowerDistCache(ctx, h, distCache)
 	}
 
+	// it is possible that some inactive fps become active after the proposal
+	// therefore, we need to ensure every active finality provider has signing info
+	for pk, dc := range distCache.GetActiveFinalityProviderSet() {
```
KonradStaniec, line 131: doesn't this fix work only for last `distCache` we iterated in the above loop i.e after above loop `distCache` pointer will point to last modified `VotingPowerDistCache`

This means that if some finality providers were activated in the intermediate distcaches, problem will persist for them.

gitferry: Ah good point. We probably need to accumulate all the active fps from halting height to current height

KonradStaniec: hmm I do not think so  🤔 we evaluate liveness after  `FinalitySigTimeout` i.e
```go
		heightToExamine := sdk.UnwrapSDKContext(ctx).HeaderInfo().Height - k.GetParams(ctx).FinalitySigTimeout

		if heightToExamine >= 1 {
			k.HandleLiveness(ctx, heightToExamine)

```

gitferry: You are right. In a case where an fp becomes active at `current_height -1` and then inactive at `current_height`, we will miss it. Maybe we do a conservative fix by accumulating all the active fps and ensure they have signing info

gitferry: Done in [853cd05](https://github.com/babylonlabs-io/babylon/pull/860/commits/853cd05346869cae1cb038c2a163349f7f0673bf)

x/finality/keeper/gov.go
```diff
@@ -95,9 +103,15 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
...
+	var distCache *ftypes.VotingPowerDistCache
+	allActiveFps := make(map[string]*ftypes.FinalityProviderDistInfo)
 	for h := uint64(haltingHeight); h <= uint64(currentHeight); h++ {
-		distCache := k.GetVotingPowerDistCache(ctx, h)
+		distCache = k.GetVotingPowerDistCache(ctx, h)
 		activeFps := distCache.GetActiveFinalityProviderSet()
+		// add active fps to the all active fps set if it does not exist
+		for pk, fp := range activeFps {
+			allActiveFps[pk] = fp
+		}
```
RafilxTenfen, line 114: :thinking: This `activeFps` is not being used
```suggestion

```

RafilxTenfen: so we don't get the active fps from the latest `*ftypes.VotingPowerDistCache`, but should iterate over the `activeFps` map to verify the signing info

x/finality/keeper/gov.go
RafilxTenfen: ahh okay, here you are going to use `allActiveFps`, I think it is missing to replace 

```go
	for pk, dc := range distCache.GetActiveFinalityProviderSet() {
```

by

```go
	for pk, dc := range allActiveFps {
```

x/finality/keeper/gov.go
gitferry: Oops. The change is not pushed correctly

x/finality/keeper/gov.go
gitferry: Sorry, just updated the commit hash. The local change was not pushed correctly

x/finality/keeper/gov.go
```diff
@@ -95,8 +103,10 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
 	}
 
 	// set the all the given finality providers voting power to 0
+	var distCache *ftypes.VotingPowerDistCache
```
RafilxTenfen, line 106: not used anymore
```suggestion
```

x/finality/keeper/gov.go
```diff
@@ -95,8 +103,10 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
...
+	allActiveFps := make(map[string]*ftypes.FinalityProviderDistInfo)
 	for h := uint64(haltingHeight); h <= uint64(currentHeight); h++ {
-		distCache := k.GetVotingPowerDistCache(ctx, h)
+		distCache = k.GetVotingPowerDistCache(ctx, h)
```
RafilxTenfen, line 109:
```suggestion
		distCache := k.GetVotingPowerDistCache(ctx, h)
```

RafilxTenfen: done at https://github.com/babylonlabs-io/babylon/pull/860/commits/7efa78fe8e1bf1a042b26de22c0e8f4a60337b47

x/finality/keeper/gov.go
```diff
@@ -113,9 +120,38 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
...
+			allActiveFps[pk] = fp
+		}
+
 		// set the voting power distribution cache of the current height
 		k.SetVotingPowerDistCache(ctx, h, distCache)
 	}
 
+	// it is possible that some inactive fps become active after the proposal
+	// therefore, we need to ensure every active finality provider has signing info
+	for pk, dc := range allActiveFps {
```
RafilxTenfen, line 134: @gitferry I had modified the logic here to avoid error equal checking, but let me know your thoughts I can rollback the changes

gitferry: Looks good to me! Thank you for taking this and adding more tests

x/finality/keeper/gov.go
```diff
@@ -113,9 +120,38 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
...
+		hasSigningInfo, err := k.FinalityProviderSigningTracker.Has(ctx, dc.BtcPk.MustMarshal())
+		if err != nil {
+			return fmt.Errorf("failed to get signing info from tracker for fp %s - %w", pk, err)
+		}
+
+		if hasSigningInfo {
+			continue
+		}
+
+		signingInfo := ftypes.NewFinalityProviderSigningInfo(
```
KonradStaniec, line 144: adding signing info is not only think that happens after fp is activated. Other other thing that happens is emitting events - https://github.com/babylonlabs-io/babylon/blob/db20d092d8fd1f3a23133d7d90341ad6b433777f/x/finality/keeper/power_dist_change.go#L97.

Do you think we should do it here to avoid breaking anybody depending on events  ? @gitferry @RafilxTenfen 

Also maybe we can double check if there aren't other missing things which happens with each fp actiavation / deactivation in the normal path.

RafilxTenfen: good catch, we should emit events

RafilxTenfen: done at https://github.com/babylonlabs-io/babylon/pull/860/commits/a5ef731ca79a45b9a4795e6abda9d86c2448d320

x/finality/keeper/gov.go
```diff
@@ -94,27 +100,39 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
...
+				k.SetVotingPower(ctx, fp.BtcPk.MustMarshal(), blkHeight, 0)
 			}
+
+			// add this finality provider to the new cache if it has voting power
+			newDc.AddFinalityProviderDistInfo(&fp)
 		}
 
-		distCache.ApplyActiveFinalityProviders(params.MaxActiveFinalityProviders)
+		// sets the total vp in the new dc
+		newDc.ApplyActiveFinalityProviders(dc.NumActiveFps)
```
KonradStaniec, line 129: why changing `params.MaxActiveFinalityProviders` to `dc.NumActiveFps` ? is this intentional to keep amount of active fp's at the same level as in old dc ?

KonradStaniec (commented):
In general looks good, Only one question

x/finality/keeper/gov.go
RafilxTenfen: yes, as it might be different param at the block height of that vp distribution cache

RafilxTenfen (approved):
LGTM

KonradStaniec (approved):

PR #862, opened 2025-05-01
[backport] feat: Wire up ibc callbacks (#847)

KonradStaniec:
Wiring based on :
https://github.com/cosmos/ibc-go/blob/modules/apps/callbacks/v0.2.0%2Bibc-go-v8.0/modules/apps/callbacks/testing/simapp/app.go#L511


go.mod
```diff
@@ -67,6 +67,8 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20241202173237-19429a94021a
 )
 
+require github.com/gogo/status v1.1.0 // indirect
```
RafilxTenfen, line 70: weird that it got split :thinking:

KonradStaniec: yup bit weird, but this is vanilla result of `go mod tidy` 🤔

PR #869, opened 2025-05-02
F/rename old cz

maurolacy:
Closes the babylon side of https://github.com/babylonlabs-io/babylon-contract/issues/132.


proto/babylon/btcstkconsumer/v1/tx.proto
```diff
@@ -16,7 +16,7 @@ service Msg {
   // UpdateParams defines a (governance) operation for updating the module
   // parameters. The authority defaults to the x/gov module account.
   rpc UpdateParams(MsgUpdateParams) returns (MsgUpdateParamsResponse);
-  // RegisterConsumer registers a CZ consumer to the btcstkconsumer module.
+  // RegisterConsumer registers a BSN consumer to the btcstkconsumer module.
   rpc RegisterConsumer(MsgRegisterConsumer) returns (MsgRegisterConsumerResponse);
```
SebastianElvis, line 20:
```suggestion
  // RegisterConsumer registers a consumer to the btcstkconsumer module.
  rpc RegisterConsumer(MsgRegisterConsumer) returns (MsgRegisterConsumerResponse);
```

how about simply say consumer rather than BSN? BSN is more like a marketing term imo and might raise confusion

same comment on other BSN occurences

proto/babylon/zoneconcierge/v1/zoneconcierge.proto
```diff
@@ -131,9 +131,9 @@ message ProofFinalizedChainInfo {
     The following fields include proofs that attest the chain info is
     BTC-finalised
   */
-  // proof_cz_header_in_epoch is the proof that the CZ header is timestamped
+  // proof_bsn_header_in_epoch is the proof that the BSN's header is timestamped
   // within a certain epoch
-  tendermint.crypto.ProofOps proof_cz_header_in_epoch = 1;
+  tendermint.crypto.ProofOps proof_bsn_header_in_epoch = 1;
```
SebastianElvis, line 136: same here, `proof_consumer_header_in_epoch` is better

Also this is going to be a breaking change :thinking:

SebastianElvis (commented):
Finally time for getting rid of this CZ thing :+1:

proto/babylon/btcstkconsumer/v1/tx.proto
maurolacy: Sure! I was tempted to do that, but not sure it'll be accepted.

proto/babylon/zoneconcierge/v1/zoneconcierge.proto
maurolacy: Yes, I know... the option here is to keep the old field name, just for compatibility reasons.

test/e2e/scripts/rly_bootstrap.sh
```diff
@@ -65,7 +65,7 @@ rly -d --home $RELAYER_CONF_DIR keys restore bbn-b val01-bbn-b "$BBN_B_E2E_VAL_M
 sleep 3
 
 # Start Cosmos relayer
-echo "Creating IBC light clients, connection, and channel between the two BSNs"
+echo "Creating IBC light clients, connection, and channel between the two networks"
```
gusin13, line 68:
```suggestion
echo "Creating IBC light clients, connection, and channel between Babylon and Consumer"
```

docs/architecture.md
```diff
@@ -199,11 +199,11 @@ submitting necessary signatures.
 
 The IBC Relayer maintains the
 [IBC protocol](https://cosmos.network/ibc/) connection
-between Babylon and other Babylon Secured Networks (BSNs).
-It is responsible for updating the BSN's light client
+between Babylon and other Cosmos Consumers.
```
gusin13, line 202: relayer could be b/w any chains which supports IBC no? (could be cosmos or non cosmos)

x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
```diff
@@ -122,12 +122,12 @@ func (k Keeper) createBTCTimestamp(
 	epochNum := finalizedInfo.EpochInfo.EpochNumber
 	epochChainInfo, err := k.GetEpochChainInfo(ctx, consumerID, epochNum)
 	if err == nil {
-		// if there is a BSN header checkpointed in this finalised epoch,
-		// add this BSN header and corresponding proofs to the BTC timestamp
+		// if there is a Consumer header checkpointed in this finalised epoch,
+		// add this Consumer header and corresponding proofs to the BTC timestamp
```
gusin13, line 126: i was thinking if "Consumer" should be lower case "consumer" but on second thought seems like its ok to have upper like you did, as it depicts a separate entity not sure

gusin13 (approved):
few minor comments, overall LGTM ✅

proto/babylon/zoneconcierge/v1/zoneconcierge.proto
SebastianElvis: sure, but maybe leave it as a TODO?

x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
maurolacy: Yes, I think using upper case is fine / clearer.

proto/babylon/zoneconcierge/v1/zoneconcierge.proto
maurolacy: Then I would say let's just do this change now, as in the future it'll be worse.

proto/buf.lock
```diff
@@ -19,8 +19,8 @@ deps:
   - remote: buf.build
     owner: cosmos
     repository: ibc
-    commit: d36f8a9d78034378bae26a8ebfd291cb
-    digest: shake256:12816035a3657aaf6edf901648033789b1c84bd5588213818fdeb8a7f2652a3ce4b916254b16c6131bbb985e4b5eb0fe3872a7213668ecf54d2b04b760e9c120
+    commit: 709f9ae1d3464a058fe7a2687ad9a229
+    digest: shake256:56b9c28319fc79a924c95421ac6005adc79621fc31707be5459d3d489bc4f5b52bebf2d58e473ceb38e4b2d551bce3aa6bcb5fad9bab81488e9de7edc996a69f
```
SebastianElvis, line 23: Why bump this dependency?

SebastianElvis (approved):

proto/buf.lock
maurolacy: `proto-gen` did that I guess.

PR #850, opened 2025-04-29
feat: BLS improvement including permission and and verification cmd

wnjoon:
This PR includes the following changes:

## 1. Set `bls_key.json` file permissions to read-only

- The file permissions of the generated BLS key (`bls_key.json`) and the password file(optional) are set to 400.
- This logic has been added to the `Save()` method in `bls.go`.

## 2. Add a command to verify whether the registered BLS key matches the local bls_key.json

- This command compares the validator’s BLS key stored in the checkpointing module with the key in the local `bls_key.json`, based on a given validator address.
- The checkpointing module uses the latest epoch value obtained from the epoch module.
- Password loading uses the same flags as `create-bls-key` and `show-bls-key`.
- [Test scenario from the comment](https://github.com/babylonlabs-io/babylon/pull/850#issuecomment-2840779588).

```
$ babylond verify-validator-bls-key {validator_address} [flags]
```

PR #871, opened 2025-05-03
feat: create update-bls-password cmd

wnjoon:
This PR includes the following changes:

## Update the password encryption of the BLS key

- The existing password must be provided either via an interactive prompt or through the environment variable. For the security issue, existing password cannot be obtained by file.
- If previous version of BLS key does not have a password, enter empty value into prompt.
- The new password is managed using the same set of flags that are used when initially generating a BLS key. 
- Users are asked to process updating password of BLS since `bls_key.json` is overwritten with new password. 
- The `BABYLON_BLS_PASSWORD` environment variable is used to load the existing password. After running `update-bls-password`, users should update this variable to reflect the new password.
- Since both `bls_key.json` and its associated password file are set to read-only in [#850](https://github.com/babylonlabs-io/babylon/pull/850), they are first removed and then recreated during the update process.
- A backup of `bls_key.json` is made during update-bls-password to protect against potential errors. After `update-bls-password` completes successfully, the backup of `bls_key.json` is removed.

```
$ babylond update-bls-password [flags]
```

Flags:
- `--no-bls-password`: Set the new password to an empty value (remove the password).
- `--bls-password-file`: Path to the file where the new password will be stored.


app/signer/bls.go
```diff
@@ -507,3 +512,49 @@ func CreateBlsKey(homeDir string, password string, passwordFilePath string, cmd
 
 	return nil
 }
+
+// UpdateBlsPassword updates the password for a BLS key
+// Takes a password that was determined by the password determination logic.
+func UpdateBlsPassword(homeDir string, blsPrivKey bls12381.PrivateKey, password string, passwordFilePath string, cmd *cobra.Command) error {
```
RafilxTenfen, line 518: I thought this would remove the `passwordFilePath` from the Bls key structure
@gitferry

gitferry: I don't understand. The `passwordFilePath` is filled in `NewBls`, no?

gitferry (commented):
Thanks for your contribution. One general comment is that maybe it's cleaner to not support user to set the new password using env var as it seems not very straightforward.

Another comment is that after the bls key password is updated, a new key file will be generated. The two key files along with potential two password files should co-exist given that we do not allow overwriting the key file and the password file, right? Given that the key file and the password file share the same file name, they probably should be stored under a different directory? Would like to understand the consequence of a user mistakenly update the password under the same directory and how we prevent this. And, it's better to have this in the description of the command

wnjoon:
> One general comment is that maybe it's cleaner to not support user to set the new password using env var as it seems not very straightforward.

I considered the possibility of a scenario where passwords might need to be rotated periodically using a script. That’s why I initially planned to support two separate environment variables for the old and new passwords.

However, if such a scenario is not necessary, how about simplifying the approach by removing the `BABYLON_OLD_BLS_PASSWORD` variable, and instead using the existing `BABYLON_BLS_PASSWORD` only for loading the existing password?

After running the `update-bls-password` command, users can manually set the new password as an environment variable if needed.

> Another comment is that after the bls key password is updated, a new key file will be generated. The two key files along with potential two password files should co-exist given that we do not allow overwriting the key file and the password file, right? Given that the key file and the password file share the same file name, they probably should be stored under a different directory? Would like to understand the consequence of a user mistakenly update the password under the same directory and how we prevent this. And, it's better to have this in the description of the command

As you mentioned, if files are generated in the same directory, any existing files (`bls_key.json`, `password.txt`) would likely be overwritten by default.

What do you think about introducing a `--force-overwrite` flag?

- If the output path is the same as an existing file, and `--force-overwrite` is not set, the command will prompt the user with a warning and ask whether to overwrite the files (y/n).
- If the flag is set, the existing files will be overwritten without any confirmation.
- If users wish to keep a backup or maintain a history of previous files, they will need to manage the file paths manually.

Let me know your thoughts on this approach.

cmd/babylond/cmd/update_bls_password.go
```diff
@@ -0,0 +1,98 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to get home directory: %w", err)
+			}
+
+			cmd.Println("\n ⚠️ IMPORTANT: To update the password of BLS key, you need to provide the old password. ⚠️")
+			cmd.Println("Due to security issues, older versions only accept passwords via environment variables or prompts.")
+			cmd.Println("If the previous bls key has no password, enter a blank string value at the prompt.")
+			cmd.Println("If this command uses both old and new passwords from environment variables, set different values for each:")
+			cmd.Println("- old password: BABYLON_OLD_BLS_PASSWORD")
+			cmd.Println("- new password: BABYLON_BLS_PASSWORD")
```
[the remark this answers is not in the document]
wnjoon: > I considered the possibility of a scenario where passwords might need to be rotated periodically using a script. 

[My opinions](https://github.com/babylonlabs-io/babylon/pull/871#issuecomment-2858504242)

gitferry:
Sorry for the delay.

> However, if such a scenario is not necessary, how about simplifying the approach by removing the BABYLON_OLD_BLS_PASSWORD variable, and instead using the existing BABYLON_BLS_PASSWORD only for loading the existing password?
> After running the update-bls-password command, users can manually set the new password as an environment variable if needed.

This sounds more straightforward.

> As you mentioned, if files are generated in the same directory, any existing files (bls_key.json, password.txt) would likely be overwritten by default.
> What do you think about introducing a --force-overwrite flag?
> If the output path is the same as an existing file, and --force-overwrite is not set, the command will prompt the user with a warning and ask whether to overwrite the files (y/n).
If the flag is set, the existing files will be overwritten without any confirmation.
If users wish to keep a backup or maintain a history of previous files, they will need to manage the file paths manually.
Let me know your thoughts on this approach.

I think it's very risky to allow overwriting of existing key file in any way. Introducing `--force-overwrite` does not mitigate the risk. In that sense, I propose the follows:
1. We need to ensure that the existing key will not be overwritten in any way. Does this [PR](https://github.com/babylonlabs-io/babylon/pull/850) ensure this?
2. If the new key file path or the password file path is the same as the existing one, we should reject the update password request and ask the user to specify a different path
3. After password is updated, two key files might co-exist, and the user should be able to choose either one with proper flags. It's up to the user to delete the old one manually.

@wnjoon wdyt?

wnjoon:
@gitferry 

Thank you for the comment. I realized I also overlooked an important detail.

The `update-bls-password` command, as the name suggests, only updates the password used to encrypt/decrypt the `bls_key.json` file — the BLS key itself remains unchanged.

As a result, there is no change in the path or content of the BLS key file, and only the password file may be overwritten if the path is the same.

Therefore, regarding point 3 you mentioned:

> Two key files might co-exist, and the user should be able to choose either one with proper flags

— this situation will not occur.

Since the same BLS key is simply re-encrypted with a different password, the user must always use the updated password (set via the `update-bls-password` command) to interact with the key going forward.

gitferry:
> Since the same BLS key is simply re-encrypted with a different password, the user must always use the updated password (set via the update-bls-password command) to interact with the key going forward.

Thank you for the explanation. Seems I have missed something. If the password is changed, then the content of the encrypted BLS key in the `bls_key.json` should also be changed, right? Wouldn't it cause overwriting?

wnjoon:
@gitferry 

If you look at line 60 of `cmd/babylond/cmd/update_bls_password.go`, the BLS key is first decrypted using the old password, and then re-encrypted using the new password, preserving the same underlying key.

```go
blsPrivKey, err := appsigner.LoadBlsPrivKey(homeDir, oldPassword)
```

As a result, the `bls_key.json` file is overwritten, and while the encrypted data changes, the actual BLS key value itself remains the same.

So yes — the file is overwritten, but it’s still representing the same key, just secured with a new password.

gitferry:
@wnjoon I see. So in the PR, the `bls_key.json` file is always overwritten after update. Wouldn't the https://github.com/babylonlabs-io/babylon/pull/850 prevents overwriting?

In this case, maybe overwriting makes sense but still risky. Maybe we should send a warning that this cmd will overwrite the existing `bls_key.json` and the user should have a backup? This warning can be in an interactive way and requires the user to type [y/n] to proceed

wnjoon:
@gitferry 

Thanks for pointing this out! You’re right — the current implementation does overwrite the existing `bls_key.json`, which could contradict the intent in [#850](https://github.com/babylonlabs-io/babylon/pull/850).

While overwriting makes sense in the context of re-encrypting the same key with a new password, I agree it’s still a risky operation.

I’ll update the command to:

- Show a clear warning before overwriting
- Prompt the user interactively with [y/n] to confirm
- Suggest making a backup beforehand

Let me know if that sounds reasonable!

gitferry:
@wnjoon The plan sounds good to me! Although a bit unsure about whether overwriting is even possible due to https://github.com/babylonlabs-io/babylon/pull/850

wnjoon:
@gitferry 

Based on standard Unix file permissions, if `bls_key.json` is set to chmod 400, any attempt to overwrite it directly will result in a permission denied error, since the file lacks write permission.

However, deletion is still possible as long as the user has write access to the directory.

So, recommended behavior is:

1. Delete and recreate the `bls_key.json` file
2. Apply the same policy to the password file

Between changing file permissions (e.g., from chmod 400 to 600) and deleting/recreating the file, I recommend the latter for the following reasons:

- Security: There is no window where the file exists with elevated write permissions.
- Simplicity: It avoids the need to track and restore file permissions.
- Predictability: File deletion followed by creation is more consistent and less error-prone across systems.

gitferry:
@wnjoon Sounds good to me. Thanks!

wnjoon:
## Commit summary: [43126de](https://github.com/babylonlabs-io/babylon/pull/871/commits/43126de990b199201e6b8a8b2a93c7a25182b84f)

- Updated prompt messages for CLI execution
- Removed the `BABYLON_OLD_BLS_PASSWORD` environment variable
- `bls_key.json` is now deleted and recreated when running `update-bls-password` cmd (reflects behavior from PR [#850](https://github.com/babylonlabs-io/babylon/pull/850))
- Set read-only permissions (chmod 400) at the time of BLS file creation
   - In contrast to PR [#850](https://github.com/babylonlabs-io/babylon/pull/850), which first created the file with 600 and then changed it to 400, this commit creates the file directly with 400 permissions
   - This change will be integrated when merging into main
- Refined prompt message for BLS password input (removed redundant wording)

## Discussion

While working on this, I found one additional point that needs to be clarified regarding the password file behavior.

In both the `babylond init` and `create-bls-key` commands, if a file already exists at the path specified by the `--bls-password-file` flag, the contents of that file are used as the password, rather than generating a new one.

Therefore, I believe the `update-bls-password` command should follow the same behavior for consistency.

Here’s a summary of the proposed behavior for `update-bls-password`:

1. The existing password for the current `bls_key.json` is provided via environment variable or prompt — no change here.
2. If a file already exists at the `--bls-password-file` path, the contents of that file should be used as the new password.
   - In this case, no password update occurs if the file still contains the old password.
   - The user should be warned via prompt that the password file already exists and must be deleted or moved manually to proceed with the update.
3. If the file does not exist at the `--bls-password-file` path, a new password file is created at that location.
4. As previously discussed, the `bls_key.json` file will be deleted and recreated during the update process if it already exists.

The key point is:
- For the `--bls-password-file` flag, we should not automatically delete or overwrite an existing file.
- If the file exists, we must respect its contents and use it as the new password — this ensures consistent behavior with `create-bls-key` and `init`.

wnjoon:
Additionally, I noticed something unusual during testing.

After merging with main earlier, I continued working on the code — but encountered issues while running some of the tests (as shown below).
To stay focused on the current PR, I decided to reset to commit [4d9733d](https://github.com/babylonlabs-io/babylon/pull/871/commits/4d9733d81bde07c41c56f7ce7170576dfe1e99fe) and continue development from there.

Related Command (whole script is in [here](https://github.com/babylonlabs-io/babylon/pull/850#issuecomment-2840779588))
```sh
...
# Error occurred in here
$ babylond gentx validator1 1500000000000stake --chain-id test-chain-1 --keyring-backend test --fees 400ubbn
...
```

Error Message:
```sh
"failed to validate genesis state: duplicate denomination ubbn"
```

@gitferry @RafilxTenfen

app/signer/bls.go
```diff
@@ -507,3 +505,54 @@ func CreateBlsKey(homeDir string, password string, passwordFilePath string, cmd
...
+
+	// Check if BLS key already exists
+	if !cmtos.FileExists(blsKeyFile) {
+		return fmt.Errorf("BLS key does not exist at %s", blsKeyFile)
+	}
+
+	// Remove BLS key file
+	if err := os.Remove(blsKeyFile); err != nil {
+		return fmt.Errorf("failed to remove BLS key file: %w", err)
+	}
```
gitferry, line 522: Can we backup the key file before removing it, e.g., `bls_key.json.bk`? We can safely remove the backup after the new key file is created. If we don't backup it, we will lose the key if something happens before the new key file is created.

gitferry (commented):
Overall lgtm. One comment about backup the bls key before removing it

cmd/babylond/cmd/update_bls_password.go
```diff
@@ -0,0 +1,103 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to get old BLS password: %w", err)
+			}
+
+			// Unset environment variable for old version of password
+			// to prevent it from being used in the next step
+			// User should reset environment variable to the new password after updating
+			if err := os.Unsetenv(appsigner.BlsPasswordEnvVar); err != nil {
+				return fmt.Errorf("failed to unset BLS password environment variable: %w", err)
+			}
```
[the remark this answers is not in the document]
wnjoon: @gitferry 

Since we are not using separate environment variables for the old and new passwords,
if the environment variable is not unset at that point, the command may read the same value again when setting the new password which would result in the password not actually being updated.

app/signer/bls.go
wnjoon: fixed [7dffa02](https://github.com/babylonlabs-io/babylon/pull/871/commits/7dffa026c6c6809bfca04d9812dcec15b3608a9a)

wnjoon:
@gitferry @RafilxTenfen 

While reviewing the code, I noticed an area that could be improved, and I’d like to share my thoughts.

Not just in `update-bls-password`, but in most of the existing BLS-related commands, the workflow currently follows this order:

1. Prompt for the BLS password (via env var, file, or prompt)
2. Then, inside functions in `bls.go`, check whether the `bls_key.json` file exists

I believe this order could be improved by reversing these two steps. 
It’s better to check for the existence of `bls_key.json` at the command level (cmd directory) rather than inside `bls.go`. The internal check in `bls.go` can then be removed to simplify the logic.

For example, in commands like `show-bls-key` or `update-bls-password`, if the `bls_key.json` file doesn’t exist, the user is still asked to enter a password — only to then be told that the operation can’t proceed. This leads to a poor user experience.

In the case of `create-bls-key`, if the file already exists, the command should immediately inform the user that the file cannot be overwritten, and suggest either deleting it or backing it up to a different path before retrying.

What are your thoughts on this proposed adjustment?

cmd/babylond/cmd/update_bls_password.go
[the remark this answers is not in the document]
gitferry: Ah now I understand the issue is within `GetBlsKeyPassword` where env var is checked and returned if it has a value.

gitferry (approved):
LGTM! Thanks for all the discussion the changes. Great work! Will merge after getting another approval from @RafilxTenfen

gitferry:
@wnjoon Totally aggree with the UX improvement proposal. Feel free to create an issue about it and address it in a separate PR. Thank you so much for your continuous contribution on BLS keys!

wnjoon:
Commit Summary: [cb0bff9](https://github.com/babylonlabs-io/babylon/pull/871/commits/cb0bff9977361457f0ac3a2e56dff0806e7febf4)

- Remove `os.Chmod` while saving BLS key in `bls.go` and set file to read only directly.

wnjoon:
> @wnjoon Totally aggree with the UX improvement proposal. Feel free to create an issue about it and address it in a separate PR. Thank you so much for your continuous contribution on BLS keys!

@gitferry 
I've just created new issue #917 and will create new PR using main branch after this PR is merged to.

RafilxTenfen (approved):
LGTM, thanks @wnjoon

PR #877, opened 2025-05-06
fix: Reward Distribution Calculated at the babylon block to reward BTC

RafilxTenfen:
Fix the distribution of BTC rewards to be precise with the voting power of the BTC delegators accordingly to the block height that is being rewarded, not the latest state

- Add two new keys to store in the `x/incentive` keeper
  -  One tracks all the events that affect the reward tracker (activated BTC dels, unbond/withdraw) indexed by the Babylon block height in which they were processed
  - The other keeps track of the latest processed block height 
- Voting power distribution cache updates now add the events instead of directly updating the BTC reward tracker structures
- Reward tracker events are processed right before rewarding the BTC block, since the reward distribution of the BTC blocks are always sequential it was possible to make this arrangement without keeping track of the VP of each BTC delegator for each unprocessed babylon height


proto/babylon/incentive/events.proto
```diff
@@ -0,0 +1,55 @@
...
+
+import "cosmos_proto/cosmos.proto";
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/babylonlabs-io/babylon/v2/x/incentive/types";
+
+// EventsPowerUpdateAtHeight stores an slice of events that happened at
+// some specific block height, it is indexed by the block height.
+message EventsPowerUpdateAtHeight {
+  repeated EventPowerUpdate events = 1;
```
KonradStaniec, line 12: I guess the events are pretty small, so we could store even 1M of them at each height and not get OOM error when retrieving them from db?

x/incentive/keeper/reward_tracker_events.go
```diff
@@ -0,0 +1,156 @@
...
+		return nil, err
+	}
+
+	return &v, nil
+}
+
+// GetRewardTrackerEventLastProcessedHeight returns the latest processed height of the events tracker
+// Note: returns zero if not found.
+func (k Keeper) GetRewardTrackerEventLastProcessedHeight(ctx context.Context) (uint64, error) {
+	found, err := k.rewardTrackerEventsLastProcessedHeight.Has(ctx)
```
KonradStaniec, line 120: what will be `rewardTrackerEventsLastProcessedHeight` set to after upgrade ? I guess `0`, so we will need to iterate over a lot of block unnecesarily. 

Maybe we should set it to some high value in the v2 upgrade ?

x/incentive/keeper/reward_tracker_events.go
```diff
@@ -0,0 +1,156 @@
...
+// GetOrNewRewardTrackerEvent returns a new reward tracker if it doesn't exists for that block height
+func (k Keeper) GetOrNewRewardTrackerEvent(ctx context.Context, height uint64) (*types.EventsPowerUpdateAtHeight, error) {
+	found, err := k.rewardTrackerEvents.Has(ctx, height)
+	if err != nil {
+		return nil, err
+	}
+
+	if !found {
+		return &types.EventsPowerUpdateAtHeight{
+			Events: make([]*types.EventPowerUpdate, 0),
```
KonradStaniec, line 105: I guess this means that just after upgrade:
- few first blocks (the ones without events) will be rewarded against latest satte
- after some time we will switch to relaying on the events

proto/babylon/incentive/events.proto
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// EventPowerUpdate is an event that affects voting power distribution
+// of BTC rewards, it is indexed by the babylon block height which
+// they were processed at the finality voting power table.
+message EventPowerUpdate {
+  // ev is the event that affects voting power distribution
+  oneof ev {
+    // BtcActivated means a BTC delegation becomes active.
+    EventBTCDelegationActivated btc_activated = 1;
```
KonradStaniec, line 22: I remember RFC mentioned we will need also slashed event. Ultimately is it unnecessary?

KonradStaniec (commented):
great work ! it is really clean addition :) Some initial questions / comments.

For me the biggest uncertainty in this is how the system will behave after upgrade i.e switching to this approach on the running chain ? Do you think it is possible to add such e2e upgrade test or it will be to unrelaiable / to long running to make sense ?

proto/babylon/incentive/events.proto
RafilxTenfen: I believe so, but I didn't make any stress tests on this
The thing is, we can't limit to handle a portion each block, otherwise we would be distributing the rewards not according to the Babylon block height again. But we could limit the amount by using an iterator if you prefer, then the store would look a bit different

`(blkHeight, idx) => event`

x/incentive/keeper/reward_tracker_events.go
RafilxTenfen: correct

proto/babylon/incentive/events.proto
RafilxTenfen: I thought it was too much change for now, since the slashing of the rewards is also not wired at v1.1

RafilxTenfen:
> For me the biggest uncertainty in this is how the system will behave after upgrade i.e switching to this approach on the running chain ? Do you think it is possible to add such e2e upgrade test or it will be to unrelaiable / to long running to make sense ?

Yes, lets add the e2e test with upgrade to v2 during the backport of the PR, as in `main` branch there is a few modules created that will not be added to v2

proto/babylon/incentive/events.proto
KonradStaniec: If events are small enough lets keep it as it is, iterators are slow as hell and should only be used as last resort (if we are sure data won't fit into memory)

proto/babylon/incentive/events.proto
```diff
@@ -0,0 +1,55 @@
+syntax = "proto3";
+package babylon.incentive;
+
+import "cosmos_proto/cosmos.proto";
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/babylonlabs-io/babylon/v2/x/incentive/types";
+
+// EventsPowerUpdateAtHeight stores an slice of events that happened at
+// some specific block height, it is indexed by the block height.
```
gitferry, line 10:
```suggestion
// EventsPowerUpdateAtHeight stores a slice of events happened at
// some specific block height, indexed by the block height.
```

proto/babylon/incentive/events.proto
```diff
@@ -0,0 +1,55 @@
...
+
+// EventPowerUpdate is an event that affects voting power distribution
+// of BTC rewards, it is indexed by the babylon block height which
+// they were processed at the finality voting power table.
+message EventPowerUpdate {
+  // ev is the event that affects voting power distribution
+  oneof ev {
+    // BtcActivated means a BTC delegation becomes active.
+    EventBTCDelegationActivated btc_activated = 1;
+    // BtcUnbonded means a BTC delegation becomes unbonded or withdraw.
```
gitferry, line 23:
```suggestion
    // BtcUnbonded means a BTC delegation becomes unbonded
```
By `withdraw`, did you mean BTC delegation expiry? It's also unbonded from Babylon' pov

proto/babylon/incentive/events.proto
```diff
@@ -0,0 +1,55 @@
...
+  string btc_del_addr = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // TotalSat the number of satoshis delegated
+  bytes total_sat = 3 [
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false
+  ];
+}
+
+// EventBTCDelegationUnbonded event that unbonded or withdraw a BTC delgation.
```
gitferry, line 43:
```suggestion
// EventBTCDelegationUnbonded event that unbonded a BTC delegation.
```

app/upgrades/v2/upgrades.go
```diff
@@ -33,6 +35,21 @@ func CreateUpgradeHandler(mm *module.Manager, configurator module.Configurator,
...
+			return nil, err
+		}
+
 		return migrations, nil
 	}
 }
+
+// UpdateRewardTrackerEventLastProcessedHeight sets the current block height -1 to the reward tracker
+// so when a new BTC block is rewarded it doesn't need to go through all the missing block heights
+// since block 0.
```
gitferry, line 50:
```suggestion
// UpdateRewardTrackerEventLastProcessedHeight sets the current block height - 1 to the reward tracker
// so that BTC reward distribution starts from the height at which the upgrade happens
```

Am I understanding it correctly?

gitferry (approved):
Lgtm! Great work!

Lazar955 (approved):
Nice work!

KonradStaniec (approved):
LGTM! great work!

PR #878, opened 2025-05-07
chore: move goreleaser job to release workflow

maiquanghiep:
When a release is created using the [release workflow](https://github.com/babylonlabs-io/babylon/blob/main/.github/workflows/release.yml), it doesn't trigger the gorelease workflow due to the limitation with `github.token` ([ref](https://github.com/actions/create-release/issues/71)).
This PR moves the goreleaser job into the release workflow, allowing it to be executed once the release is created


Lazar955 (approved):
utACKe

filippos47 (approved):
Do we still need the independent goreleaser workflow for the manually created releases?

maiquanghiep:
@filippos47 Yes I do think we need it. gorelease is still needed in case we create the release without using the workflow

PR #879, opened 2025-05-07
chore: Add e2e test for v2 upgrade

Vvaradinov:
Add v2 e2e upgrade

PR #880, opened 2025-05-07
[backport] chore: Add e2e test for v2 upgrade #879

Vvaradinov:
Backports e2e upgrade PR. Here the test should pass as there is no extra modules wired unlike main where the extra phase 3 modules added make the test fail.


KonradStaniec (approved):

RafilxTenfen (approved):
:tada:

PR #881, opened 2025-05-07
chore(x/mint): update gen logic

GAtom22:


proto/babylon/mint/v1/genesis.proto
```diff
@@ -1,12 +1,16 @@
...
 option go_package = "github.com/babylonlabs-io/babylon/v2/x/mint/types";
 
 // GenesisState defines the mint module's genesis state.
 message GenesisState {
-  reserved 1; // 1 was previously used for the `Minter` field.
-
-  // BondDenom is the denomination of the token that should be minted.
-  string bond_denom = 2;
+  // minter represents the mint state.
+  Minter minter = 1;
```
RafilxTenfen, line 11: this seems suspicius, are we going back to use minter structure?

GAtom22: It's the struct that is store in state.
Maybe @SebastianElvis has more context on the changes

RafilxTenfen: this minter is different than what was there before https://github.com/babylonlabs-io/babylon/pull/211/files#r1822021542

RafilxTenfen (approved):
LGTM

PR #897, opened 2025-05-12
feat: wire ICA and ICQ modules

GAtom22:


test/e2e/configurer/base.go
```diff
@@ -115,6 +115,84 @@ func (bc *baseConfigurer) RunIBCTransferChannel() error {
 	return nil
 }
 
+// CompleteIBCChannelHandshake completes the channel handshake in cases when ChanOpenInit was initiated
+// by some transaction that was previously executed on the chain. For example,
+// ICA MsgRegisterInterchainAccount will perform ChanOpenInit during its execution.
+func (bc *baseConfigurer) CompleteIBCChannelHandshake(
```
RafilxTenfen, line 121: :tada:

Vvaradinov (approved):
LGTM!

KonradStaniec (approved):

Lazar955 (approved):
lgtm!

app/upgrades/v2/upgrades.go
```diff
@@ -33,6 +38,18 @@ func CreateUpgradeHandler(mm *module.Manager, configurator module.Configurator,
 			return nil, err
 		}
 
+		// By default, ICQ allowed queries are empty. So no queries will be allowed until
```
KonradStaniec, line 41: Shouldn't we set the params for ICQ eitherway in the upgrade ? Otherwise the keeper will return zero value struct:
```
func (k Keeper) GetParams(ctx sdk.Context) types.Params {
	var p types.Params

	store := ctx.KVStore(k.storeKey)
	bz := store.Get(types.ParamsKey)
	if bz == nil {
		return p
	}

	k.cdc.MustUnmarshal(bz, &p)
	return p
}
which may be different than `DefaultParams` in the module ? 
```

KonradStaniec: the same question touches `icacontrollerkeeper.NewKeeper` , which default set param sets `ControllerEnabled` to `true` , but default zero value in golang is `false` (from what I remember 😅 )

KonradStaniec: ahh no if the params we do not set match whatever is produced by `func DefaultParams() Params` of each module then imo we are fine 👍 

Though I will double check on implementation level how it works as if params are not set modules usually return go zero value:
```
	var p types.Params
```

which by default set bools to `false` 🤔

PR #901, opened 2025-05-13
fix: btcstaking test fix

samricotta:
When merging the btcstaking test, there must have been a change to the test as it was being merged as it is now failing after merge although it wasnt before. Ive added nil to the test to handle this issue

### DIFF
--- a/test/replay/btcstaking_test.go
+++ b/test/replay/btcstaking_test.go
@@ -362,7 +362,7 @@ func TestSlashingandFinalizingBlocks(t *testing.T) {
 	scenario.InitScenario(6, 1) // 6 finality providers
 
 	numBlocksInTest := uint64(10)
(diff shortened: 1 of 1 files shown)


test/replay/btcstaking_test.go
```diff
@@ -362,7 +362,7 @@ func TestSlashingandFinalizingBlocks(t *testing.T) {
 	scenario.InitScenario(6, 1) // 6 finality providers
 
 	numBlocksInTest := uint64(10)
-	lastVotedBlkHeight := scenario.FinalityFinalizeBlocks(scenario.activationHeight, numBlocksInTest)
+	lastVotedBlkHeight := scenario.FinalityFinalizeBlocks(scenario.activationHeight, numBlocksInTest, nil)
```
KonradStaniec, line 365: you probably need to pass here `scenario.FpMapBtcPkHex()`

KonradStaniec (approved):

PR #911, opened 2025-05-14
chore: backport v2x 877 rwd fix

RafilxTenfen:
- Backport #877 
- Fix proto-gen and mocks
- Add BTC rewards check to upgrade e2e test


proto/babylon/incentive/rewards.proto
```diff
@@ -5,7 +5,7 @@ import "gogoproto/gogo.proto";
 import "cosmos/base/v1beta1/coin.proto";
 import "cosmos_proto/cosmos.proto";
 
-option go_package = "github.com/babylonlabs-io/babylon/x/incentive/types";
+option go_package = "github.com/babylonlabs-io/babylon/v2/x/incentive/types";
```
KonradStaniec, line 8: Oh I see some protos have this changed to v2, some not. Should we have it changed in all of them ? (if so maybe we can do that in separate pr)

PR #923, opened 2025-05-16
[backport] feat: create update-bls-password cmd (#871)

gitferry:
This PR includes the following changes:

## Update the password encryption of the BLS key

- The existing password must be provided either via an interactive prompt or through the environment variable. For the security issue, existing password cannot be obtained by file.
- If previous version of BLS key does not have a password, enter empty value into prompt.
- The new password is managed using the same set of flags that are used when initially generating a BLS key.
- Users are asked to process updating password of BLS since `bls_key.json` is overwritten with new password.
- The `BABYLON_BLS_PASSWORD` environment variable is used to load the existing password. After running `update-bls-password`, users should update this variable to reflect the new password.
- Since both `bls_key.json` and its associated password file are set to read-only in [#850](https://github.com/babylonlabs-io/babylon/pull/850), they are first removed and then recreated during the update process.
- A backup of `bls_key.json` is made during update-bls-password to protect against potential errors. After `update-bls-password` completes successfully, the backup of `bls_key.json` is removed.

```
$ babylond update-bls-password [flags]
```

Flags:
- `--no-bls-password`: Set the new password to an empty value (remove the password).
- `--bls-password-file`: Path to the file where the new password will be stored.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [#847](https://github.com/babylonlabs-io/babylon/pull/847) Add ibc callbacks to
 transfer stack
 - [#859](https://github.com/babylonlabs-io/babylon/pull/859) Update init/export genesis logic in `x/finality` module
+- [#871](https://github.com/babylonlabs-io/babylon/pull/871) create `update-bls-password` cmd
 - [#876](https://github.com/babylonlabs-io/babylon/pull/876) Add Packet Forwarding Middleware (PFM) module.
 - [#881](https://github.com/babylonlabs-io/babylon/pull/881) Update init/export genesis logic in `x/mint` module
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [#847](https://github.com/babylonlabs-io/babylon/pull/847) Add ibc callbacks to
 transfer stack
 - [#859](https://github.com/babylonlabs-io/babylon/pull/859) Update init/export genesis logic in `x/finality` module
+- [#871](https://github.com/babylonlabs-io/babylon/pull/871) create `update-bls-password` cmd
```
RafilxTenfen, line 54: we should put it under unreleased

RafilxTenfen: true, my bad :thinking:

RafilxTenfen (approved):
LGTM

PR #926, opened 2025-05-19
chore(gen): Update genesis initialization validations

GAtom22:
# Description

After testing the export and import genesis on a local chain, encountered these validations that should be updated

### DIFF
--- a/x/monitor/types/genesis.go
+++ b/x/monitor/types/genesis.go
@@ -33,17 +33,7 @@ func (gs GenesisState) Validate() error {
 	return nil
 }
 
-func (e EpochEndLightClient) Validate() error {
-	if e.BtcLightClientHeight == 0 {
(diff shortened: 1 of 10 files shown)


x/monitor/types/genesis.go
```diff
@@ -33,17 +33,7 @@ func (gs GenesisState) Validate() error {
 	return nil
 }
 
-func (e EpochEndLightClient) Validate() error {
-	if e.BtcLightClientHeight == 0 {
```
RafilxTenfen, line 37: why remove these checks ?

GAtom22: can happen that there's a epoch where there was no BTC lightclient height stored, so will default to 0

RafilxTenfen (approved):
LGTM

PR #928, opened 2025-05-19
[backport] chore(gen): Update genesis initialization validations (#926)

GAtom22:
# Description

After testing the export and import genesis on a local chain, encountered these validations that should be updated


KonradStaniec:
@GAtom22 given that:
- we won't need export/import functionality for now on release/v2.x (though we may need in the future, we will see)
- `release/v2.x` already have  `rc.0` tag which means it should only receive fixe
maybe lets not backport for now ?

GAtom22:
> @GAtom22 given that:
> 
> * we won't need export/import functionality for now on release/v2.x (though we may need in the future, we will see)
> * `release/v2.x` already have  `rc.0` tag which means it should only receive fixe
>   maybe lets not backport for now ?

Sure, will close it then

PR #992, opened 2025-05-29
fix: resume fp halt height

RafilxTenfen:
- Add checks for bad halting height in `HandleResumeFinalityProposal`
- Calculate total vp of distribution cache in the same loop


GAtom22 (approved):
LGTM! Great work @RafilxTenfen!! :rocket:

PR #1004, opened 2025-05-30
fix: use correct err type in MsgResumeFinalityProposal validate basic

RafilxTenfen:


GAtom22 (approved):
LGTM! Thanks @RafilxTenfen!!

PR #970, opened 2025-05-28
F/op e2e 3

maurolacy:
#946 follow-up. Create delegation, add covenant sigs.

PR #1015, opened 2025-06-02
F/op e2e 4

maurolacy:
#970 follup-up. Commit pub rand, submit finality sig. Closes most of #https://github.com/babylonlabs-io/pm/issues/374.


RafilxTenfen (approved):
a few suggestions, but looks good to me :tada:

PR #1040, opened 2025-06-03
chore(btcstkconsumer): rename ETH L2 to rollup 

gusin13:
ETH L2 was the old terminology, we should use `rollup` to be consistent


RafilxTenfen (approved):
Code changes look good to me, one note if there is a running chain with the module `btcstkconsumer` it will break just change the binary, it will require starting another chain from scratch due to proto changes

gusin13:
> Code changes look good to me, one note if there is a running chain with the module `btcstkconsumer` it will break just change the binary, it will require starting another chain from scratch due to proto changes

cc @liam-icheng-lai @filippos47 
v4 devnet is using this module, so for the next release will keep this in mind

gusin13:
> Code changes look good to me, one note if there is a running chain with the module `btcstkconsumer` it will break just change the binary, it will require starting another chain from scratch due to proto changes

cc @liam-icheng-lai @filippos47 
v4 devnet is using this module, so for the next release will need to keep this in mind

PR #1060, opened 2025-06-04
imp(finality): optimize PubRandCommit lookup

GAtom22:
# Description

Introduce a secondary index to enable efficient binary search-style lookups for PubRandCommit entries associated with a given finality provider public key (FpBtcPk). This improves read performance when searching for the commitment covering a specific height.

- [RFC](https://github.com/babylonlabs-io/pm/blob/main/rfc/rfc-038-optimize-pub-rand-lookup.md)
- [PR with ADR](https://github.com/babylonlabs-io/pm/pull/418)


GAtom22:
Benchmark tests results

```
go test -benchmem -run=^$ -bench ^BenchmarkCompareIndexVsReverse$ github.com/babylonlabs-io/babylon/v4/x/finality/keeper

goos: linux
goarch: amd64
pkg: github.com/babylonlabs-io/babylon/v4/x/finality/keeper
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkCompareIndexVsReverse/IndexPath_100-8         	  575005	      2400 ns/op	    2408 B/op	      16 allocs/op
BenchmarkCompareIndexVsReverse/ReversePath_100-8       	   57477	     22649 ns/op	    9873 B/op	      78 allocs/op
BenchmarkCompareIndexVsReverse/IndexPath_1000-8        	  173629	      6402 ns/op	    9704 B/op	      16 allocs/op
BenchmarkCompareIndexVsReverse/ReversePath_1000-8      	    7189	    179537 ns/op	   40595 B/op	      81 allocs/op
BenchmarkCompareIndexVsReverse/IndexPath_10000-8       	   22334	     53174 ns/op	   83432 B/op	      16 allocs/op
BenchmarkCompareIndexVsReverse/ReversePath_10000-8     	     572	   2040863 ns/op	  671415 B/op	      88 allocs/op
PASS
ok  	github.com/babylonlabs-io/babylon/v4/x/finality/keeper	10.324s
```

proto/babylon/finality/v1/finality.proto
```diff
@@ -8,106 +8,135 @@ import "gogoproto/gogo.proto";
...
+}
+
+// PubRandCommitIndexValue represents a list of sorted start heights for public
+// randomness commitments.
+message PubRandCommitIndexValue {
+  option (gogoproto.goproto_getters) = false;
+  // A list of start heights corresponding to committed public randomness
+  // ranges.
+  repeated uint64 heights = 1;
 }
```
RafilxTenfen, line 90: only change here is the `PubRandCommitIndexValue`

x/finality/keeper/genesis.go
```diff
@@ -30,7 +30,11 @@ func (k Keeper) InitGenesis(ctx context.Context, gs types.GenesisState) error {
 	}
 
 	for _, prc := range gs.PubRandCommit {
-		k.SetPubRandCommit(ctx, prc.FpBtcPk, prc.PubRandCommit)
+		// This func sets also the PubRandCommit index
```
RafilxTenfen, line 33: :+1:

x/finality/keeper/genesis.go
```diff
@@ -322,6 +332,28 @@ func (k Keeper) votingPowersDistCacheBlkHeight(ctx context.Context) ([]*types.Vo
 	return vps, nil
 }
 
+func (k Keeper) pubRandCommitIdxs(ctx context.Context) ([]*types.PubRandCommitIdx, error) {
+	idxs := make([]*types.PubRandCommitIdx, 0)
+	err := k.pubRandCommitIndex.Walk(ctx, nil, func(fpPkBytes []byte, idx types.PubRandCommitIndexValue) (stop bool, err error) {
+		fpPk, err := bbn.NewBIP340PubKey(fpPkBytes)
+		if err != nil {
+			return true, err
+		}
```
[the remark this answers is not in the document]
GAtom22: yep, the only way there's an error here is if the DB got corrupted and somehow a fpPK has a length different than the expected

proto/babylon/finality/v1/finality.proto
GAtom22: yes, the other diff is just formatting

x/finality/module.go
```diff
@@ -29,6 +29,8 @@ var (
 	_ module.AppModuleBasic     = AppModuleBasic{}
 )
 
+const consensusVersion = 2
```
RafilxTenfen, line 32: :100:

RafilxTenfen (approved):
a few suggestion, but looks good to me

x/finality/keeper/public_randomness.go
```diff
@@ -18,8 +21,65 @@ import (
...
+	if err == nil {
+		return commit, nil
+	}
+	// error occurred on index
+	// log it with WARN level and proceed with fallback
+	logger := k.Logger(sdk.UnwrapSDKContext(ctx))
+	logger.Warn("getPubRandCommitByIndex failed. Fallback to reverse iterator", "error", err.Error())
+
+	// Fallback to legacy reverse iteration
+	return k.getPubRandCommitByReverseScan(ctx, fpBtcPK, height)
```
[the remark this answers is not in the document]
GAtom22: after the migration, the fallback would never be used. I added it initially just in case the migration takes too long. The idea is to time the migration with testnet data to ensure will not be long the down time for this. But will remove this for now, and we'll have the changes for reference if after the testing we decide to avoid the migration.

PR #1062, opened 2025-06-04
chore(incentive): Add `anteHandler` to avoid fee grants on refundable tx

GAtom22:


x/incentive/keeper/reward_tracker_events.go
RafilxTenfen, line 1: split sec-45 fix for its own PR

x/incentive/keeper/refundable_msg_index.go
```diff
@@ -1,31 +1,99 @@
...
+	return k.bankKeeper.SendCoinsFromModuleToAccount(ctx, k.feeCollectorName, feeGranter, refund)
+}
+
+// restoreDeletedFeeGrant restores the fee grant as a basic allowance
+// with a 2 day expiration
+func (k Keeper) restoreDeletedFeeGrant(ctx context.Context, feeGranter, feePayer sdk.AccAddress, refund sdk.Coins) error {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+
+	// Set expiration time (48 hours from current block time)
+	expiration := sdkCtx.BlockTime().Add(48 * time.Hour)
```
RafilxTenfen, line 80: why do we keep the 2 days?

GAtom22: Good question! Here we have to decide what to do, if set a hardcoded expiration time (could be 2 days or whatever we decide), or somehow try to get what was the consumed grant expiration. For the time being, I believe setting a hardcoded expiration time is the best approach. Then we have to decide what is the period we want to set here

x/incentive/keeper/refundable_msg_index.go
```diff
@@ -1,31 +1,99 @@
...
+		return k.refundToFeeGranter(ctx, feeGranter, feePayer, txFee)
+	}
+
+	// refund to fee payer
+	return k.bankKeeper.SendCoinsFromModuleToAccount(ctx, k.feeCollectorName, feePayer, txFee)
+}
+
+// refundToFeeGranter restores the fee grant allowance by increasing the spend limit or restoring the deleted grant
+// and sends the refund to the fee granter
+func (k Keeper) refundToFeeGranter(ctx context.Context, feeGranter, feePayer sdk.AccAddress, refund sdk.Coins) error {
```
RafilxTenfen, line 41: this part got complex, could you create tests for this specific function with a real `feegrantKeeper` structure, not mock

RafilxTenfen: with expected allowance+balance checks after the `refundToFeeGranter` is called

GAtom22: yes, will work on additional tests

RafilxTenfen:
31. Fee grant allowance not restored during refunds causes silent grant depletion

x/incentive/keeper/refundable_msg_index.go
GAtom22: refactored the code to not restore the depleted allowance for this case.
So we avoid setting up a different allowance type than the one depleted

RafilxTenfen (dismissed):
Code looks good to me, but since it will not be in v2, let's wait for @KonradStaniec to take a look as well

Lazar955 (dismissed):

GAtom22:
> to gather some opinions, what do you guys think as about disallowing fee grants for the messages that are being refunded ? (as an alternative solution to the problem)
> 
> In my mind this would simplify things a bit, and there is case to be made that you do not need fee grant if the message is already refunded.

Yep, sounds good!
Not sure if any FP/node operator is using this, but if we take this approach, we'll need to communicate to them accordingly

x/incentive/keeper/refund_tx_decorator.go
```diff
@@ -18,6 +29,29 @@ func NewRefundTxDecorator(k *Keeper) *RefundTxDecorator {
 	}
 }
 
+// Makes sure there's no fee granter assigned in refundable txs
+// so there's no need to restore allowances
+func (d *RefundTxDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
```
KonradStaniec, line 34: maybe we can some docs in `x/incentive` module that explain that it is not possible to use fee grants for refundable transactions ?

KonradStaniec (dismissed):

x/incentive/keeper/refund_tx_decorator.go
GAtom22: updated the module's README. LMK if we want to include this somewhere else

PR #1064, opened 2025-06-05
fix: signing info validation for start height and missed block counter

Vvaradinov:
validation of `StartHeight` and `MissedBlockCounter` are greater than or equal to zero


RafilxTenfen (commented):
Changes looks good, it needs a test case explicity checking the err msg
Good remainder @samricotta

PR #1065, opened 2025-06-05
fix: add check for rewards period

Vvaradinov:
Adds check for period to be greater than zero


RafilxTenfen (changes requested):
needs a test case

Lazar955 (approved):
lgmt

x/incentive/types/rewards.go
```diff
@@ -97,6 +97,12 @@ func (f *FinalityProviderCurrentRewards) Validate() error {
 	if f.TotalActiveSat.IsNegative() {
 		return fmt.Errorf("current rewards has a negative total active satoshi delegated value: %s", f.TotalActiveSat.String())
 	}
+
+	//nolint:staticcheck
+	if f.Period < 0 {
+		return fmt.Errorf("period cannot be negative")
+	}
```
RafilxTenfen, line 104:
```suggestion
	if f.Period == 0 {
		return fmt.Errorf("fp current rewards period must be positive")
	}
```

PR #1071, opened 2025-06-05
chore: update `fpdistributioninfo` validations

samricotta:
Address needed to be validated if empty and if it is indeed a bech32 type


x/finality/types/power_table.go
```diff
@@ -223,6 +223,26 @@ func (fpdi FinalityProviderDistInfo) Validate() error {
 	if fpdi.BtcPk.Size() != bbn.BIP340PubKeyLen {
 		return fmt.Errorf("invalid fp dist info. finality provider BTC public key length: got %d, want %d", fpdi.BtcPk.Size(), bbn.BIP340PubKeyLen)
 	}
+
+	if fpdi.Addr == nil {
+		return fmt.Errorf("invalid fp dist info. empty finality provider address")
+	}
+
+	if _, err := sdk.AccAddressFromBech32(string(fpdi.Addr)); err != nil {
```
RafilxTenfen, line 231:
```suggestion
	if _, err := sdk.AccAddressFromBech32(sdk.AccAddress(fpdi.Addr).String()); err != nil {
```

PR #1098, opened 2025-06-05
imp: extend ValidateEntries func

GAtom22:
related to num-16


RafilxTenfen (commented):
Suggestion to add

```go
	if err := types.ValidateEntries(gs.BtcDelegations, func(b *BTCDelegation) string {
		return b.MustGetStakingTxHash().String()
	}); err != nil {
		return err
	}
```

in the validation of the genstate and also check that `gs.BTCDelegator` contains all the btc staking transaction from `gs.BtcDelegations` 

Each fp,del => stk txs (BTCDelegatorDelegationIndex)

x/btcstaking/types/genesis.go
```diff
@@ -183,6 +180,73 @@ func (e ConsumerEvent) Validate() error {
 	return nil
 }
 
+// validateBTCDelegations validates the BTC delegation and returns the
+// expected delegation index to compare with the provided on genesis state
+func validateBTCDelegations(delegations []*BTCDelegation) (map[string]*BTCDelegatorDelegationIndex, error) {
```
RafilxTenfen, line 185: nice :tada:

RafilxTenfen (approved):

PR #1146, opened 2025-06-06
chore: remove `x/crisis`

RafilxTenfen:
`x/crisis` module became [deprecated](https://github.com/cosmos/cosmos-sdk/blob/810d15dd6d16084c9ba6dfc018e9431172cd4786/x/crisis/README.md?plain=1#L7)

sec 9. Usage of deprecated x/crisis module allows attackers to DoS the chain

### DIFF
--- a/app/upgrades/v2/upgrades.go
+++ b/app/upgrades/v2/upgrades.go
@@ -46,6 +46,9 @@ const (
 	//
 	// To fully decouple from the module now, we hardcode the store name here.
 	InterchainQueryStoreName = "interchainquery"
+	// CrisisStoreName `x/crisis` module is deprecated at cosmos-sdk v0.53 and
+	// will be removed in the next release.
+	CrisisStoreName = "crisis"
 )
 
 var (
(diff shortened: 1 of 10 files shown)


app/upgrades/v2rc2/testnet/upgrade.go
```diff
@@ -18,7 +18,7 @@ var Upgrade = upgrades.Upgrade{
 	CreateUpgradeHandler: CreateUpgradeHandler,
 	StoreUpgrades: store.StoreUpgrades{
 		Added:   []string{},
-		Deleted: []string{v2.InterchainQueryStoreName},
+		Deleted: []string{v2.InterchainQueryStoreName, v2.CrisisStoreName},
```
RafilxTenfen, line 21: There is no harm in deleting the same store key more than once in the upgrade handler

app/upgrades/v2/upgrades.go
```diff
@@ -67,7 +70,7 @@ func CreateUpgrade(includeAsyncICQ bool) upgrades.Upgrade {
 		CreateUpgradeHandler: CreateUpgradeHandler,
 		StoreUpgrades: store.StoreUpgrades{
 			Added:   addedStoreUpgrades,
-			Deleted: []string{ibcfeetypes.StoreKey},
+			Deleted: []string{ibcfeetypes.StoreKey, CrisisStoreName},
```
vitsalis, line 73: Do we need to do the same thing we did with `addedStoreUpgrades`? Given that the testnet upgrade did not include the Crisis module being removed.

vitsalis (commented):
It's a pretty last minute change and not required for v2 (as v2 depends on 0.50 in which crisis is not deprecated). Are we sure we want to make the change now?

Vvaradinov (approved):
LGTM.

GAtom22 (approved):

PR #1152, opened 2025-06-06
chore: validate power non negative

Lazar955:
20. Missing validation of the Power field of a Validator may result in potential overflows


RafilxTenfen:
20. Missing validation of the Power field of a Validator may
result in potential overflows

PR #1229, opened 2025-06-12
feat: EVM inclusion

Vvaradinov:
This PR includes the minimum necessary changes to make the Babylon Genesis chain EVM compatible. This includes:

- `x/vm` module
- `x/erc20` module
- `x/feemarket` module
-  Static precompiles for bank, gov, slashing, evidence, bech32 and p256
- Token pair only for the native BABY token 
- Antehandler for EVM transactions
- `x/precisebank` module for gas conversions
-  Included config for EVM and JSON-RPC for `testnet` command

## How to test

Start the chain with the `testnet` command:

```
 make build && ./build/babylond testnet \
    --v                     1 \
    --output-dir            ./.testnet \
    --starting-ip-address   192.168.10.2 \
    --keyring-backend       test \
    --chain-id              chain-test \
    --additional-sender-account true \
&& ./build/babylond start --home ./.testnet/node0/babylond --log_level=info
```

This will create a chain with a new `eth_secp` account that you can import in Metamask with the following known dev mnemonic:

```
doll midnight silk carpet brush boring pluck office gown inquiry duck chief aim exit gain never tennis crime fragile ship cloud surface exotic patch
```

The configuration for adding a new chain to metamask is the following:

- **Network Name**: Babylon Genesis
- **RPC URL**: http://127.0.0.1:8545
- **Chain ID**: 6901 (can be modified to whatever we want here)
- **Currency Symbol**: BABY

## Known Issues

Due to limitations with the Cosmos EVM and the gas mechanism the EVM does not currently work with our `PriorityNonceMempool` so the `NoOp` mempool is configured in the `testnet` command. The issue was raised with the Cosmos EVM team.


app/app.go
```diff
@@ -801,7 +859,25 @@ func (app *BabylonApp) RegisterNodeService(clientCtx client.Context, cfg config.
 
 // DefaultGenesis returns a default genesis from the registered AppModuleBasic's.
 func (a *BabylonApp) DefaultGenesis() map[string]json.RawMessage {
-	return a.BasicModuleManager.DefaultGenesis(a.appCodec)
+	genesis := a.BasicModuleManager.DefaultGenesis(a.appCodec)
+	// TODO: Do we need this min denom ?
```
RafilxTenfen, line 863: I think without this, the mint inflation would not be in `ubbn`

app/app_test.go
```diff
@@ -29,7 +33,7 @@ import (
 
 var (
 	expectedMaccPerms = map[string][]string{
-		authtypes.FeeCollectorName:   nil, // fee collector account
+		authtypes.FeeCollectorName:   {authtypes.Burner}, // fee collector account
```
RafilxTenfen, line 36: why the feecollector can burn?

app/token_pairs.go
```diff
@@ -0,0 +1,17 @@
+package app
+
+import erc20types "github.com/cosmos/evm/x/erc20/types"
+
+// WTokenContractMainnet is the WrappedToken contract address for mainnet
+const WTokenContractMainnet = "0xD4949664cD82660AaE99bEdc034a0deA8A0bd517"
+
+// TokenPairs creates a slice of token pairs, that contains a pair for the native denom of the example chain
+// implementation.
+var TokenPairs = []erc20types.TokenPair{
```
RafilxTenfen, line 10: can we add more details of what a token pair means? 
Correct me if I am wrong: This is needed so the BABY `ubbn` token from cosmos side works can go back and forth with evm

cmd/babylond/cmd/root.go
```diff
@@ -122,6 +134,12 @@ func NewRootCmd() *cobra.Command {
 
 	initRootCmd(rootCmd, tempApp.TxConfig(), tempApp.BasicModuleManager)
 
+	if initClientCtx.ChainID != "" {
```
RafilxTenfen, line 137: can you add a comment for the reason about this?

app/app_test.go
RafilxTenfen: so the fees do not go to feecollector?

app/keepers/precompiles.go
```diff
@@ -0,0 +1,79 @@
...
+	bankKeeper precisebankkeeper.Keeper,
+	erc20Keeper erc20Keeper.Keeper,
+	govKeeper govkeeper.Keeper,
+	slashingKeeper slashingkeeper.Keeper,
+	evidenceKeeper evidencekeeper.Keeper,
+) map[common.Address]vm.PrecompiledContract {
+	// TODO: We can add more custom precompiles here for Babylon Modules
+
+	// Clone the mapping from the latest EVM fork.
+	precompiles := maps.Clone(vm.PrecompiledContractsBerlin)
```
GAtom22, line 38: latest updated precompiles would be `PrecompiledContractsPrague`
And by default, this is the active fork on the evm config

app/app.go
```diff
@@ -873,15 +944,24 @@ func GetMaccPerms() map[string][]string {
...
 	}
 
 	// allow the following addresses to receive funds
-	delete(modAccAddrs, appparams.AccGov.String())
+	delete(blockedAddrs, appparams.AccGov.String())
 
-	return modAccAddrs
+	// Block precompiled contracts
+	blockedPrecompilesHex := evmtypes.AvailableStaticPrecompiles
+	for _, addr := range vm.PrecompiledAddressesBerlin {
```
GAtom22, line 957: latest available precompiles is `PrecompiledAddressesPrague`. Should we update to it?

app/token_pairs.go
```diff
@@ -0,0 +1,17 @@
+package app
+
+import erc20types "github.com/cosmos/evm/x/erc20/types"
+
+// WTokenContractMainnet is the WrappedToken contract address for mainnet
+const WTokenContractMainnet = "0xD4949664cD82660AaE99bEdc034a0deA8A0bd517"
```
GAtom22, line 9: can you please add a comment where this address come from?

app/token_pairs.go
GAtom22, line 14: Nit: rename to DefaultTokenPairs
```suggestion
var DefaultTokenPairs = []erc20types.TokenPair{
```

app/app_test.go
RafilxTenfen: thanks for explaining

cmd/babylond/cmd/genesis.go
```diff
@@ -218,6 +223,24 @@ func PrepareGenesis(
 	authGenState.Accounts = genesisParams.AuthAccounts
 	genesisState[authtypes.ModuleName] = cdc.MustMarshalJSON(authGenState)
 
+	// Add EVM genesis configuration
+	evmGenState := evmtypes.DefaultGenesisState()
+	evmGenState.Params.ActiveStaticPrecompiles = evmtypes.AvailableStaticPrecompiles
```
GAtom22, line 228: should we update this list to the ones we later on wire up? to avoid confusion

cmd/babylond/cmd/root.go
```diff
@@ -122,6 +135,13 @@ func NewRootCmd() *cobra.Command {
 
 	initRootCmd(rootCmd, tempApp.TxConfig(), tempApp.BasicModuleManager)
 
+	// If the regular chain id set, ensure we also set an EVM specific chain id.
+	if initClientCtx.ChainID != "" {
```
GAtom22, line 139: This is causing a panic when running for example the gentx cmd

If I remove the conditional works OK. Is there a specific reason why we want to keep this conditional?
'cause the resulting EVMAppOptions will be the same
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x339c339]

goroutine 1 [running]:
github.com/cosmos/evm/x/vm/types.GetEVMCoinDecimals(...)
        github.com/cosmos/evm@v1.0.0-rc0.0.20250602235914-3eb2135b9103/x/vm/types/denom_config.go:53
github.com/cosmos/evm/x/precisebank/types.ConversionFactor()
        github.com/cosmos/evm@v1.0.0-rc0.0.20250602235914-3eb2135b9103/x/precisebank/types/fractional_balance.go:18 +0x19
github.com/cosmos/evm/x/precisebank/types.(*GenesisState).Validate(0xc0038825e0)
        github.com/cosmos/evm@v1.0.0-rc0.0.20250602235914-3eb2135b9103/x/precisebank/types/genesis.go:42 +0xcf
github.com/cosmos/evm/x/precisebank.AppModuleBasic.ValidateGenesis({}, {0x5c57ee0, 0xc00196d8e0}, {0x48a4d80?, 0x5c57e01?}, {0xc003c15200, 0x34, 0x40})
        github.com/cosmos/evm@v1.0.0-rc0.0.20250602235914-3eb2135b9103/x/precisebank/module.go:81 +0x8c
github.com/cosmos/cosmos-sdk/types/module.coreAppModuleBasicAdaptor.ValidateGenesis({{0x499696a?, 0x42541f?}, {0x7f8eadfad498?, 0xc00227e660?}}, {0x5c57ee0, 0xc00196d8e0}, 
```

cmd/babylond/cmd/custom_babylon_config.go
```diff
@@ -57,6 +58,11 @@ type BabylonAppConfig struct {
...
 
 	BabylonMempoolConfig BabylonMempoolConfig `mapstructure:"babylon-mempool"`
+
+	// EVM config
+	EVM     evmserverconfig.EVMConfig     `mapstructure:"evm"`
+	JSONRPC evmserverconfig.JSONRPCConfig `mapstructure:"json-rpc"`
+	TLS     evmserverconfig.TLSConfig     `mapstructure:"tls"`
 }
 
 func DefaultBabylonAppConfig() *BabylonAppConfig {
```
GAtom22, line 68: Here we should also set default evm chain id `evm-chain-id` on the evm config

Vvaradinov:
> We should make eth_secp256k1 be the default now (or support both if we want to keep secp256k1)

It should be both I think because changing the default to eth_secp256k1 breaks a lot of the e2e tests.

app/token_pairs.go
Vvaradinov: I believe this is arbitrary since we are launching the EVM without the previous existence of a WETH contract. By upgrade time we could choose this and instantiate the precompile for it.

GAtom22:
when submitting a tx via metamask, I'm getting this error. Maybe is related to the wiring of the custom GetSigner for the evm msg
```
{
  "code": -32603,
  "message": "Internal JSON-RPC error.",
  "data": {
    "code": -32000,
    "message": "tx must have at least one signer: internal",
    "cause": null
  },
  "stack": "Error: Internal JSON-RPC error.\n    at new a (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/common-3.js:7:5787)\n    at o (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/common-3.js:7:8878)\n    at Object.internal (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/common-3.js:7:9487)\n    at chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/common-3.js:3:311186\n    at async chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/common-2.js:1:1363609"
}
```
looking into this...

KonradStaniec (approved):
code looks good mostly question and comments how to document some work that will need to be done in the future (upgrade and mempool)

cmd/babylond/cmd/testnet.go
```diff
@@ -197,6 +204,14 @@ func InitTestnet(
 	babylonConfig.GRPC.Enable = true
 	babylonConfig.GRPC.Address = "0.0.0.0:9090"
 
+	// Update babylonConfig to include Ethereum JSON-RPC settings and other settings
+	babylonConfig.Mempool.MaxTxs = -1 // NOTE: No-op mempool required
```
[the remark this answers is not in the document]
Vvaradinov: WIl replace it as default and add a TODO in the code directly.

GAtom22:
> tx must have at least one signer

Yes, that was the issue

GAtom22 (approved):
tACK! Great work @Vvaradinov!!
Tested some basic txs and all worked well :rocket:

cmd/babylond/cmd/root.go
```diff
@@ -135,13 +135,6 @@ func NewRootCmd() *cobra.Command {
 
 	initRootCmd(rootCmd, tempApp.TxConfig(), tempApp.BasicModuleManager)
 
-	// If the regular chain id set, ensure we also set an EVM specific chain id.
-	if initClientCtx.ChainID != "" {
-		if err := app.EVMAppOptions(appparams.EVMChainID); err != nil {
-			panic(err)
-		}
```
GAtom22: we need this, otherwise there's a panic when using some commands (eg. gentx)

cmd/babylond/cmd/root.go
```diff
@@ -122,6 +136,13 @@ func NewRootCmd() *cobra.Command {
 
 	initRootCmd(rootCmd, tempApp.TxConfig(), tempApp.BasicModuleManager)
 
+	// If the regular chain id set, ensure we also set an EVM specific chain id.
+	if initClientCtx.ChainID != "" {
+		if err := app.EVMAppOptions(appparams.EVMChainID); err != nil {
+			panic(err)
+		}
+	}
```
GAtom22, line 144:
```suggestion
	if err := app.EVMAppOptions(appparams.EVMChainID); err != nil {
		panic(err)
	}
```

app/app.go
```diff
@@ -172,6 +189,10 @@ var (
 )
 
 func init() {
+	// Replace defaults
+	// manually update the power reduction by replacing micro (u) -> atto (a)
+	sdk.DefaultPowerReduction = cosmosevmtypes.MicroPowerReduction
+	stakingtypes.DefaultMinCommissionRate = math.LegacyZeroDec()
```
GAtom22, line 195: This is [already zero](https://github.com/cosmos/cosmos-sdk/blob/e265bb9b42308dc70743b6200a70db9aafb70527/x/staking/types/params.go#L35)

app/app.go
```diff
@@ -172,6 +189,10 @@ var (
 )
 
 func init() {
+	// Replace defaults
+	// manually update the power reduction by replacing micro (u) -> atto (a)
+	sdk.DefaultPowerReduction = cosmosevmtypes.MicroPowerReduction
```
GAtom22, line 194: This should be the same, no?
Cause the[ DefaultPowerReduction on the sdk is 1000000](https://github.com/cosmos/cosmos-sdk/blob/e265bb9b42308dc70743b6200a70db9aafb70527/types/staking.go#L24)
and on the evm, the [MicroPowerReduction is also 1e6 (1000000)](https://github.com/cosmos/evm/blob/73bc2435c3886d7ae181b992a7e6dd2e1c04093b/types/power.go#L14)

app/app.go
```diff
@@ -801,7 +854,20 @@ func (app *BabylonApp) RegisterNodeService(clientCtx client.Context, cfg config.
...
+	evmGenState.Params.ActiveStaticPrecompiles = evmtypes.AvailableStaticPrecompiles
+	evmGenState.Params.EvmDenom = appparams.BaseCosmosDenom
+	genesis[evmtypes.ModuleName] = a.appCodec.MustMarshalJSON(evmGenState)
+
+	// Add ERC20 genesis configuration
+	erc20GenState := erc20types.DefaultGenesisState()
+	erc20GenState.TokenPairs = DefaultTokenPairs
+	erc20GenState.Params.NativePrecompiles = append(erc20GenState.Params.NativePrecompiles, WTokenContractMainnet)
+	genesis[erc20types.ModuleName] = a.appCodec.MustMarshalJSON(erc20GenState)
+
```
GAtom22, line 869: Here we should add also the default genesis for feemarket module (ei, initial base fee and min gas price)

app/keepers/keepers.go
```diff
@@ -758,6 +845,12 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
 	paramsKeeper.Subspace(tokenfactorytypes.ModuleName)
 	paramsKeeper.Subspace(ratelimittypes.ModuleName)
 
+	// Subspaces for EVM modules
+	// TODO: We probably don't need this.
```
GAtom22, line 849: yes, this is not really necessary

app/params/config.go
```diff
@@ -17,6 +17,15 @@ const (
 
 	// Bech32PrefixAccAddr defines the Bech32 prefix of an account's address.
 	Bech32PrefixAccAddr = "bbn"
+
+	// BaseCosmosDenom is the base denomination for 6 decimals.
+	BaseCosmosDenom = "ubbn"
+	// BaseEVMDenom is the base denomination for 18 decimals.
+	BaseEVMDenom = "abbn"
+	// DisplayDenom is the display denom of the Cosmos EVM
+	DisplayDenom = "bbn"
```
GAtom22, line 26: is this different from `HumanCoinUnit`? or we can use that instead of adding a new const?

PR #1243, opened 2025-06-18
fix: add ante handler to block potential valset update messages end epoch

RafilxTenfen:


meetrick (commented):
LGTM, great

meetrick:
This solution is limited to the unjail message, right?
Blocking the TX when it’s the last block of the epoch and it contains a MsgUnjail seems like a reasonable approach.
Once the chain is up and running again, I guess we’ll need to look into other corner cases as well.

PR #1252, opened 2025-06-23
Implement context separator signing

KonradStaniec:
Implements  scheme described in: https://github.com/babylonlabs-io/pm/blob/main/rfc/rfc-036-replay-attack-protection.md


testutil/datagen/btcstaking.go
```diff
@@ -32,36 +32,52 @@ const (
...
 	btcSK, _, err := GenRandomBTCKeyPair(r)
 	if err != nil {
 		return nil, err
 	}
-	return GenRandomCreateFinalityProviderMsgWithBTCBabylonSKs(r, btcSK, GenRandomAccount().GetAddress())
+	return GenRandomCreateFinalityProviderMsgWithBTCBabylonSKs(r, btcSK, signingContext, GenRandomAccount().GetAddress())
 }
 
-func CreateNFinalityProviders(r *rand.Rand, t *testing.T, n int) []*bstypes.FinalityProvider {
+func CreateNFinalityProviders(
```
Lazar955, line 59: Same comment as above, regarding struct as param

Lazar955 (dismissed):

app/upgrades/v1rc8/testnet/upgrade_test.go
```diff
@@ -37,7 +37,7 @@ func FuzzMigrateFinalityProviders(f *testing.F) {
 		// slice of the expected finality providers after the migration
 		expFps := make([]btcstakingtypes.FinalityProvider, fpCount)
 		for i := range expFps {
-			fp, err := datagen.GenRandomFinalityProvider(r)
+			fp, err := datagen.GenRandomFinalityProvider(r, "")
```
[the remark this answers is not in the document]
KonradStaniec: In general, I agree with application code 👍 

Here I am a bit torn as `datagen.GenRandomFinalityProvider` comes from `testutil` package, which is mostly for testing Babylon node and we mostly change it adjust for Babylon node (not other programs)

On the other hand, case can be made that it should not be public if we do not want maintain clear interfaces for this package.

> Review what functions are exported

I think all functions in `testutil/datagen` package are exported, as we add them there whenever we need one . This makes it a bit random package wise, without clear docs and interfaces.

x/btcstaking/types/pop.go
```diff
@@ -41,14 +41,20 @@ func (pop *ProofOfPossessionBTC) ToHexStr() (string, error) {
...
 	switch pop.BtcSigType {
 	case BTCSigType_BIP340:
-		return pop.VerifyBIP340(staker, bip340PK)
+		return pop.VerifyBIP340(signingContext, address, bip340PK)
 	case BTCSigType_BIP322:
-		return pop.VerifyBIP322(staker, bip340PK, net)
+		return pop.VerifyBIP322(signingContext, address, bip340PK, net)
 	case BTCSigType_ECDSA:
-		return pop.VerifyECDSA(staker, bip340PK)
+		return pop.VerifyECDSA(signingContext+stakerBech32Addr, bip340PK)
```
RafilxTenfen, line 57: suggestion for us to keep the pattern as the signingContext as a first parameter

Lazar955 (approved):

RafilxTenfen (approved):

Issue #1026, opened 2025-06-03
zoneconcierge: remove unused features


SebastianElvis:
Many features in Zone Concierge module is not relevant anymore. Those include

- `chain_info_indexer.go` [file](https://github.com/babylonlabs-io/babylon/blob/b9774782f38e9758c4f5aafab1e1e45dde0f3838/x/zoneconcierge/keeper/chain_info_indexer.go). It was used for timestamping phase-1 protocol (Babylon as a canonical chain oracle).
- Simialrly, `QueryChainsInfo`query can be removed.
- [QueryChainList](https://github.com/babylonlabs-io/babylon/blob/3efca9d827caf7e203d33d3bcee16f62a0ae1119/x/zoneconcierge/keeper/grpc_query.go#L26-L50). It was used for querying all chains integrated with timestamping phase-1 protocol. Now this is not relevant and is conflicting with the consumer registry in `x/btcstkconsumer` module
- `ListEpochHeaders` [query](https://github.com/babylonlabs-io/babylon/blob/3efca9d827caf7e203d33d3bcee16f62a0ae1119/x/zoneconcierge/keeper/grpc_query.go#L196-L197). It was used for supporting the front-end explorer of timestamping phase-1 protocol.

PR #1265, opened 2025-06-27
chore(zoneconcierge): Remove `ListEpochHeaders` and `QueryChainList`  query

Vvaradinov:
Removing old query used for timestamping ref - https://github.com/babylonlabs-io/babylon/issues/1026


SebastianElvis (approved):

SebastianElvis:
Btw, do we need to remove relevant CLI commands for these queries?

Vvaradinov:
@SebastianElvis seems like there were no CLI commands added for these queries .

PR #1271, opened 2025-07-01
rollup: equivocation e2e test for rollup BSN

### LINKED ISSUE #948: Update consumer FPs after slashing
SlashFinalityProvider [function](https://github.com/babylonlabs-io/babylon/blob/8aa48db108d9be4c35e08584490a95b2d9e344d9/x/btcstaking/keeper/finality_providers.go#L108-L116) in the BTC staking module only checks Babylon FP in the BTC staking module, but not consumer FP in the BTC staking consumer module. So the current implementation is not keeping `SlashedBabylonHeight` or `SlashedBtcHeight` values in the finality provider struct in sync.

### LINKED ISSUE #1253: finality: validation of MsgEquivocationEvidence
https://github.com/babylonlabs-io/babylon/blob/main/x/finality/keeper/msg_server.go#L398 require some validations that provided evidence is valid. Otherwise anybody can slash anybody

SebastianElvis:
Part of https://github.com/babylonlabs-io/pm/issues/412

This PR adds e2e test for equivocation in the rollup BSN contract, in particular the handling of equivocation evidence sent from the CosmWasm contract. This includes

- [x] Adjust e2e and scripts due to repo renaming
- [x] Add e2e for equivocation
- [x] Custom message handler for EquivocationEvidence
- [x] Add full validations for the EquivocationEvidence handler, and fix unit test (#1253)
- [x] Update consumer FP status upon slashing (#948)
- [x] Fix equivocation evidence submission (this requires https://github.com/babylonlabs-io/rollup-bsn-contracts/issues/19)

Resolves #1253 
Resolves #948 

TODO before merging:

- [x] merge https://github.com/babylonlabs-io/rollup-bsn-contracts/pull/23


gusin13 (approved):
nice work, lgtm!

x/btcstaking/keeper/finality_providers.go
```diff
@@ -107,33 +107,47 @@ func (k Keeper) GetFinalityProvider(ctx context.Context, fpBTCPK []byte) (*types
...
+	}
+
+	// Then try to get as consumer finality provider
+	// Get the consumer ID for this finality provider
+	consumerID, err := k.BscKeeper.GetConsumerOfFinalityProvider(ctx, fpBTCPKObj)
+	if err != nil {
+		return fmt.Errorf("failed to get consumer of finality provider: %w", err)
+	}
+	// Use the existing SlashConsumerFinalityProvider function
+	return k.SlashConsumerFinalityProvider(ctx, consumerID, fpBTCPKObj)
```
KonradStaniec, line 150: Not for this pr probably, but I have noticed that `SlashConsumerFinalityProvider`, add those new delegation power update events:

```go
	// Process all delegations for this consumer finality provider and record slashed events
	err = k.HandleFPBTCDelegations(ctx, fpBTCPK, func(btcDel *types.BTCDelegation) error {
		stakingTxHash := btcDel.MustGetStakingTxHash().String()
		eventSlashedBTCDelegation := types.NewEventPowerDistUpdateWithSlashedBTCDelegation(stakingTxHash)
		k.addPowerDistUpdateEvent(ctx, btcTip.Height, eventSlashedBTCDelegation)
		return nil
	})
```

but those events are never processed in our `power_dist_change.go` loop:
```go
for _, event := range events {
  switch typedEvent := event.Ev.(type) {
  case *types.EventPowerDistUpdate_BtcDelStateUpdate:

  case *types.EventPowerDistUpdate_SlashedFp:

  case *types.EventPowerDistUpdate_JailedFp:

  case *types.EventPowerDistUpdate_UnjailedFp:

}
```

is this by desigh or something weird is happening here ?

x/finality/types/msg.go
```diff
@@ -229,27 +229,50 @@ func (m *MsgResumeFinalityProposal) ValidateBasic() error {
...
+
+	return &Evidence{
+		FpBtcPk:              fpBtcPk,
+		BlockHeight:          m.BlockHeight,
+		PubRand:              pubRand,
+		CanonicalAppHash:     canonicalAppHash,
+		ForkAppHash:          forkAppHash,
+		CanonicalFinalitySig: canonicalFinalitySig,
+		ForkFinalitySig:      forkFinalitySig,
+		SigningContext:       m.SigningContext,
```
KonradStaniec, line 276: we could also validate that `m.SigningContext` has 64 bytes (it is hex encoded 32byte hash)

KonradStaniec (approved):
code lgtm!

I am mostly concerned why this new `EventPowerDistUpdate_SlashedBtcDelegation` power event is needed if we never use it ?

x/btcstaking/keeper/finality_providers.go
SebastianElvis: Hmm that's weird. From my side they are processed https://github.com/babylonlabs-io/babylon/blob/754f5c5cbab6e017e5c29f07a3b56a62de3cc453/x/finality/keeper/power_dist_change.go#L213-L267. What is off here?

x/finality/types/msg.go
SebastianElvis: Hmm we could, but this might not improve security, given that the signature extraction takes into account the signed message and thus the signing context anyway. Also we haven't implemented the signing context in rollup BSN contract. We can revisit this afterwards. Wdyt?

x/btcstaking/keeper/finality_providers.go
KonradStaniec: I mean if you look at the type switch: https://github.com/babylonlabs-io/babylon/blob/main/x/finality/keeper/power_dist_change.go#L200, the is not `EventPowerDistUpdate_SlashedBtcDelegation` anywhere in here ?

SebastianElvis: I see your question. So this event is used here https://github.com/babylonlabs-io/babylon/blob/c9e801b48ef6b04a9edf7b56289f860ff95bc9d7/x/btcstaking/keeper/finality_providers.go#L175-L184

Basically we use this enum type to construct events that a BTC delegation is slashed, and the events will be assembled into IBC packets sent to Cosmos BSNs. Cosmos BSNs will then label those slashed BTC delegations as inactive and discount their voting power.

SebastianElvis: Hmm good find. Actually the event propagation happens here https://github.com/babylonlabs-io/babylon/blob/c9e801b48ef6b04a9edf7b56289f860ff95bc9d7/x/finality/keeper/msg_server.go#L380-L383

But we are not processing the slashed BTC delegation events locally. This seems a bug imo -- Babylon should also process these events such that all slashed BTC delegations' voting power should be discounted in Babylon FPs, i.e., they should be unbonded. I remember we had this before but somehow it's gone now. @gusin13 any ideas?

gusin13: ok i found the original pr which added this feature [here](https://github.com/babylonlabs-io/babylon/pull/85/files#diff-fc43ea596a96556ec4069740b046cb194c311de20eb01c2d94cbc230b1a465d7R203), 

it had the switch case which Konrad mentioned [here](https://github.com/babylonlabs-io/babylon/pull/85/files#diff-fc43ea596a96556ec4069740b046cb194c311de20eb01c2d94cbc230b1a465d7R203), and then later such slashed delegations were skipped [here](https://github.com/babylonlabs-io/babylon/pull/85/files#diff-fc43ea596a96556ec4069740b046cb194c311de20eb01c2d94cbc230b1a465d7R254)

maybe when we merged the base branch to main it got removed smh, this is definitely a bug

Issue #1251, opened 2025-06-23
zoneconcierge: Non-deterministic iteration map may cause consensus failure and chain halt


dongsam:
Non-deterministic iteration over `BTCStakingConsumerIBCPacket` map may cause consensus failure and chain halt

---

### Summary

The use of `map` iteration in Go without ordering in the processing of `BTCStakingConsumerIBCPackets` introduces nondeterminism across nodes, which may result in inconsistent state commitments, IBC packet sequences, and potential consensus failure or chain halts.

---

### Description

In Go, iteration over `map` types is not deterministic and can result in different traversal orders across nodes due to runtime and memory state.

(Ref: [Go Blog - Map Iteration Order](https://go.dev/blog/maps?#iteration-order))

Currently, the method `GetAllBTCStakingConsumerIBCPackets`
https://github.com/babylonlabs-io/babylon/blob/1c1c399c654a97b6b94a35a4fc86e7c5bcc80d62/x/btcstaking/keeper/btc_staking_consumer_events.go#L63-L78

returns a `map[string]*BTCStakingIBCPacket` that is iterated directly in at least two locations:

---

### 1. In `BroadcastBTCStakingConsumerEvents` (executed in `EndBlock`)

https://github.com/babylonlabs-io/babylon/blob/1c1c399c654a97b6b94a35a4fc86e7c5bcc80d62/x/zoneconcierge/keeper/ibc_packet_btc_staking_consumer_event.go#L40-L66

- The IBC packets are iterated and sent via `SendIBCPacket`, where packet sequence is stateful and determined per `(portID, channelID)`.
- If packet order differs across nodes, [`SendPacket` may increment sequence values](https://github.com/cosmos/ibc-go/blob/e120ef5d4778c3e659ce57b59f028b250be5bb2e/modules/core/04-channel/keeper/packet.go#L37-L82) differently, breaking IAVL and AppHash consistency.
- This can result in:
    - Consensus failures and chain halts
    - Packet delivery failures due to sequence mismatch (from a relayer’s perspective)

---

### 2. In `btcstaking` module’s `ExportGenesis`

https://github.com/babylonlabs-io/babylon/blob/1c1c399c654a97b6b94a35a4fc86e7c5bcc80d62/x/btcstaking/keeper/genesis.go#L273-L284

- The `GetAllBTCStakingConsumerIBCPackets` map is iterated to build a `[]*ConsumerEvent` slice.
[...]

PR #1273, opened 2025-07-01
fix(zoneconcierge): sort map keysbefore iterating to avoid AppHash errors.

Vvaradinov:
Due to the nature of map iterating in Go being indeterministic we need to first sort the map by it's keys. Closes https://github.com/babylonlabs-io/babylon/issues/1251


SebastianElvis:
also should this be backproted to v3?

gusin13 (approved):

SebastianElvis (approved):

Issue #1239, opened 2025-06-18
zoneconcierge: optimise BTC light client sync


SebastianElvis:
The current `x/zoneconceirge` module implements a baseline protocol for synchronising BTC headers with Cosmos BSNs. The baseline protocol falls short of the efficiency. In particular, upon any BTC fork observed, Babylon will forward `w` latest BTC headers to all Cosmos BSNs. On the Babylon mainnet, `w = 300`. This incurs non-negligible overhead.

We will need to implement an optimised algorithm for sending BTC headers upon fork.

PR #1285, opened 2025-07-02
fix(zoneconcierge): use the k value for synchronising BTC headers with Cosmos BSNs

Vvaradinov:
Closes - https://github.com/babylonlabs-io/babylon/issues/1239

### DIFF
--- a/x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
+++ b/x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
@@ -87,11 +87,11 @@ func (k Keeper) createBTCTimestamp(
 ) (*types.BTCTimestamp, error) {
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	// if the Babylon contract in this channel has not been initialised, get headers from
-	// the tip to (w+1+len(finalizedInfo.BTCHeaders))-deep header
+	// the tip to (k+1+len(finalizedInfo.BTCHeaders))-deep header for efficient initialization
 	var btcHeaders []*btclctypes.BTCHeaderInfo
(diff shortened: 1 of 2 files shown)


x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
```diff
@@ -140,13 +140,13 @@ func (k Keeper) createBTCTimestamp(
...
 }
 
-// getDeepEnoughBTCHeaders returns the last w+1 BTC headers, in which the 1st BTC header
-// must be in the canonical chain assuming w-long reorg will never happen
-// This function will only be triggered upon a finalised epoch, where w-deep BTC checkpoint
-// is guaranteed. Thus the function is safe to be called upon generating BTC timestamps
+// getDeepEnoughBTCHeaders returns the last k+1 BTC headers for fork scenarios,
+// where k is the confirmation depth. This provides sufficient safety against reorgs
+// while being more efficient than using the finalization timeout w.
+// For normal broadcasting (non-fork), this uses k instead of w for efficiency.
```
SebastianElvis, line 146: We probably don't need to mention this

SebastianElvis (approved):
Maybe not for this PR, I'm wondering whether there is a way to have unit tests for IBC related functionalities in zone concierge module. We managed to have it before Cosmos SDK v0.50 but not anymore. It would be great if we can use unit tests for those rather than totally relying on e2e tests

Vvaradinov:
@SebastianElvis I have generally only tested IBC with either the interhcain e2e testing suite or our custom nix setup. But I've also seen this coordinator approach that the ibc-go team uses - https://github.com/cosmos/ibc-go/tree/main/testing

PR #1288, opened 2025-07-03
chore(zoneconcierge): remove unused queries from `x/zoneconcierge`

Vvaradinov:
Remove some old unused queries that relied on phase-1 timestamping protocol and `ChainInfo`. @SebastianElvis Can you double check if we need all of the other queries that we currently have ?


SebastianElvis (approved):

proto/babylon/zoneconcierge/v1/query.proto
```diff
@@ -134,17 +73,14 @@ message QueryFinalizedChainInfoUntilHeightRequest {
 // QueryFinalizedChainInfoUntilHeightResponse is response type for the
 // Query/FinalizedChainInfoUntilHeight RPC method.
 message QueryFinalizedChainInfoUntilHeightResponse {
-  // finalized_chain_info is the info of the Consumer
-  babylon.zoneconcierge.v1.ChainInfo finalized_chain_info = 1;
```
vitsalis, line 138: Why do we remove this field?

vitsalis: sounds good

PR #1295, opened 2025-07-04
chore: rename consumer_id to bsn_id for finality providers

vitsalis:
At the moment, every finality provider is associated with a "Consumer ID". This might make sense in the case of "consumer BSNs", but in the case of Babylon Genesis (a BSN itself), we are storing the Babylon Genesis chain ID as the "Consumer ID" of the finality provider. This makes the "consumer" term a very loaded one, which I recommend we only use for BSNs that consume information from Babylon Genesis (and therefore not all BSNs).

In the case of finality providers, I believe "BSN ID" is a better term to associate them with (as it covers both Babylon Genesis and consumer BSNs). Created a PR to do this renaming.

Note: the PR tries to avoid modifying all instances of "consumer", especially ones that are not related to the finality provider structures and definitely not touching the x/zoneconcierge and x/btcstkconsumer modules. Given that the x/btcstkconsumer module is going to act as a registry for "consumer BSNs" and x/zoneconcierge interacts with "consumer BSNs", I think this terminology is fine for those modules. There might be left overs associated with finality providers on x/btcstkconsumer, but they can be cleaned up when we transition finality providers to x/btcstaking.

### DIFF
--- a/client/docs/swagger-ui/swagger.yaml
+++ b/client/docs/swagger-ui/swagger.yaml
@@ -4881,16 +4881,11 @@ paths:
                           description: >-
                             update_time is the last time the commission rate was
                             changed.
-                    consumer_id:
+                    bsn_id:
                       type: string
-                      description: >-
-                        consumer_id is the ID of the consumer the finality
-                        provider is operating
-
-                        on. If it's missing / empty, it's assumed the finality
-                        provider is
-
-                        operating in the Babylon chain.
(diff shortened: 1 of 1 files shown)


KonradStaniec (commented):
please run `make proto-swagger-gen` to regenerate swagger file

KonradStaniec (approved):

PR #1296, opened 2025-07-04
queue events only for cosmos bsns

KonradStaniec:
Events to send via IBC are only queued for cosmos zones

### DIFF
--- a/x/btcstaking/keeper/btc_staking_consumer_events_test.go
+++ b/x/btcstaking/keeper/btc_staking_consumer_events_test.go
@@ -391,16 +391,17 @@ func TestDeterministicOrdering(t *testing.T) {
 	bsnIds := []string{"bsn-z", "bsn-a", "bsn-m", "bsn-b"}
 
 	// register consumers as cosmos consumers
-	for _, consumerID := range consumerIDs {
+	for _, bsnId := range bsnIds {
 		consumerRegister := &bsctypes.ConsumerRegister{
-			ConsumerId: consumerID,
+			ConsumerId: bsnId,
 			ConsumerMetadata: &bsctypes.ConsumerRegister_CosmosConsumerMetadata{
 				CosmosConsumerMetadata: &bsctypes.CosmosConsumerMetadata{
-					ChannelId: consumerID,
+					ChannelId: bsnId,
 				},
 			},
 		}
-		h.BTCStkConsumerKeeper.RegisterConsumer(h.Ctx, consumerRegister)
+		err := h.BTCStkConsumerKeeper.RegisterConsumer(h.Ctx, consumerRegister)
+		require.NoError(t, err)
 	}
 
 	// Add consumer events in random order
--- a/x/btcstaking/keeper/genesis_test.go
+++ b/x/btcstaking/keeper/genesis_test.go
@@ -113,7 +113,8 @@ func TestExportGenesis(t *testing.T) {
 				},
 			},
 		}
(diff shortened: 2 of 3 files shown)


x/btcstaking/keeper/btc_staking_consumer_events.go
```diff
@@ -9,11 +9,28 @@ import (
...
+// if the provided consumer is not a cosmos consumer this function will be noop
+// Consumer must be already registered in btcstkconsumer module, if not, function will panic.
 func (k Keeper) AddBTCStakingConsumerEvent(ctx context.Context, consumerID string, event *types.BTCStakingConsumerEvent) error {
 	return k.AddBTCStakingConsumerEvents(ctx, consumerID, []*types.BTCStakingConsumerEvent{event})
 }
 
+// AddBTCStakingConsumerEvents adds events that must be sent to cosmos consumer zones
+// if the provided consumer is not a cosmos consumer this function will be noop
+// Consumer must be already registered in btcstkconsumer module, if not, function will panic.
 func (k Keeper) AddBTCStakingConsumerEvents(ctx context.Context, consumerID string, events []*types.BTCStakingConsumerEvent) error {
```
gusin13, line 22: the fix proposed in this PR works fine but ideally we should fix this on the caller level, as we still looping through all the consumers [in this code](https://github.com/babylonlabs-io/babylon/blob/d3d81178dc38c172edaf5651c72b296bb9371a48/x/btcstaking/keeper/btc_delegations.go#L158) which is bit inefficient, so i had following in mind but this is nice to have
1. we can optimize all callers to pre-filter cosmos consumer and then feed  `AddBTCStakingConsumerEvents`
2. we can still keep the defensive check inside `AddBTCStakingConsumerEvents` if needed 
3. we might need to update function signature/names accordingly to depict this is only cosmos consumer events

something like this
```
func (k Keeper) notifyConsumersOnActiveBTCDel(ctx context.Context, btcDel *types.BTCDelegation) {
    // get cosmos consumer ids only
    cosmosConsumerIDs, err := k.multiStakedFPCosmosConsumerIDs(ctx, btcDel.FpBtcPkList)
    if err != nil {
        panic(fmt.Errorf("failed to get cosmos consumer ids: %w", err))
    }
    
    consumerEvent, err := types.CreateActiveBTCDelegationEvent(btcDel)
    if err != nil {
        panic(fmt.Errorf("failed to create active BTC delegation event: %w", err))
    }
    
    // Now we only loop through cosmos consumers
    for _, consumerID := range cosmosConsumerIDs {
        // No need to check if cosmos consumer - we already know it is
        if err := k.AddBTCStakingConsumerEvent(ctx, consumerID, consumerEvent); err != nil {
            panic(fmt.Errorf("failed to add active BTC delegation event: %w", err))
        }
    }
}
```

KonradStaniec: In general I agree we can further optimize it, though I would say this is for subsequent pr's (after we also clean up the situation with large events)

In general,  `AddBTCStakingConsumerEvents`  should maintain checks added int this pr either way, as otherwise it is to easy to add events for other consumers than cosmos accidentally.

gusin13: yep agree 👌 , the checks in the events fn are needed anyways. 
lets track the pending/future items in a separate ticket then

KonradStaniec: added ticket: https://github.com/babylonlabs-io/babylon/issues/1313

PR #1300, opened 2025-07-06
refactor: x/btcstaking handles fp ops

vitsalis:
At the moment, finality provider operations are handled both in the x/btcstaking and x/btcstkconsumer modules. This is not very effective for the following reasons:
* Complex logic: For every finality provider operation, we have to check whether this is a Babylon Genesis fp or a consumer BSN fp and route to the appropriate model.
* Duplicated logic: Many data structures need to be duplicated leading to additional stores (3 new stores for fps on the x/btcstkconsumer) or the reusage of internal module data structures (such as the x/btcstaking FinalityProvider) across different modules.

To resolve these considerations, this PR transfers all finality provider logic to the x/btcstaking module, while allowing the x/btcstkconsumer module handle only consumer BSN registration tracking operations. Below we outline the major modifications of the PR:

* **Message flow**: x/btcstaking `MsgCreateFinalityProvider` now handles all finality provider registrations instead of forwarding consumer BSN ones to the x/btcstkconsumer module. The corresponding fuzz test has been updated to reflect the full test case.
* **Storage**: the following storage modifications wewre made:
  * Removed the `fpPk => bsnId` mapping from the x/btcstkconsumer storage.
  * Consumer finality providers are now stored as their Babylon Genesis peers on x/btcstaking.
  * To facilitate "Give me all fps securing a BSN ID" queries (which are common ones) we introduce a secondary index for fast lookup by BSN ID: `bsn_id || fp_btc_pk -> nil (just existence marker)`.
    * This allows for limited overhead for getting all `fp_btc_pk` for a `bsn_id` using an index store.
* **Queries**: Moved all x/btcstkconsumer finality provider queries to x/btcstaking.
  * This only required the modification of the `QueryFinalityProviders` query to now support querying by BSN ID.

This PR also comments out the finality contract e2e test as it fails due to the interfaces undergoing a major change.

**TODOs** (after the PR is merged):
* **Replace chain ID-based FP type detection**: Instead of using `fp.BsnId == ctx.ChainID()` to identify Babylon genesis FPs, implement a more explicit method (e.g., a dedicated field, method like `fp.IsConsumerFP()`, or constants). Nonetheless we'll need an upgrade handler. #1306 
* **Unify finality provider logic**: Audit and consolidate the scattered type-specific logic in x/btcstaking (identify specific methods/files where Babylon vs Consumer FP separation exists and determine if it's necessary) #1309 
* **Expand consumer FP test coverage**: Most of the fuzz tests now work with Babylon Genesis fps. We can try to expand our test suite to have more cases that contain a mix of Babylon Genesis and consumer BSN fps. #1310 
* **Update documentation**: The `x/btcstaking` module documentation should now be updated to reflect the increased responsibilities of the `x/btcstaking` module. #1307 
* **Collections usage**: Examine using a collection for keys as per the comment [here](https://github.com/babylonlabs-io/babylon/pull/1300#discussion_r2188791383). Nonetheless we'll need an upgrade handler for that. #1308 

We can create github issues for the above once we are in alignment.

P.S. this is a pretty big change that can lead to lots of conflicts, so I suggest we prioritize its review. If we agree on this direction, we can address some of the comments/changes in subsequent smaller PRs.

### DIFF
--- a/x/btcstaking/keeper/finality_providers.go
+++ b/x/btcstaking/keeper/finality_providers.go
@@ -66,7 +66,6 @@ func (k Keeper) AddFinalityProvider(goCtx context.Context, msg *types.MsgCreateF
 	k.bsnIndexFinalityProvider(ctx, &fp)
 
 	// Create BTC Staking Consumer Event for the new finality provider
-	// TODO: examine whether these events are really needed
 	if !fp.SecuresBabylonGenesis(ctx) {
 		if err := k.AddBTCStakingConsumerEvent(ctx, fp.BsnId, types.CreateNewFinalityProviderEvent(&fp)); err != nil {
 			return err
@@ -84,6 +83,8 @@ func (k Keeper) setFinalityProvider(ctx context.Context, fp *types.FinalityProvi
 	store.Set(fp.BtcPk.MustMarshal(), fpBytes)
 }
 
+// TODO: Examine collections and upgrade handler for this store
+// https://github.com/babylonlabs-io/babylon/issues/1308
 func (k Keeper) bsnIndexFinalityProvider(ctx context.Context, fp *types.FinalityProvider) {
 	indexStore := k.finalityProviderBsnIndexStore(ctx)
 	bsnKey := types.BuildBsnIndexKey(fp.BsnId, fp.BtcPk)
--- a/x/btcstaking/types/finality_provider.go
+++ b/x/btcstaking/types/finality_provider.go
@@ -5,6 +5,7 @@ import sdk "github.com/cosmos/cosmos-sdk/types"
 // TODO: Rethink how to check whether an fp is a Babylon Genesis fp or not.
 // Checking through the ChainID set by context is very brittle and will
 // certainly lead to issues down the road.
+// https://github.com/babylonlabs-io/babylon/issues/1306
 func (fp *FinalityProvider) SecuresBabylonGenesis(ctx sdk.Context) bool {
 	return fp.BsnId == ctx.ChainID()
 }
--- a/x/btcstaking/README.md
+++ b/x/btcstaking/README.md
@@ -1,5 +1,8 @@
 # BTCStaking
 
+<!-- TODO: Update docs for phase-3:
+https://github.com/babylonlabs-io/babylon/issues/1307 -->
+
 The BTC staking module is responsible for maintaining the set of finality
 providers and BTC delegations under them. This includes:
 


x/btcstaking/keeper/finality_providers.go
```diff
@@ -57,10 +62,13 @@ func (k Keeper) AddFinalityProvider(goCtx context.Context, msg *types.MsgCreateF
...
-	if bsnID == ctx.ChainID() {
-		k.setFinalityProvider(ctx, &fp)
-	} else {
-		if err := k.SetConsumerFinalityProvider(ctx, &fp, bsnID); err != nil {
+	k.setFinalityProvider(ctx, &fp)
+	k.bsnIndexFinalityProvider(ctx, &fp)
+
+	// Create BTC Staking Consumer Event for the new finality provider
+	// TODO: examine whether these events are really needed
+	if !fp.SecuresBabylonGenesis(ctx) {
```
SebastianElvis, line 70: I'm not sure if it's a good idea to refer to "Genesis" in the codebase. It refers to Babylon Genesis mainnet but the codebase is generic to any Babylon network, e.g., testnets

x/btcstaking/types/finality_provider.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// TODO: Rethink how to check whether an fp is a Babylon Genesis fp or not.
+// Checking through the ChainID set by context is very brittle and will
+// certainly lead to issues down the road.
+func (fp *FinalityProvider) SecuresBabylonGenesis(ctx sdk.Context) bool {
```
SebastianElvis, line 8: The other way (that is less brittle) is to leave `BsnID` empty for Babylon FPs in KV store. This at least avoids bugs that exploit chain ID in context.

x/btcstaking/keeper/finality_providers.go
```diff
@@ -76,6 +84,12 @@ func (k Keeper) setFinalityProvider(ctx context.Context, fp *types.FinalityProvi
 	store.Set(fp.BtcPk.MustMarshal(), fpBytes)
 }
 
+func (k Keeper) bsnIndexFinalityProvider(ctx context.Context, fp *types.FinalityProvider) {
+	indexStore := k.finalityProviderBsnIndexStore(ctx)
+	bsnKey := types.BuildBsnIndexKey(fp.BsnId, fp.BtcPk)
+	indexStore.Set(bsnKey, []byte{}) // Empty value, just for existence
+}
```
SebastianElvis, line 91: To achieve the goal "given BSN ID, output list of BTC PKs of FPs securing this BSN", I think we can use the `KeySet` type provided since Cosmos SDK v0.50. See an example [here](https://github.com/babylonlabs-io/babylon/blob/d253be02ef2fb8443f01760f6714064003d6b469/x/incentive/keeper/keeper.go#L24-L26). This will be much cleaner compared to the old manner prior to v0.50

x/btcstaking/keeper/grpc_query.go
```diff
@@ -245,6 +229,34 @@ func (k Keeper) LargestBtcReOrg(ctx context.Context, _ *types.QueryLargestBtcReO
 	}, nil
 }
 
+// FinalityProviderBSNId returns the BSN ID for the finality provider with the specified finality provider BTC PK
+func (k Keeper) FinalityProviderBSNId(ctx context.Context, req *types.QueryFinalityProviderBSNIdRequest) (*types.QueryFinalityProviderBSNIdResponse, error) {
```
SebastianElvis, line 233: Do we still need this query given that the FP object contains BSN ID as a field?

SebastianElvis (commented):
Much cleaner! Some initial comments

x/btcstaking/keeper/finality_providers_test.go
SebastianElvis, line 1: Do we want to migrate unit tests here to `x/btcstaking` module as well?

x/btcstaking/keeper/grpc_query_test.go
SebastianElvis, line 1: Same here, also some of the BSN FP unit tests might not be relevant anymore

x/btcstaking/keeper/multistaking_test.go
SebastianElvis, line 1: I saw occurence of restaking in this file, maybe good opportunity to change it as well?

x/btcstaking/keeper/finality_providers.go
vitsalis: Babylon Genesis is the network operated by the node in this repository. Babylon is the protocol. For example, our testnet is the Babylon Genesis testnet, not the Babylon testnet. In the long run, we should aim to start using "Babylon Genesis" more to refer to the chain, as just "Babylon" is not correct terminology.

Nonetheless, if you see my comment on the `SecuresBabylonGenesis` function, it is a temporary one that should be replaced by a better heuristic for identifying whether a finality provider secures genesis or not.

x/btcstaking/types/finality_provider.go
vitsalis: Yep, we can examine this in a separate PR. Leaving this as is now as it requires discussion.

x/btcstaking/keeper/grpc_query_test.go
vitsalis: Which tests are you referring to?

x/btcstaking/keeper/grpc_query.go
vitsalis: Will remove

x/btcstaking/keeper/finality_providers.go
vitsalis: I'll add it as a TODO for after the PR is merged

client/query/btcstaking.go
```diff
@@ -66,11 +66,12 @@ func (c *QueryClient) FinalityProvider(fpBtcPkHex string) (*btcstakingtypes.Quer
 }
 
 // FinalityProviders queries the BTCStaking module for all finality providers
-func (c *QueryClient) FinalityProviders(pagination *sdkquerytypes.PageRequest) (*btcstakingtypes.QueryFinalityProvidersResponse, error) {
+func (c *QueryClient) FinalityProviders(bsnId string, pagination *sdkquerytypes.PageRequest) (*btcstakingtypes.QueryFinalityProvidersResponse, error) {
```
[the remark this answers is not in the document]
vitsalis: Will do.

x/btcstaking/keeper/finality_providers.go
```diff
@@ -45,6 +45,11 @@ func (k Keeper) AddFinalityProvider(goCtx context.Context, msg *types.MsgCreateF
 		// Babylon chain ID
 		bsnID = ctx.ChainID()
```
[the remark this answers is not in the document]
vitsalis: Let's discuss this in a separate issue? I have a TODO for this.

vitsalis:
> One question given that we have multiple todos added, wdyt about addint gh issues for all the missing pieces and then referencing them in code ?

Done for the ones referring to the codebase and belonging to a single place. A couple general ones I did not add.

KonradStaniec (approved):
Looks good ! thanks

x/btcstaking/types/finality_provider.go
SebastianElvis: >  though this will influence the indext bsnId -> fkpKeys

Babylon FP shall not be in this index, right?

PR #1304, opened 2025-07-07
feat/rollup finality e2e

maurolacy:
Update rollup e2e tests.


SebastianElvis:
Thanks for the PR! 

However, I thought we have [decided](https://github.com/babylonlabs-io/babylon/issues/1298) to move the e2e tests to https://github.com/babylonlabs-io/rollup-bsn-contracts?

maurolacy:
Sure, I can port these changes there, after you finish your PR, and it's merged.

maurolacy:
After offline talk, it seems changes are already there, so, no need for backporting this.

I think that we can still merge this here, for reference, and until we decide to remove these tests from Babylon. Tests are currently disabled, so merging will not affect / impact anything. WDYT?

SebastianElvis (approved):

vitsalis:
> After offline talk, it seems changes are already there, so, no need for backporting this.
> 
> I think that we can still merge this here, for reference, and until we decide to remove these tests from Babylon. Tests are currently disabled, so merging will not affect / impact anything. WDYT?

Shall we follow up this PR with one that deletes the tests from Babylon if they are already in the rollup BSN contract?

maurolacy:
I think we should keep them for some time. The approach is different, and they can serve as a reference, at least until the other tests are stable / mature.

PR #1326, opened 2025-07-09
fix: pubrand ts e2e

maurolacy:
A small follow-up to #1304.

If we're going to keep (however disabled) these tests for now, let's at least make sure that they pass in their current version.


vitsalis (approved):
lgtm!

PR #1328, opened 2025-07-09
feat: global limit for multi staked fps

KonradStaniec:
- Adds `max_finality_providers` param to the stkaing params
- parameter defines maxium amount of finality provider keys in script
- add migration of old param version to set it to `1`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,7 +116,7 @@ when the babylon block is being BTC rewarded.
 - [#1296](https://github.com/babylonlabs-io/babylon/pull/1296) IBC events are only
 queued for cosmos bsns
 - [#1328](https://github.com/babylonlabs-io/babylon/pull/1328) Add global limit
-for mulistaking fps
+for multistaking fps
 
 ### Bug fixes
 
--- a/x/btcstaking/keeper/params.go
+++ b/x/btcstaking/keeper/params.go
@@ -216,20 +216,3 @@ func (k Keeper) GetParamsForBtcHeight(ctx context.Context, height uint64) (*type
 
 	return k.GetParamsByVersion(ctx, version), version, nil
 }
-
-func (k Keeper) IterateAndModifyParams(ctx context.Context, f func(ctx context.Context, p *types.Params) error) error {
-	paramsStore := k.paramsStore(ctx)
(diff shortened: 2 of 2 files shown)


x/btcstaking/types/validate_parsed_message.go
```diff
@@ -28,6 +28,12 @@ func ValidateParsedMessageAgainstTheParams(
 			pm.UnbondingTime, parameters.UnbondingTimeBlocks)
 	}
 
+	// 2. Validate finality provider is below or equal to max finality providers
+	if pm.FinalityProviderKeys.Len() > int(parameters.MaxFinalityProviders) {
```
GAtom22, line 32: this is a backward incompatible change, no?
'Cause while syncing a node, this will always fail. Should we add some handling for this?
e.g. if MaxFinalityProvivers == 0, then skip this check

x/btcstaking/keeper/migrations.go
```diff
@@ -0,0 +1,31 @@
...
+func NewMigrator(keeper Keeper) Migrator {
+	return Migrator{
+		keeper: keeper,
+	}
+}
+
+// Migrate1to2 migrates from version 1 to 2.
+func (m Migrator) Migrate1to2(ctx sdk.Context) error {
+	store := runtime.KVStoreAdapter(m.keeper.storeService.OpenKVStore(ctx))
+	return v2.MigrateStore(ctx, store, m.keeper.cdc, func(ctx context.Context, p *types.Params) error {
```
RafilxTenfen, line 27: should we use `IterateAndModifyParams` or delete the unused func

RafilxTenfen (approved):
LGTM, a few question/suggestion

x/btcstaking/types/validate_parsed_message.go
RafilxTenfen: I think this is a state breaking change and should be released in a coordinated upgrade, so it could be incompatible

gusin13: shouldn't we perform this check [at this line](https://github.com/babylonlabs-io/babylon/blob/5fffaac09b8c6072d99b8b7c9aa2f6c292b431a5/x/btcstaking/keeper/btc_delegations.go#L52) or before this

Issue #1302, opened 2025-07-07
Synchronisation of BTC light client between Babylon and Babylon contract


gitferry:
[Context](https://github.com/babylonlabs-io/babylon/issues/1240#issuecomment-3043224224)

PR #1331, opened 2025-07-10
feat(zoneconcierge): BSN Header synchronization 

Vvaradinov:
New flow optimized for BSN header synchronization:

BSN connects --> Babylon sets BSN base header to current tip
Babylon sends headers from base to tip --> BSN processes them
If reorg occurs --> Babylon falls back to BSN base header

The implementation also includes BSNs updating their base header via IBC packet for more flexible synchronization.

Closes - https://github.com/babylonlabs-io/babylon/issues/1302


x/zoneconcierge/keeper/ibc_packet_btc_staking_consumer_event.go
```diff
@@ -184,3 +196,38 @@ func (k Keeper) HandleConsumerSlashing(
...
+	// Get the client ID from the channel to identify the BSN
+	clientID, _, err := k.channelKeeper.GetChannelClientState(ctx, portID, channelID)
+	if err != nil {
+		return fmt.Errorf("failed to get client state: %w", err)
+	}
+
+	// Validate the base BTC header
+	if baseBTCHeader == nil {
+		return fmt.Errorf("base BTC header is nil")
+	}
```
gitferry, line 217: Should we call `baseBTCHeader.Validate()` here?

x/zoneconcierge/keeper/ibc_packet_btc_headers.go
```diff
@@ -18,35 +19,47 @@ func (k Keeper) BroadcastBTCHeaders(ctx context.Context) error {
...
-		k.Logger(sdkCtx).Info("skipping BTC header broadcast",
-			"reason", "no headers to broadcast",
-		)
-		return nil
-	}
+	// New behavior using Consumer-specific base headers:
+	// - If no Consumer base header exists: fallback to sending k+1 from tip
+	// - If Consumer base header exists but no headers sent: from Consumer base to tip
+	// - If headers previously sent: from child of most recent valid header to tip
+	// - If reorg detected: from Consumer base to tip
```
gitferry, line 26: I feel like it's not efficient to handle reorg here. We probably need to send a reorg event and send from the reorged header to tip? Maybe a TODO for now

x/zoneconcierge/keeper/ibc_packet_btc_headers.go
```diff
@@ -18,35 +19,47 @@ func (k Keeper) BroadcastBTCHeaders(ctx context.Context) error {
...
+
+		headers := k.getHeadersToBroadcastForConsumer(ctx, consumerID)
+		if len(headers) == 0 {
+			k.Logger(sdkCtx).Debug("skipping BTC header broadcast for consumer, no headers to broadcast",
+				"consumerID", consumerID,
+			)
+			continue
+		}
+
+		packet := types.NewBTCHeadersPacketData(&types.BTCHeaders{Headers: headers})
```
gitferry, line 45: Do we have a limit of headers for a packet?

gitferry (commented):
Sorry for the delay. Some comments. Also better to have @SebastianElvis 's review

x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
```diff
@@ -148,10 +148,47 @@ func (k Keeper) getDeepEnoughBTCHeaders(ctx context.Context) []*btclctypes.BTCHe
...
+
+	// If no Consumer base header exists, fallback to the old behavior
+	if baseHeader == nil {
+		return k.getHeadersToBroadcast(ctx)
+	}
+
+	// If we haven't sent any headers yet, send from Consumer base to tip
+	if lastSegment == nil {
+		return k.btclcKeeper.GetMainChainFrom(ctx, baseHeader.Height+1)
+	}
```
SebastianElvis, line 169: It might be possible that the BSN chooses a super old base BTC header such that L168 will get a large number of BTC headers. This might cause unexpected behaviours.

Since it's BSN's responsibility to pick a recent header, what do you think we add a restriction here, e.g., if `tip.Height - baseHeader.Height > k` then we return error?

x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
```diff
@@ -148,10 +148,47 @@ func (k Keeper) getDeepEnoughBTCHeaders(ctx context.Context) []*btclctypes.BTCHe
...
+		if header, err := k.btclcKeeper.GetHeaderByHash(ctx, header.Hash); err == nil && header != nil {
+			initHeader = header
+			break
+		}
+	}
+
+	// If no header from last segment is still valid (reorg), send from Consumer base to tip
+	if initHeader == nil {
+		return k.btclcKeeper.GetMainChainFrom(ctx, baseHeader.Height+1)
+	}
```
SebastianElvis, line 184: Shall we send from base to tip or the last `k` headers in this particular case?

x/zoneconcierge/keeper/ibc_packet_btc_headers.go
```diff
@@ -18,35 +19,47 @@ func (k Keeper) BroadcastBTCHeaders(ctx context.Context) error {
...
 
-	k.setLastSentSegment(ctx, &types.BTCChainSegment{
-		BtcHeaders: headers,
-	})
+	// Update last sent segment if any headers were sent
+	if lastHeadersSent != nil {
+		k.setLastSentSegment(ctx, &types.BTCChainSegment{
+			BtcHeaders: lastHeadersSent,
+		})
+	}
```
SebastianElvis, line 62: Hmm this is a bit tricky -- Previously for every BSN the lastSentSegment is same, but after introducing the concept of base header, this is not always the case anymore. So the current code will always use the last sent segment of the last IBC channel.

One approach of fixing this is to extend this KV store to have `consumerID -> BTCChainSegment` structure so that it remembers last sent segment for every consumer ID

SebastianElvis: This KV store can also be used for saving the base BTC headers for BSNs (i.e., we can merge `consumer_base_btc_header.go` storage to this storage)

SebastianElvis (commented):
Great work, and agree with the current direction :+1:

x/zoneconcierge/keeper/ibc_packet_btc_headers.go
Vvaradinov: There is no explicit header count limit. Should we have one ?

gitferry (approved):
Lgtm!

PR #1344, opened 2025-07-11
fix: bump cosmos-sdk and remove send restrictions

KonradStaniec:
- new cosmos-sdk version contains fix to `x/distribution` module (https://github.com/cosmos/cosmos-sdk/security/advisories/GHSA-p22h-3m2v-cmgh) which we originally fixed by introducing bank restrictions
- with the fix bank restrictions can be removed


RafilxTenfen (approved):
LGTM

Issue #1312, opened 2025-07-07
epoching: enrich QueuedMessage API


SebastianElvis:
`msg`  field ends up being of an identical encoding structure between delegation and undelegation messages. For example, from this epoch on testnet 3852:

This is a [delegation message](https://www.mintscan.io/babylon-testnet/tx/37c4f4f813d2061f81c8b0e414fe0e7c4a2a839ffcbd102e50c853d691dbe33f) coming from the endpoint:

```
{
      "tx_id": "37c4f4f813d2061f81c8b0e414fe0e7c4a2a839ffcbd102e50c853d691dbe33f",
      "msg_id": "05be0b6753fd91ae792ef41cea02fb2618bea9fdcd1d207d537b699ad83f485c",
      ...
      "msg": "delegator_address:\"bbn1045lg4c4an0sdem2krhxatqg604amvnmxn92mg\" validator_address:\"bbnvaloper109x4ruspxarwt62puwcenhclw36l9v7j92f0ex\" amount:<denom:\"ubbn\" amount:\"3000000\" > "
 },
```

This is an [undelegation message](https://www.mintscan.io/babylon-testnet/tx/62936d53b755a9485e43a2e8e1dab78f4c9ba446c834248e262c3bdcbaaeffc4) coming from the endpoint:

```
{
      "tx_id": "62936d53b755a9485e43a2e8e1dab78f4c9ba446c834248e262c3bdcbaaeffc4",
      "msg_id": "24ad035b4075b89404526bd2abc766f849cdf88edf1c385c66686c2fa08ed335",
      ...
      "msg": "delegator_address:\"bbn1045lg4c4an0sdem2krhxatqg604amvnmxn92mg\" validator_address:\"bbnvaloper109x4ruspxarwt62puwcenhclw36l9v7j92f0ex\" amount:<denom:\"ubbn\" amount:\"1000000\" > "
}
```

They're identical in structure because MsgDelegate  and MsgUndelegate  have the same shape.

To fix this, Babylon side needs to enrich the query to add a field identifying the message type

PR #1347, opened 2025-07-11
chore(epoching): enhance the QueuedMsgs API 

samricotta:
closes: https://github.com/babylonlabs-io/babylon/issues/1312

Adds a new enriched_msg field to `QueuedMessageResponse`, allowing clients to distinguish between messages like `MsgDelegate` and `MsgUndelegate` even if their raw msg field looks identical.

### DIFF
--- a/x/epoching/types/query_test.go
+++ b/x/epoching/types/query_test.go
@@ -92,7 +92,6 @@ func TestQueuedMessage_ToResponse(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			if tc.expectPanic {
 				require.Panics(t, func() {
-					ch
 					_ = tc.inputQueuedMessage.ToResponse()
 				})
 				return


proto/babylon/epoching/v1/query.proto
```diff
@@ -224,6 +225,21 @@ message QueuedMessageResponse {
   // msg is the actual message that is sent by a user and is queued by the
   // epoching module as string.
   string msg = 5;
+  // enriched_msg contains a parsed, structured version of the original message
+  EnrichedMsg enriched_msg = 6;
```
KonradStaniec, line 229: I think additional field could be something like `string queued_msg_type = 6` i.e we do not need the whole structure here but only full name of the type of the message being queued.

x/epoching/types/query.go
```diff
@@ -18,12 +20,39 @@ func (e *Epoch) ToResponse() *EpochResponse {
 
 // ToResponse parses a QueuedMessage into a query response queued message struct.
 func (q *QueuedMessage) ToResponse() *QueuedMessageResponse {
+	var enriched *EnrichedMsg
+	sdkMsg := q.UnwrapToSdkMsg()
+
+	switch msg := sdkMsg.(type) {
```
KonradStaniec, line 26: I think we need to recover type for every possible message that can queued, not only `MsgDelegate` and `MsgDelegate`

proto/babylon/epoching/v1/query.proto
```diff
@@ -224,6 +225,8 @@ message QueuedMessageResponse {
   // msg is the actual message that is sent by a user and is queued by the
   // epoching module as string.
   string msg = 5;
+  // enriched_msg contains a parsed, structured version of the original message
+  string queued_msg_type = 6;
```
SebastianElvis, line 229: Docstring is not consistent with field. Also maybe we can simply call it `msg_type`

SebastianElvis (approved):

KonradStaniec (approved):
Minor comments ! lgtm!

Lazar955 (approved):

PR #1358, opened 2025-07-14
chore (btcstaking): simplify selective slashing message handler 

samricotta:
closes: https://github.com/babylonlabs-io/pm/issues/445

- Removes `staking_tx_hash` from `MsgSelectiveSlashingEvidence` and the internal Evidence struct.
- Refactors the handler to validate the slash using only the provided secret key to find the Finality Provider.
- Simplifies the `fuzz` test to remove `BTCDelegation` setup and directly test the new logic.


x/btcstaking/types/msg_test.go
```diff
@@ -405,468 +389,3 @@ func TestStructFieldConsistency(t *testing.T) {
 		t.Errorf("MsgCreateBTCDelegation is missing fields (except final 2) from MsgBtcStakeExpand: %v", missingFromCreate)
 	}
 }
-
-func TestMsgAddBsnRewardsValidateBasic(t *testing.T) {
```
KonradStaniec, line 409: we should not remove those tests (probably some conflict resolution error)

KonradStaniec (approved):

PR #1371, opened 2025-07-15
fix: register tokenfactory bindings

benluelo:
attempted to interact with the tokenfactory module via cosmwasm and got this response:

> failed to execute message; message index: 0: dispatch: no handler found: unknown message from the contract [CosmWasm/wasmd@v0.54.1/x/wasm/keeper/handler_plugin.go:173]

<del>i'm not sure if this is the best way to register the module, nor if it actually works correctly as i cannot test it locally, but the node builds and runs 👍</del> update: i was able to get a babylond localnet running and can confirm this works correctly

### DIFF
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -85,6 +85,7 @@ import (
 	ibccallbacks "github.com/cosmos/ibc-go/v10/modules/apps/callbacks"
 	ibcexported "github.com/cosmos/ibc-go/v10/modules/core/exported"
 	ibckeeper "github.com/cosmos/ibc-go/v10/modules/core/keeper"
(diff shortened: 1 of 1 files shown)


SebastianElvis (approved):
Good catch! Will wait for another approval from @babylonlabs-io/core-dev

SebastianElvis:
I added backport label for v3, do we want to backport this to v2 so that v2 supports tokenfactory wasm bindings? @babylonlabs-io/core-dev

KonradStaniec:
Hey @benluelo thanks for contribution! I will close this one in favour of https://github.com/babylonlabs-io/babylon/pull/1398 (it is much quicker to go over all the backports if pr is made from account with write access)

PR #1381, opened 2025-07-16
feat: add multi-staking allow-list logic

GAtom22:


x/btcstaking/keeper/allowed_transaction_index.go
```diff
@@ -22,3 +22,35 @@ func (k Keeper) IsStakingTransactionAllowed(ctx context.Context, txHash *chainha
...
+func (k Keeper) IndexAllowedMultiStakingTransaction(ctx context.Context, txHash *chainhash.Hash) {
+	err := k.allowedMultiStakingTxHashesKeySet.Set(ctx, txHash[:])
+	if err != nil {
+		panic(err) // encoding issue; this can only be a programming error
+	}
+}
+
+// IsMultiStakingAllowed checks if the given staking transaction is elegible
+// for multi-staking.
+func (k Keeper) IsMultiStakingAllowed(ctx context.Context, txHash *chainhash.Hash) (bool, error) {
```
RafilxTenfen, line 37: :100:

x/btcstaking/migrations/v2/store.go
```diff
@@ -16,7 +17,21 @@ func MigrateStore(
...
 	c codec.BinaryCodec,
-	modifyParams func(ctx context.Context, p *types.Params) error) error {
+	modifyParams func(ctx context.Context, p *types.Params) error,
+	indexTxHash func(ctx context.Context, txHash *chainhash.Hash),
+) error {
+	if err := migrateParams(ctx, s, c, modifyParams); err != nil {
+		return err
+	}
+
+	if err := indexAllowedMultiStakingTxs(ctx, indexTxHash); err != nil {
```
RafilxTenfen, line 27: I don't know if this should be a migration, in my mind would be something we would input in the upgrade handler, same way it was done at v1

RafilxTenfen (approved):
The multistaking allowlist handling and verification is pretty clean and look good to me, but in my mind we would insert the allow list in the upgrade handler and not by store migrations

This could also be done in another PR after we have the v3 upgrade setup

x/btcstaking/keeper/btc_delegations.go
```diff
@@ -71,6 +71,28 @@ func (k Keeper) CreateBTCDelegation(ctx sdk.Context, parsedMsg *types.ParsedCrea
 		}
 	}
 
+	// Check multi-staking allow list
+	// During multi-staking allow-list period, only existing BTC delegations
+	// in the allow-list can become multi-staked via stake expansion or
+	// already existing multi-staking delegation (extended from the allow-list)
+	isMultiStaking := len(parsedMsg.FinalityProviderKeys.PublicKeysBbnFormat) > 1
```
KonradStaniec, line 78: you can call `parsedMsg.FinalityProviderKeys.Len()`

x/btcstaking/types/multi_staking_allow_list_placeholder.go
```diff
@@ -0,0 +1,19 @@
...
+package types
+
+// This is a placeholder for the multi-staking allow list and allow list expiration height.
+// Values here are not used in production and are only for testing purposes.
+// The actual allow list is defined in the files with build tags:
+// multi_staking_allow_list_mainnet.go and multi_staking_allow_list_testnet.go
+// Make sure to include the correct tx hashes for the
+// mainnet or testnet as needed in the corresponding
+// files with the build tags
+const multiStakingAllowListExpirationHeight = 5
```
KonradStaniec, line 12: maybe we can leave default as `0` i.e disabled allow list ? this way it won't accidentally crash some e2e tests in repos depending on babylon.

x/btcstaking/types/multi_staking_allow_list_testnet.go
```diff
@@ -0,0 +1,16 @@
+//go:build testnet
+
+package types
+
+// TODO TBD
+const multiStakingAllowListExpirationHeight = 0
+
```
KonradStaniec, line 7: maybe we can have separate package like `types/allowlist` for all those hardcoded files ? The introduce a lot of noise in `types` package

x/btcstaking/migrations/v2/store.go
KonradStaniec: In general, if we did not have migration already I would agree with this i.e putting it in upgrade handler would be cleaner.

But as we are doing the migration either way, imo, doing it here nicely hides this whole thing from the rest of the app (functions can be private etc.)

x/btcstaking/types/multi_staking_allow_list_test.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"testing"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/stretchr/testify/require"
+
+	"github.com/babylonlabs-io/babylon/v3/x/btcstaking/types"
+)
+
+func TestLoadMultiStakingAllowList(t *testing.T) {
```
KonradStaniec, line 12: Ideally we would have some test checking that neither of allowlist (mainnet, testnet, default) contains duplicates as duplicate can mean that somebody did make mistake while creating the allowlist 🤔

x/btcstaking/types/multi_staking_allow_list_placeholder.go
GAtom22: But while building the docker image, will pick up the mainnet values by default. Unless we change the flag for building the docker image with the desired values

x/btcstaking/types/multi_staking_allow_list_testnet.go
GAtom22: done

PR #1388, opened 2025-07-17
chore: state migration to support transition from v2.2 to v3

samricotta:
closes: https://github.com/babylonlabs-io/babylon/issues/280

Updates v3 handler with e2e tests, closed the other PR as this has both

Note: This transition is for the current time from v2.2 - v3. If there are any additional changes they will need to be introduced later in another PR

### DIFF
--- a/test/e2e/software_upgrade_e2e_v3_test.go
+++ b/test/e2e/software_upgrade_e2e_v3_test.go
@@ -154,9 +154,10 @@ func (s *SoftwareUpgradeV3TestSuite) TestUpgradeV3() {
 	resp := n.QueryAppliedPlan(v3.UpgradeName)
 	s.EqualValues(expectedUpgradeHeight, resp.Height, "the plan should be applied at the height %d", expectedUpgradeHeight)
 
-	// check fps have the same chain id
-	s.Require().Equal(n.ChainID(), s.fp1.BsnId)
-	s.Require().Equal(n.ChainID(), s.fp2.BsnId)
+	fp1 := n.QueryFinalityProvider(s.fp1.BtcPk.MarshalHex())
+	s.Require().Equal(fp1.BsnId, n.ChainID())
+	fp2 := n.QueryFinalityProvider(s.fp2.BtcPk.MarshalHex())
+	s.Require().Equal(fp2.BsnId, n.ChainID())
 
 	// query pub randomness
 	fp1CommitPubRand := n.QueryListPubRandCommit(s.fp1CommitPubRandList.FpBtcPk)
--- a/test/e2e/govProps/v3.json
+++ b/test/e2e/govProps/v3.json
@@ -6,7 +6,7 @@
       "plan": {
         "name": "v3",
         "time": "0001-01-01T00:00:00Z",
-        "height": "37",
+        "height": "34",
         "info": "Upgrade to v3",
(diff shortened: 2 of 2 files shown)


RafilxTenfen:
is this a PR for us to keep changing as needed and should contain all the necessary migrations + possible allow list inclusion
or should we review and merge ? @samricotta

KonradStaniec:
> is this a PR for us to keep changing as needed and should contain all the necessary migrations + possible allow list inclusion
or should we review and merge ? @samricotta

Lets verify that the upgrade handler is correct at current time i.e it is proper transition from v2.2 to v3, if there will be changes introduced by allowlist or any other pr, they should be added approprietly

test/e2e/containers/config.go
```diff
@@ -17,8 +17,8 @@ const (
 	// name of babylon image produced by running `make build-docker`
 	BabylonContainerName = "babylonlabs-io/babylond"
 	// name of babylon image before the upgrade
-	BabylonContainerNameBeforeUpgrade = "babylonlabs/babylond"
-	BabylonContainerTagBeforeUpgrade  = "v1.0.1"
+	BabylonContainerNameBeforeUpgrade = "babylonlabs-io/babylond"
```
RafilxTenfen, line 20: I think the name in docker hub is still `babylonlabs/babylond`
https://hub.docker.com/r/babylonlabs/babylond

test/e2e/software_upgrade_e2e_v3_test.go
```diff
@@ -0,0 +1,228 @@
...
+func (s *SoftwareUpgradeV3TestSuite) SetupSuite() {
+	s.T().Skip("Temporarily skipping v3 upgrade test")
+	s.T().Log("setting up e2e integration test suite for v2.2.0 to v3 upgrade...")
+	var err error
+	s.balancesBeforeUpgrade = make(map[string]sdk.Coin)
+
+	btcHeaderGenesis := sample.SignetBtcHeader195552(s.T())
+
+	// func runs right before the upgrade proposal is sent
+	preUpgradeFunc := func(chains []*chain.Config) {
```
KonradStaniec, line 52: all pre upgrade state should happen in this `preUpgradeFunc` i.e if we want to check whther finality providers were migrated correctly: 
- we should create them here in this callback 
- later in `TestUpgradeV3`  check that they have valid chain-id

Basically, all things that happen in `TestUpgradeV3` test, are already after upgrade happens

test/e2e/software_upgrade_e2e_v3_test.go
```diff
@@ -0,0 +1,221 @@
...
+	fp2BTCSK              *btcec.PrivateKey
+	fp1                   *bstypes.FinalityProvider
+	fp2                   *bstypes.FinalityProvider
+	fp1CommitPubRandList  *types.MsgCommitPubRandList
+	fp2CommitPubRandList  *types.MsgCommitPubRandList
+	commitStartHeight     uint64
+}
+
+func (s *SoftwareUpgradeV3TestSuite) SetupSuite() {
+	s.T().Skip("Temporarily skipping v3 upgrade test")
```
RafilxTenfen, line 47: Was this skip on purpose? only noticed while doing https://github.com/babylonlabs-io/babylon/pull/1473

PR #1390, opened 2025-07-18
feat: additional gas per multistaked fp

KonradStaniec:
- adds additional gas cost per multistaked fp `GasCostPerMultiStakedFP = 7000`
- value calculated based on covenant message cost when adding covenant signatures


RafilxTenfen (approved):
LGTM

PR #1404, opened 2025-07-22
chore: refactor power dist event processing

GAtom22:
[RFC](https://github.com/babylonlabs-io/pm/blob/main/rfc/rfc-040-power-dist-event-processing.md)

Closes https://github.com/babylonlabs-io/pm/issues/400 
Closes https://github.com/babylonlabs-io/pm/issues/398

### DIFF
--- a/x/finality/keeper/power_dist_change.go
+++ b/x/finality/keeper/power_dist_change.go
@@ -180,9 +180,9 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 		k.processEventsAtHeight(ctx, sdkCtx, btcHeight, state)
 	}
 
-	// Process deferred EXPIRED events
+	// Process events for terminal states (EXPIRED btc delegations and SLASHED finality providers)
 	k.processExpiredEvents(ctx, sdkCtx, state)
-
+	processSlashedEvents(sdkCtx, state)
 	/*
 		At this point, there is voting power update.
 		Then, construct a voting power dist cache by reconciling the previous
@@ -323,9 +323,12 @@ func (k Keeper) processEventsAtHeight(ctx context.Context, sdkCtx sdk.Context, b
 				// Process ACTIVE/UNBONDED events immediately
 				k.processBtcDelUpdateImmediate(ctx, state, typedEvent)
 			}
+		case *types.EventPowerDistUpdate_SlashedFp:
+			// Defer SLASHED events for later processing
+			state.SlashedEvents = append(state.SlashedEvents, typedEvent)
 		default:
-			// Process all FP events immediately
-			k.processFPEventImmediate(sdkCtx, state, event)
+			// Process all other FP events immediately
+			processFPEventImmediate(sdkCtx, state, event)
 		}
 	}
 }
@@ -391,20 +394,8 @@ func (k Keeper) processExpiredEvents(ctx context.Context, sdkCtx sdk.Context, st
 	}
 }
 
-func (k Keeper) processFPEventImmediate(ctx sdk.Context, state *ftypes.ProcessingState, event types.EventPowerDistUpdate) {
+func processFPEventImmediate(ctx sdk.Context, state *ftypes.ProcessingState, event types.EventPowerDistUpdate) {
 	switch typedEvent := event.Ev.(type) {
-	case *types.EventPowerDistUpdate_SlashedFp:
-		// record slashed fps
-		types.EmitSlashedFPEvent(ctx, typedEvent.SlashedFp.Pk)
-		fpBTCPKHex := typedEvent.SlashedFp.Pk.MarshalHex()
-		state.FPStatesByBtcPk[fpBTCPKHex] = ftypes.FinalityProviderState_SLASHED
-		// TODO(rafilx): handle slashed fps prunning
-		// It is not possible to slash fp and delete all of his data at the
-		// babylon block height that is being processed, because
-		// the function RewardBTCStaking is called a few blocks behind.
-		// If the data is deleted at the slash event, when slashed fps are
-		// receveing rewards from a few blocks behind HandleRewarding
-		// verifies the next block height to be rewarded.
 	case *types.EventPowerDistUpdate_JailedFp:
 		// record jailed fps
 		types.EmitJailedFPEvent(ctx, typedEvent.JailedFp.Pk)
@@ -412,6 +403,22 @@ func (k Keeper) processFPEventImmediate(ctx sdk.Context, state *ftypes.Processin
 	case *types.EventPowerDistUpdate_UnjailedFp:
 		// record unjailed fps
 		state.FPStatesByBtcPk[typedEvent.UnjailedFp.Pk.MarshalHex()] = ftypes.FinalityProviderState_UNJAILED
+	}
+}
+
+func processSlashedEvents(ctx sdk.Context, state *ftypes.ProcessingState) {
+	for _, event := range state.SlashedEvents {
+		// record slashed fps
+		types.EmitSlashedFPEvent(ctx, event.SlashedFp.Pk)
+		fpBTCPKHex := event.SlashedFp.Pk.MarshalHex()
+		state.FPStatesByBtcPk[fpBTCPKHex] = ftypes.FinalityProviderState_SLASHED
+		// TODO(rafilx): handle slashed fps prunning
+		// It is not possible to slash fp and delete all of his data at the
+		// babylon block height that is being processed, because
+		// the function RewardBTCStaking is called a few blocks behind.
+		// If the data is deleted at the slash event, when slashed fps are
+		// receiving rewards from a few blocks behind HandleRewarding
+		// verifies the next block height to be rewarded.
 	}
 }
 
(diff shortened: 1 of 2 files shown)


testutil/btcstaking-helper/keeper.go
```diff
@@ -168,8 +193,17 @@ func NewHelperWithStoreAndIncentive(
 	btccKForBtcStaking *types.MockBtcCheckpointKeeper,
 	btccKForFinality *ftypes.MockCheckpointingKeeper,
 	ictvKeeper IctvKeeperI,
+	storeKeys ...*storetypes.KVStoreKey,
 ) *Helper {
-	k, _ := keepertest.BTCStakingKeeperWithStore(t, db, stateStore, nil, btclcKeeper, btccKForBtcStaking, ictvKeeper)
+	var btcStkStoreKey *storetypes.KVStoreKey
+	for _, k := range storeKeys {
```
RafilxTenfen, line 199: not sure if this is a good choice, I would prefer to update all the calls to call as `nil` in the ones that don't need storeKey and expect to receive the btc staking store key as parameter

but mostly your call on this one

x/finality/keeper/power_dist_change.go
```diff
@@ -396,23 +313,133 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
+		// - it has all required covenant signatures
+		hasQuorum, err := k.BTCStakingKeeper.BtcDelHasCovenantQuorums(ctx, btcDel, delParams.CovenantQuorum)
+		if err != nil {
+			panic(err)
+		}
+		if !btcDel.IsUnbondedEarly() && hasQuorum {
+			// only adds to the new unbonded list if it hasn't
+			// previously unbonded with types.BTCDelegationStatus_UNBONDED
+			k.processPowerDistUpdateEventUnbond(ctx, state, btcDel)
+		}
```
RafilxTenfen, line 399:
```suggestion
                types.EmitExpiredDelegationEvent(sdkCtx, delStkTxHash)
                if btcDel.IsUnbondedEarly() {
                   continue
                 }
		delParams := k.BTCStakingKeeper.GetParamsByVersion(ctx, btcDel.ParamsVersion)

		
		// We process expired event if:
		// - it hasn't unbonded early
		// - it has all required covenant signatures
		hasQuorum, err := k.BTCStakingKeeper.BtcDelHasCovenantQuorums(ctx, btcDel, delParams.CovenantQuorum)
		if err != nil {
			panic(err)
		}
		if hasQuorum {
			// only adds to the new unbonded list if it hasn't
			// previously unbonded with types.BTCDelegationStatus_UNBONDED
			k.processPowerDistUpdateEventUnbond(ctx, state, btcDel)
		}
```

RafilxTenfen: just to avoid checks if it was already unbonded

x/finality/types/finality.go
```diff
@@ -9,8 +9,52 @@ import (
...
+// The state is updated during the power distribution change process and is used to generate the
+// final power distribution cache
+type ProcessingState struct {
+	// FPStatesByBtcPk is a map of the finality providers' state
+	FPStatesByBtcPk map[string]FinalityProviderState
+	FpByBtcPk       map[string]*btcstktypes.FinalityProvider
+	// DeltaSatsByFpBtcPk is a map where key is finality provider's BTC PK hex and value is the
+	// delta of BTC delegations satoshis that were added or removed from the provider
+	// during the power distribution change process
+	DeltaSatsByFpBtcPk map[string]int64
```
GAtom22, line 44: int64 range is -9223372036854775808 through 9223372036854775807
considering 1 sat is 0.00000001 BTC, and there'll be only 21 M BTC, we should be OK,
as the delta int64 range in BTC would be -92,233,720,368.54775808 thru 92,233,720,368.54775808

x/finality/types/finality.go
```diff
@@ -9,8 +9,52 @@ import (
...
+// Processing state during the power distribution change process
+// It holds the state of finality providers, BTC delegations, and events
+// It is used to track the changes in the finality providers' states and the BTC delegations
+// It is also used to determine the final state of each finality provider after the power distribution change
+// The state is updated during the power distribution change process and is used to generate the
+// final power distribution cache
+type ProcessingState struct {
+	// FPStatesByBtcPk is a map of the finality providers' state
+	FPStatesByBtcPk map[string]FinalityProviderState
+	FpByBtcPk       map[string]*btcstktypes.FinalityProvider
```
KonradStaniec, line 40: I would add comment that this acts mostly as cache for  already retrieved FPs

x/finality/keeper/power_dist_change.go
```diff
@@ -396,23 +301,136 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
+	}
+}
+
+func (k Keeper) processFPEventImmediate(ctx sdk.Context, state *ftypes.ProcessingState, event types.EventPowerDistUpdate) {
+	switch typedEvent := event.Ev.(type) {
+	case *types.EventPowerDistUpdate_SlashedFp:
```
KonradStaniec, line 396: maybe lets double check that there is no operations leading to following sequences:
1. Slashed event
2. Jailed event

or

1. Slashed event
2. Unjail event

as this would mean we will not correctly process `Slashing` event  later down the line i.e we won't do
```go
		case ftypes.FinalityProviderState_SLASHED:
			// if this finality provider is slashed, continue to avoid
			// assigning delegation to it
			fp.IsSlashed = true
			continue
```
which means guy will remain part of voting `VotingPowerDistCache`

KonradStaniec (commented):
looks much better than before 👍 some questions / comments

x/finality/types/power_table.go
```diff
@@ -208,6 +208,10 @@ func (v *FinalityProviderDistInfo) AddBondedSats(sats uint64) {
 }
 
 func (v *FinalityProviderDistInfo) RemoveBondedSats(sats uint64) {
+	// safeguard against underflow in total bonded satoshis
+	if v.TotalBondedSat < sats {
```
RafilxTenfen, line 212: this should never happen, right?

GAtom22: yes, it is just a safeguard

RafilxTenfen (approved):
It looks pretty good to me, good job @GAtom22 :tada: 

Might want to wait for resolution of slash event comment to merge

x/finality/keeper/power_dist_change.go
```diff
@@ -366,25 +274,22 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
...
-		for _, unbodedSats := range fpUnbondedSats {
-			fpDistInfo.RemoveBondedSats(unbodedSats)
+		// update the bonded sats for this finality provider
+		// if had any delta sats during the power distribution change
+		fpDeltaSats := state.DeltaSatsByFpBtcPk[fpBTCPKHex]
+		switch {
+		case fpDeltaSats > 0:
+			fpDistInfo.AddBondedSats(uint64(fpDeltaSats))
+		case fpDeltaSats < 0:
+			satsToRemove := uint64(-fpDeltaSats)
```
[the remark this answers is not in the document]
GAtom22: addressed

x/finality/types/finality.go
GAtom22: done

x/finality/keeper/power_dist_change.go
GAtom22: In the code:
- `[JailFinalityProvider](https://github.com/babylonlabs-io/babylon/blob/91403d33146fc5b1a0b0a0e68c421339fa5fb12f/x/btcstaking/keeper/finality_providers.go#L250-L251)` checks if FP is slashed. So **not** possible to have the sequence SLASHED -> JAILED
- `SlashFinalityProvider` does not check if is the FP is jailed. 
So could happen that a jailed FP is slashed.
- `[UnjailFinalityProvider](https://github.com/babylonlabs-io/babylon/blob/91403d33146fc5b1a0b0a0e68c421339fa5fb12f/x/btcstaking/keeper/finality_providers.go#L285-L286)` does not check if FP is slashed, so could be unjailed
Thus, the sequence SLASHED -> UNJAILED seems to be possible

GAtom22: considering SLASHED is the terminal state for FPs, I'm aligned with @RafilxTenfen 's idea of deferring the slashed events (as we do with the EXPIRED delegations)

RafilxTenfen: > unless we add a check in the UnjailFinalityProvider to avoid unjailing slashed FPs
that also makes sense to me

KonradStaniec: Hmm I am slightly leaning to:
> considering this, and that SLASHED is the terminal state for FPs, I'm aligned with @RafilxTenfen 's idea of deferring the slashed events (as we do with the EXPIRED delegations)
mostly because this better matches semantics of slashing i.e if slashing happens it should not be possible to do anything to change this state. Like the switch that you can flick one time.

So if we were to introduce yet another state for fp in the future, slashing should still be the one overriding this state.

GAtom22: updated with these changes

KonradStaniec (approved):
LGTM ! great work on this one 👍

PR #1406, opened 2025-07-23
feat(btcstkconsumer): Use `collections` for KV store handling.

Vvaradinov:
Closes: https://github.com/babylonlabs-io/babylon/issues/423


x/btcstkconsumer/keeper/consumer_registry.go
```diff
@@ -61,23 +54,14 @@ func (k Keeper) IsCosmosConsumer(ctx context.Context, consumerID string) (bool,
...
-		consumerID := string(consumerIDBytes)
+	var consumerIDs []string
+	err := k.ConsumerRegistry.Walk(ctx, nil, func(consumerID string, _ types.ConsumerRegister) (bool, error) {
 		consumerIDs = append(consumerIDs, consumerID)
+		return false, nil
+	})
+	if err != nil {
+		// Return empty list on error rather than panicking
+		return []string{}
 	}
```
SebastianElvis, line 65: Hmm why not panick here? If `Walk` failed then there might be sth really bad in DB already

SebastianElvis (approved):

PR #1416, opened 2025-07-25
feat: relax requirements for funding transaction

KonradStaniec:
As for now, when expanding stake it is impossible to use transaction which is another staking tx (by hash) as funding tx. This lead to sub par UX.

This pr:
- relaxes this requirement, so that one cannot use another staking output as funding output (not full tx)
- it also fixes validation of staking amount in stake expansion. Babylon cannot assume that staking output in stake expansion wil have the same index as in expanded tx


RafilxTenfen (approved):
LGTM

PR #1417, opened 2025-07-25
fix(zoneconcierge): Add param for BTC header sending limit

Vvaradinov:
Closes: https://github.com/babylonlabs-io/babylon/issues/1373

### DIFF
--- a/x/zoneconcierge/types/params.go
+++ b/x/zoneconcierge/types/params.go
@@ -12,6 +12,9 @@ const (
 
 // NewParams creates a new Params instance
 func NewParams(ibcPacketTimeoutSeconds uint32, maxHeadersPerPacket uint32) Params {
+	if maxHeadersPerPacket == 0 {
+		maxHeadersPerPacket = DefaultMaxHeadersPerPacket
+	}
 	return Params{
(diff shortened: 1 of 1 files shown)


x/zoneconcierge/types/params.go
```diff
@@ -7,18 +7,23 @@ import (
 const (
 	DefaultIbcPacketTimeoutSeconds uint32 = 60 * 60 * 24       // 24 hours
 	MaxIbcPacketTimeoutSeconds     uint32 = 60 * 60 * 24 * 365 // 1 year
+	DefaultMaxHeadersPerPacket     uint32 = 100                // Max BTC headers per IBC packet
```
SebastianElvis, line 10: One note: due to BTC retargeting interval of 2016 blocks, the base header of BTC light client contract will have to be a muliple of 2016. Then upon a handshake, Babylon might need to send over at most 2015 headers to BSN side. If we have a `MaxHeadersPerPacket` smaller than that, then in worst case the handshake might fail. But on the other hand, sending 2015 headers might be too much.

I think BTC light client contract might need some extra treatment on this, e.g., taking a vector of BTC headers upon instanitation rather than just 1 header? @KonradStaniec @liuchengxu

SebastianElvis (approved):

PR #1421, opened 2025-07-25
docs (btcstaking): documentation for reward distribution 

KonradStaniec:
Resolves https://github.com/babylonlabs-io/pm/issues/456

### DIFF
--- a/x/btcstaking/docs/rewards-distribution.md
+++ b/x/btcstaking/docs/rewards-distribution.md
@@ -1,11 +1,6 @@
 # BSN Rewards Distribution on Babylon Genesis
 
 1. [Introduction](#1-introduction)
-    1. [What is BSN Rewards Distribution?](#11-what-is-bsn-rewards-distribution)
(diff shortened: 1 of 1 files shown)


x/btcstaking/docs/rewards-distribution.md
```diff
@@ -0,0 +1,145 @@
...
+> error will be returned to the caller.
+
+> **⚡ Important** The consumer identified by `BsnConsumerId` must exist on
+> Babylon Genesis.
+
+When Babylon Genesis receives a `MsgAddBsnRewards` message, it extracts all
+relevant fields and delegates processing to the `AddBsnRewards` function
+described earlier.
+
+> **⚡ Important** Babylon does not enforce how the `FpRatios` must be
```
RafilxTenfen, line 83: Might be worth to mention that the ratios must sum to 1, even though is already noted in the `FpRatios` attribute in `MsgAddBsnRewards struct`

PR #1428, opened 2025-07-28
fix: update stk exp events

GAtom22:


x/btcstaking/keeper/msg_server.go
```diff
@@ -622,16 +622,16 @@ func (ms msgServer) BTCUndelegate(goCtx context.Context, req *types.MsgBTCUndele
...
-		}
-		types.EmitStakeExpansionActivatedEvent(ctx,
+		// emit an early unbonded event for the delegation (the one that is being expanded)
+		types.EmitEarlyUnbondedEventByStakeExpansion(
+			ctx,
 			btcDel.MustGetStakingTxHash().String(),
 			spendStakeTxHash.String(),
-			req.StakeSpendingTxInclusionProof.Key.Hash.MarshalHex(),
-			req.StakeSpendingTxInclusionProof.Key.Index,
+			stakerSpendigTxHeader.Height,
```
RafilxTenfen, line 630: Should we add this field to `BTCStakingConsumerEvent_UnbondedDel` as well or it is not needed ?

PR #1430, opened 2025-07-28
fix(evm): Bump cosmos evm to 0.41 and add eth signer extractor for the priority nonce mempool

Vvaradinov:
This PR updates the evm to the latest commit hash that included the priority nonce mempool https://github.com/cosmos/evm/pull/245. This adds the mandatory signer extractor to the mempool configuration. 

Additional changes in our testnet command, cosmos evm version and other configurations are needed.

### DIFF
--- a/cmd/babylond/cmd/custom_babylon_config.go
+++ b/cmd/babylond/cmd/custom_babylon_config.go
@@ -68,8 +68,7 @@ type BabylonAppConfig struct {
 func DefaultBabylonAppConfig() *BabylonAppConfig {
 	baseConfig := *serverconfig.DefaultConfig()
 	// Update the default Mempool.MaxTxs to be 0 to make sure the PriorityNonceMempool is used
-	// TODO: Using no-op mempool for EVM until issue with mempool support is resolved
-	baseConfig.Mempool.MaxTxs = -1
+	baseConfig.Mempool.MaxTxs = 0
 	// The SDK's default minimum gas price is set to "0.002ubbn" (empty value) inside
 	// app.toml, in order to avoid spamming attacks due to transactions with 0 gas price.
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -983,10 +989,19 @@ func getAppMempool(appOpts servertypes.AppOptions) mempool.Mempool {
 		maxTxs     = cast.ToInt(appOpts.Get(server.FlagMempoolMaxTxs))
 		mempoolCfg = mempool.DefaultPriorityNonceMempoolConfig()
 	)
+	mempoolCfg.SignerExtractor = NewEthSignerExtractionAdapter(mempool.NewDefaultSignerExtractionAdapter())
```
canu0205, line 992: also needs to set EthSignerExtractionAdapter for proposal handler which is used at PrepareProposalHandler.

in the `NewProposalHandler` of `x/checkpointing/prepare/proposal.go`, set EthSignerExtractionAdapter like this below:

```
defaultHandler := baseapp.NewDefaultProposalHandler(mp, bApp)
defaultHandler.SetSignerExtractionAdapter(ethSignerExtractor)
```

even though this doesn't affect txs handling right now, I think it would be better to keep proper mempool setup.

reference - https://github.com/cosmos/evm/blob/main/evmd/app.go#L785

canu0205 (commented):
left a comment on setup EthSignerExtractionAdapter on proposal handler.

besides that, wiring cosmos evm v0.4.1 looks correct.

canu0205:
@Vvaradinov is it ready for review it again?

Vvaradinov:
@canu0205 yes can you please re-review

KonradStaniec (approved):

canu0205 (approved):
LGTM!

one minor comment regarding changing the maxTxs at testnet command and default babylon config for utilizing priority nonce mempool (it can be modified future)

PR #1437, opened 2025-07-28
fix(zoneconcierge): Send base headers before the fallback cases

Vvaradinov:
Include base header if it's within `k` blocks of tip, otherwise fall back to standard `k+1` headers

### DIFF
--- a/x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
+++ b/x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
@@ -148,6 +148,26 @@ func (k Keeper) getDeepEnoughBTCHeaders(ctx context.Context) []*btclctypes.BTCHe
 	return k.btclcKeeper.GetMainChainFrom(ctx, startHeight)
 }
 
+// getHeadersFromBaseOrFallback tries to include base header when possible, otherwise falls back to k+1 headers
+func (k Keeper) getHeadersFromBaseOrFallback(ctx context.Context, consumerID string) []*btclctypes.BTCHeaderInfo {
+	baseHeader := k.GetBSNBaseBTCHeader(ctx, consumerID)
+	if baseHeader == nil {
+		return k.getDeepEnoughBTCHeaders(ctx)
+	}
+
+	// Check if we can start from base header without sending too many headers
+	tipHeight := k.btclcKeeper.GetTipInfo(ctx).Height
+	kValue := k.btccKeeper.GetParams(ctx).BtcConfirmationDepth
+
+	if tipHeight > baseHeader.Height && tipHeight-baseHeader.Height > kValue {
+		// Base header is too old, fall back to k+1 headers from tip
+		return k.getDeepEnoughBTCHeaders(ctx)
+	}
+
+	// Include base header for BSN initialization
+	return k.btclcKeeper.GetMainChainFrom(ctx, baseHeader.Height)
+}
+
 // getHeadersToBroadcastForConsumer retrieves headers to be broadcasted to a specific BSN
 // The headers to be broadcasted are:
 // - If no BSN base header exists: use the last k+1 headers from tip (fallback)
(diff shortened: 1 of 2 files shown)


x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
```diff
@@ -148,6 +148,34 @@ func (k Keeper) getDeepEnoughBTCHeaders(ctx context.Context) []*btclctypes.BTCHe
...
+	}
+
+	// Check if we can start from base header without sending too many headers
+	tipHeight := k.btclcKeeper.GetTipInfo(ctx).Height
+	kValue := k.btccKeeper.GetParams(ctx).BtcConfirmationDepth
+
+	if tipHeight >= baseHeader.Height && tipHeight-baseHeader.Height > kValue {
+		// Base header is too old, fall back to k+1 headers from tip
+		return k.getDeepEnoughBTCHeaders(ctx)
+	}
```
SebastianElvis, line 165: Not for this PR, but there is a situation we need to deal with: Due to Bitcoin's difficulty adjustment interval of 2016 blocks, BSN's base header height `h` needs to satisfy `h % 2016 = 0`. This means even if the BSN chooses the most recent possible header, `tipHeight >= baseHeader.Height` can be at most 2015. However, `getDeepEnoughBTCHeaders` sends only `k+1` headers. This needs to be fixed on the contract side.

I see various approaches:

- BSN contracts take a list of BTC headers with the first one being the retargeting header
- BSN contracts accept BTC headers that are not retargeting ones as base header, and before retargeting BSN contracts enforce less verifictaion rules
- We get rid of BTC light client completely as we go with lite verification anyway

What do you guys think?

@KonradStaniec @gitferry @liuchengxu @maurolacy

SebastianElvis (approved):

x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
gitferry: In the current impl, the base header is set during IBC channel creation `https://github.com/babylonlabs-io/babylon/blob/e9b6aa84626ae6356b53457fe824c65bf8a2d8ba/x/zoneconcierge/keeper/ibc_packet_btc_staking_consumer_event.go#L141`, which caused the issue that if the first batch was sent after `k` block time, the base header set here is not consistent with the base header that will be set on the BSNs side (accept first header of the first batch as base header if the base header is not specified during instantiation).

What about we do a lazy way where we do not set base header during IBC channel creation, but we send [deep enough headers](https://github.com/babylonlabs-io/babylon/blob/eb23aca8afde10f62f097631466d6d075947b7e4/x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go#L163) if the base header not set and set the first header as the base header for the BSN?

gitferry (commented):
In the current impl, the base header is set during

x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
SebastianElvis: > if the first batch was sent after k block time

I don't think this will happen. Babylon is broadcasting BTC header as soon as there is a new one, upon each [EndBlock](https://github.com/babylonlabs-io/babylon/blob/62ec2306e9a4347f6001f72badd17c8b345bb86d/x/zoneconcierge/abci.go#L26-L30). So in theory the gap between channel handshake and a new BTC header IBC packet will be within 1 BTC block time.

liuchengxu: I have no strong opinions, honestly. There are multiple factors here, and I don't know which ones are more important.

Furthermore, I'm curious—why isn’t sending the 2015 headers from Babylon to the BSN side considered a viable option? This can be another easy solution if acceptable.

Each Bitcoin header is only 80 bytes, so 2015 headers amount to roughly 2015 * 80 / 1024 ≈ 158 KiB. Even with IBC framing overhead, the total packet size should still be well under 200 KiB. That’s far below the typical IBC packet size limit (~4 MiB), and even within the conservative 1 MiB cap many production chains use from ChatGPT.

gitferry: > We get rid of BTC header verification in BSN contracts and let BSN contracts accept any header sent from Babylon

Seems we are going with this approach?

gitferry (approved):

PR #1447, opened 2025-07-29
imp: add allow list for e2e test

GAtom22:
# Descrition

Add allow list with expiration height = 0 to be used on e2e tests.

Currently, e2e tests docker image is using the testnet values. Once we update these
to the desired values, some e2e test will fail due to impossibility of creating multi-staking delegations.
With the changes introduced here, no tests will be broken when updating testnet values

### DIFF
--- a/x/btcstaking/types/allowlist/multi_staking_allow_list_e2e.go
+++ b/x/btcstaking/types/allowlist/multi_staking_allow_list_e2e.go
@@ -3,6 +3,7 @@
 package allowlist
 
 // Values here are not used in production and are only used for e2e tests.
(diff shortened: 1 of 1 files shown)


x/btcstaking/types/allowlist/multi_staking_allow_list_testnet.go
```diff
@@ -1,4 +1,4 @@
-//go:build testnet
+//go:build testnet && !e2e_upgrade
```
RafilxTenfen, line 1: starting to get complex :exploding_head:

RafilxTenfen (approved):
Do we need to add any flags to makefile while running the e2e?

PR #1449, opened 2025-07-30
fix: math overflow incentive reward tracker

RafilxTenfen:


x/incentive/migrations/v2/store.go
```diff
@@ -0,0 +1,35 @@
...
+
+// MigrateStore performs in-place store migrations.
+// Migration updates all finality provider current rewards by the hardcoded decimal rewards
+// which adds 20 decimals points to increase precision at the time to calculate how much
+// rewards each satoshi staked is entitled to receive.
+func MigrateStore(
+	ctx sdk.Context,
+	k Keeper,
+) error {
+	return k.IterateFpCurrentRewards(ctx, func(fp sdk.AccAddress, fpCurrRwds types.FinalityProviderCurrentRewards) error {
```
KonradStaniec, line 26: It would be good to do an experiment with lots of rewards how long migration will take (this can be done after merging pr)

x/incentive/migrations/v2/store.go
```diff
@@ -0,0 +1,35 @@
...
+		currentRewardsWithDecimals, err := bbntypes.CoinsSafeMulInt(fpCurrRwds.CurrentRewards, types.DecimalRewards)
+		if err != nil {
+			return err
```
KonradStaniec, line 29: lets wrap this error with some more info that this is in fact migration that failed so something like:
```go
fmt.Errorf("incentive module migration failed: %w", err)
```
maybe we can even add addreess etc to the message, it will help with debugging if something goes wrong

types/sdk_math.go
```diff
@@ -0,0 +1,48 @@
...
+func CoinsSafeMulInt(coins sdk.Coins, x sdkmath.Int) (sdk.Coins, error) {
+	if x.IsZero() {
+		return nil, fmt.Errorf("%w: cannot multiply coins by zero", ErrInvalidAmount)
+	}
+
+	res := make(sdk.Coins, len(coins))
+	for i, coin := range coins {
+		amt, err := coin.Amount.SafeMul(x)
+		if err != nil {
+			return sdk.Coins{}, err
```
KonradStaniec, line 22: maybe lets wrap those errors nicely

x/incentive/migrations/v2/store.go
RafilxTenfen: it should be a small number, the number of different active fps that got rewards at some point. But I will ask devops to get this data from mainnet so we can make sure it will not take long

PR #1453, opened 2025-07-30
fix: inactive fp event

GAtom22:


GAtom22:
@KonradStaniec do we want this backported to v2?

x/finality/types/power_table.go
```diff
@@ -62,6 +62,22 @@ func (dc *VotingPowerDistCache) FindNewInactiveFinalityProviders(prevDc *VotingP
 		}
 	}
 
+	// Also check for finality providers that were active before but now
```
RafilxTenfen, line 65: what if we throw the event if the `fpDistinfo` is out of power 
```go
		if fpDistInfo.TotalBondedSat > 0 {
			newDc.AddFinalityProviderDistInfo(fpDistInfo)
		} else {
			// emit inactive event
```
and then remove `FindNewInactiveFinalityProviders`, I didn't double checked if that is enough but just an idea

KonradStaniec:
> The quick and easy fix is[ keeping the fps with 0 sats](https://github.com/babylonlabs-io/babylon/blob/954b1c2a8929002830fef3a8839f2e78c83416eb/x/finality/keeper/power_dist_change.go#L235) in the cache

I vaguely remember we should not do this as this has negative impact on block tallying in finality module i.e if finality provider has 0 voting power he should not be added to cache and saved in active FPs store

x/finality/types/power_table.go
```diff
@@ -62,6 +62,22 @@ func (dc *VotingPowerDistCache) FindNewInactiveFinalityProviders(prevDc *VotingP
...
+	// are not present in the current cache due to full unbonding.
+	currFps := make(map[string]struct{})
+	for _, fp := range dc.FinalityProviders {
+		currFps[fp.BtcPk.MarshalHex()] = struct{}{}
+	}
+	prevActivePfs := prevDc.GetActiveFinalityProviderSet()
```
KonradStaniec, line 71: Note here: at this point in event processing we are after the `k.RecordVotingPowerAndCache(ctx, newDc)` function 
which means both `dc` and `prevDc` contains only active finality providers.  

Does it mean whole begining of the function i.e:
```go
inactiveFps := dc.GetInactiveFinalityProviderSet()
prevInactiveFps := prevDc.GetInactiveFinalityProviderSet()
newInactiveFps := make([]*FinalityProviderDistInfo, 0)
for pk, fp := range inactiveFps {
  _, exists := prevInactiveFps[pk]
  if !exists {
	  newInactiveFps = append(newInactiveFps, fp)
  }
}
```

is essentially one big no-op that we can delete ? (`GetInactiveFinalityProviderSet` will always return empty set for both dcs)

x/finality/types/power_table.go
GAtom22: We'll need to add more logic to check if it was active before emitting the INACTIVE event. I believe will complicate and more this part of the code

x/finality/types/power_table.go
KonradStaniec: scratch that, I misunderstood something there 👍

PR #1455, opened 2025-07-31
fix: correct file path reference in finality README

VolodymyrBg:
- Change reference from "./keeper/voting_power_table.go" to "./keeper/power_table.go"
The actual file is named "power_table.go", not "voting_power_table.go"


RafilxTenfen:
main changed

PR #1473, opened 2025-08-02
chore: e2e test for incentives migration in upgrade v3

RafilxTenfen:
- Add test for incentive rewards structure migrations for upgrade v2 -> v3
- Add missing modules to run if build tag `e2e_upgrade` is set
- Update v3 upgrade func parameters order
- Possibly fix to flake e2e rewards test where fp1 that had higher voting power (enough to finalize block by himself) was sending the finality sig first and fp2 was sending right after, but sometimes the block might get finalized without fp2 having included his vote


app/upgrades/v3/upgrades.go
```diff
@@ -22,68 +22,68 @@ const (
...
 				DeletedCapabilityStoreKey,
 			},
 		},
 	}
 }
 
-func CreateUpgradeHandler(fpCount uint32, btcActivationHeight uint32,
-	permissionedIntegration bool, ibcPacketTimeoutSeconds uint32) upgrades.
-	UpgradeHandlerCreator {
+func CreateUpgradeHandler(
```
RafilxTenfen, line 49: just change param order to (bool, ...uint32)

test/e2e/btc_rewards_distribution_e2e_test.go
```diff
@@ -297,19 +297,6 @@ func (s *BtcRewardsDistribution) Test4CommitPublicRandomnessAndSealed() {
...
-			s.finalityBlockHeightVoted,
-			s.fp1RandListInfo.SRList[s.finalityIdx],
-			&s.fp1RandListInfo.PRList[s.finalityIdx],
-			*s.fp1RandListInfo.ProofList[s.finalityIdx].ToProto(),
-			fmt.Sprintf("--from=%s", wFp1),
-		)
-	}()
-
 	go func() {
 		defer wg.Done()
```
RafilxTenfen, line 301: fp2 should vote first as he has lower voting power 

This was a fix for a flaky e2e test where it could happen that the block would be finalized only with fp1 vote (since he had enough vp to finalize the block by himself)

GAtom22 (approved):
LGTM! Thanks @RafilxTenfen!!

PR #1474, opened 2025-08-03
chore: add incentives query fp current rewards

RafilxTenfen:
- Fix export genesis for incentive module (it can exist fp current rewards with empty coins)
- Add new query to get the current fp rewards for a given fp address


GAtom22 (approved):
LGTM! Thanks @RafilxTenfen!! :rocket:

Issue #1475, opened 2025-08-04
btcstaking: reject FP registration requests for Cosmos BSNs without zone concierge IBC channel


SebastianElvis:
We need to add a check at [CreateFinalityProvider](https://github.com/babylonlabs-io/babylon/blob/8e728e8d992a3b81403e0449a6e843261dcdf25c/x/btcstaking/keeper/msg_server.go#L65-L66) handler that, if the given BSN is Cosmos BSN and its IBC channel is still nil , then the function should return error. 

This prevents users from registering FPs for Cosmos BSNs that haven't created an IBC channel, and from creating BTC delegations to this BSNs' FPs

PR #1485, opened 2025-08-04
imp: add bsn chan check on fp creation

GAtom22:
Closes https://github.com/babylonlabs-io/babylon/issues/1475

To be able to check if there's a channel open for the corresponding BSN, there's a cyclic dependency with the zoneconcierge keeper. To avoid this, this PR performs a refactor to decouple a zoneconcierge channel keeper to be used on the BTC consumer staking keeper and the general zoneconcierge keeper


KonradStaniec (approved):
Only minor comment.

LGTM! imo it is cleaner to have all zoneconcierge channel operations separated into separate keepek 👍

PR #1489, opened 2025-08-04
feat: determine consumer ID from IBC packet in memo add bsn rewards

RafilxTenfen:
- Fix importing cycle in upgrade v3 test
- Add `ChannelKeeper`  to `btcstkconsumer` to calculate the consumer ID based on a port and channel ID
- Removed `AccountKeeper` and `BankKeeper` from `btcstkconsumer`, it was not being used
- Update BSN add rewards by IBC callbacks to load the consumer ID if it is not set in the memo field


x/btcstaking/keeper/ibc_callbacks.go
```diff
@@ -142,7 +142,16 @@ func (k Keeper) processAddBsnRewards(
 		return status.Errorf(codes.InvalidArgument, "invalid address %s: %v", transferData.Receiver, err)
 	}
 
-	return k.AddBsnRewards(ctx, receiverOnBbnAddr, callbackAddBsnRewards.BsnConsumerID, sdk.NewCoins(bsnReward), callbackAddBsnRewards.FpRatios)
+	bsnConsID := callbackAddBsnRewards.BsnConsumerID
+	if len(bsnConsID) == 0 {
+		// uses the destination because it is the port and channel registered on babylon side
```
RafilxTenfen, line 147: @SebastianElvis @gitferry please confirm that this should be the Dest part of the packet, not source

PR #1506, opened 2025-08-06
fix!(zoneconcierge): revert base header related logics

### LINKED ISSUE #1452: zoneconcierge: Investigate the removal of global `lastSentSegment`
Zoneconcierge has two separate `lastSentSegment` tracking mechanisms:

   1. Global lastSegment (`GetLastSentSegment`) - used by the fallback method `getHeadersToBroadcast` and only updated when checkpoints finalize via hooks.
   2. Per-BSN lastSegment (`GetBSNLastSentSegment`) - used by the new method `getHeadersToBroadcastForConsumer` and updated every time we broadcast new headers.

It might be confusing to keep both of these so a optimal refactor would leave only the per consumer `BSNLastSentSegment` and adjust the fallback cases. It might also be causing a bug where the module continuously sends `k+1` headers as the global `lastSentSegment` is only being update in the `AfterRawCheckpointFinalized` hook [here](https://github.com/babylonlabs-io/babylon/blob/8445a9eb3c9d81fc228d195cce5c05addae27688/x/zoneconcierge/keeper/hooks.go#L49). which causes infinite fallback case trigger [here](https://github.com/babylonlabs-io/babylon/blob/8445a9eb3c9d81fc228d195cce5c05addae27688/x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go#L237)

dongsam:
closes: #1452

### Context:
BaseHeader is currently updated only when a `BsnBaseBtcHeader` packet is received from BSN after initial IBC handshake setup. However, in the current bsn-contracts, there is only a struct definition for this packet and no logic found that actually sends it. This might be the reason why the fallback logic for BTC header packet keeps getting triggered repeatedly.

### Direction
- Revert base header related logic of #1331, #1437 to fix #1452
- Detailed context, summary on [here](https://github.com/babylonlabs-io/babylon/issues/1452#issuecomment-3158565161)

_(It is still a draft. I will convert it to R4R if reviews or questions are needed)_


dongsam:
Closing this PR because it was cherry-picked from #1512, finalized, and merged

PR #1524, opened 2025-08-07
chore: remove slash consumer event from babylon add fp sig

RafilxTenfen:
- Removed [PropagateFPSlashingToConsumers](https://github.com/babylonlabs-io/babylon/commit/333e5ca6751dc02986788b5ec03e392b86b2d7d2) from add finality sig call 
- Structure `SlashedBTCDelegation` was not removed, due to being referenced in another places, including rust contracts [ref](https://github.com/search?q=org%3Ababylonlabs-io%20BTCStakingIBCPacket&type=code) (optionally PR # was based on this branch and it is removing the structure from babylon node)
- Add replay test with `200` BTC delegations active


RafilxTenfen:
close in favor of #1525

PR #1540, opened 2025-08-11
imp(zoneconcierge): refactor end-blocker to avoid duplicated queries

GAtom22:


x/zoneconcierge/keeper/ibc_packet_btc_headers.go
```diff
@@ -28,17 +30,26 @@ func (k Keeper) BroadcastBTCHeaders(ctx context.Context) error {
 	// TODO: Improve reorg handling efficiency - instead of sending from Consumer base to tip,
 	// we should send a dedicated reorg event and then send headers from the reorged point to tip
 
+	// Build a map for O(1) channel lookups
+	consumerChannelMap, err := k.buildConsumerChannelMap(ctx, consumerChannels)
```
RafilxTenfen, line 34: Can we construct this consumer map only once in the `EndBlocker` and pass as parameter for the funcs `BroadcastBTCHeaders` and `BroadcastBTCStakingConsumerEvents`

x/zoneconcierge/keeper/ibc_packet_btc_timestamp.go
```diff
@@ -238,33 +248,36 @@ func (k Keeper) BroadcastBTCTimestamps(
...
-			)
-			continue
-		}
-
-		packet := types.NewBTCTimestampPacketData(btcTimestamp)
-		if err := k.SendIBCPacket(ctx, channel, packet); err != nil {
-			if errors.Is(err, clienttypes.ErrClientNotActive) {
-				k.Logger(sdkCtx).Info("IBC client is not active, skipping consumer",
+		// Send to all channels for this consumer
+		for _, channel := range channels {
```
RafilxTenfen, line 252: In my mind we are always only going to have one channel per consumer 

Any opinions here? @gusin13 @SebastianElvis @gitferry

KonradStaniec: yep, there is always only one channel per consumer, see : https://github.com/babylonlabs-io/babylon/blob/v3.0.0-snapshot.250807/x/zoneconcierge/keeper/ibc_packet_btc_staking_consumer_event.go#L101
in which we enforce this during channel handshake

KonradStaniec (commented):
In general looks good. If I understand correctly this won't be state breaking change ?

x/zoneconcierge/keeper/channel.go
```diff
@@ -0,0 +1,22 @@
...
+	consumerChannelMap := make(map[string]channeltypes.IdentifiedChannel)
+	for _, channel := range channels {
+		clientID, err := k.channelKeeper.GetClientID(ctx, channel)
+		if err != nil {
+			return nil, err
+		}
+		consumerChannelMap[clientID] = channel
+	}
+	return consumerChannelMap, nil
+}
```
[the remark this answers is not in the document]
GAtom22: we need this here because it is called in the EndBlocker for the zoneconcierge module

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [#1529](https://github.com/babylonlabs-io/babylon/pull/1529) Allow `FinalityProviderHistoricalRewards` to have empty coins
 to export genesis.
 - [#1539](https://github.com/babylonlabs-io/babylon/pull/1539) update unbonding in replay testsuite
+- [#1540](https://github.com/babylonlabs-io/babylon/pull/1540) Optimizations on `x/zoneconcierge` `EndBlocker` logic.
```
KonradStaniec, line 46: Please re-base on lates `main` to have this in correct place.

KonradStaniec (approved):
If this is state breaking we can release this along v3.0.0-rc.3 after we will have some state breaking fix there 👍 

LGTM!

PR #1545, opened 2025-08-13
fix: remove unnecessary validation

KonradStaniec:
Removes the unnecessary check that covenant members signed parent delegaiton


RafilxTenfen (approved):
UtACK

Issue #1542, opened 2025-08-12
zc: query APIs for exposing non-BTC-timestamped information


SebastianElvis:
- [ ] Get last epoch header of a given consumer ID (exposing https://github.com/babylonlabs-io/babylon/blob/ff15aa54445a82de9705beec2f4072bfc2a6db0c/x/zoneconcierge/keeper/epoch_header_indexer.go#L125-L133)
- [ ] Get last sent segment of a given consumer ID (exposing https://github.com/babylonlabs-io/babylon/blob/ff15aa54445a82de9705beec2f4072bfc2a6db0c/x/zoneconcierge/keeper/consumer_btc_state.go#L22-L30)
- [ ] Get sealed epoch proof of a given epoch (exposing https://github.com/babylonlabs-io/babylon/blob/ff15aa54445a82de9705beec2f4072bfc2a6db0c/x/zoneconcierge/keeper/epochs.go#L29-L37)

PR #1549, opened 2025-08-13
chore(zc): query APIs, exposing non-BTC-timestamped information

samricotta:
closes: https://github.com/babylonlabs-io/babylon/issues/1542

Add query APIs for exposing non-BTC-timestamped information:

- `LatestEpochHeader` - Get last epoch header of a given consumer ID
- `BSNLastSentSegment` - Get last sent segment of a given consumer ID  
- `GetSealedEpochProof` - Get sealed epoch proof of a given epoch -prefixed with `Get` due to naming conflict with existing keeper field)
- tests for all

### DIFF
--- a/test/e2e/configurer/chain/queries_zoneconcierge.go
+++ b/test/e2e/configurer/chain/queries_zoneconcierge.go
@@ -1,6 +1,7 @@
 package chain
 
 import (
+	"fmt"
 	"net/url"
 
 	"github.com/stretchr/testify/require"
@@ -9,6 +10,7 @@ import (
 	zoneconciergetype "github.com/babylonlabs-io/babylon/v4/x/zoneconcierge/types"
 )
 
+// QueryZoneConciergeParams retrieves the current parameters for the ZoneConcierge module
 func (n *NodeConfig) QueryZoneConciergeParams() *zoneconciergetype.Params {
 	bz, err := n.QueryGRPCGateway("/babylon/zoneconcierge/v1/params", url.Values{})
 	require.NoError(n.t, err)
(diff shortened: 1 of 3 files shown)


proto/babylon/zoneconcierge/v1/query.proto
```diff
@@ -63,3 +81,35 @@ message FinalizedBSNData {
...
+// QueryLatestEpochHeadersRequest RPC method.
+message QueryLatestEpochHeaderRequest {
+  string consumer_id = 1;
+}
+
+// QueryLatestEpochHeaderResponse is response type for the
+// QueryLatestEpochHeaderResponse RPC method.
+message QueryLatestEpochHeaderResponse {
+  IndexedHeader header = 1;
+}
```
SebastianElvis, line 95: Since we are distinguishing DB objects and query API objects, we will need to avoid using `IndexedHeader`, `BTCChainSegment`, and `ProofEpochSealed`, and instead create new response structs e.g., `IndexedHeaderResponse`. At this stage they can be exactly same as the DB objects, and they might diverge in the futrue

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -107,3 +107,70 @@ func (k Keeper) FinalizedBSNsInfo(c context.Context, req *types.QueryFinalizedBS
 
 	return resp, nil
 }
+
+func (k Keeper) LatestEpochHeader(goCtx context.Context,
+	req *types.QueryLatestEpochHeaderRequest) (*types.
+QueryLatestEpochHeaderResponse, error) {
```
SebastianElvis, line 113: Indent seems weird

x/zoneconcierge/keeper/grpc_query.go
SebastianElvis, line 1: can we also have CLIs for those queries?

SebastianElvis (approved):

SebastianElvis (approved):
Great work!

Issue #1395, opened 2025-07-20
chore: e2e test life improvement


RafilxTenfen:
Our dev experience in e2e test is broken, where it increases a lot the time to develop and add new checks/tests

I believe that the main culprit is the use of `TestSuite` and having test orders `Test1... Test2...` to be executed
If one of the checks ex.: `require.Len(s.T(), actualFps, 4)` fails in `Test3...`  it jumps to execute the next test `Test4...`  and this slows down a lot the bug finding

Other issue that slows the e2e test execution down is acc sequence missmatch errors, we should include sequence on tx execution and keep track in the signer cases, otherwise we can always only send one tx per wallet per block

KonradStaniec:
One thing we should consider while doing any improvements to e2e test is what those tests should be checking ? 

Given that those tests are pretty heavy and not easy to setup, we should probably avoid checking for any edge cases or such things but rather focus on whole user stories touching whole node. 

So probably there are few such big stories that could be separate whole tests:
- create btc delegations through every possible way (pre/post registration and stake expansion), vote on few blocks, retrieve rewards by stakers and fps
- timestamping whole flow (finish epoch / create checkpoint / send it to btc / report to Babylon / finalize)
- upgrade handler for the next upcoming upgrade with assertions required to verify migrations were performed and all new modules were wired properly. (for example, I do not think we need to forever have e2e test `Test3PacketForwarding` , it should be part of this upgrade handler test)
- e2e tests for app specific gov proposals (resume finality handler)

PR #1552, opened 2025-08-13
chore:  e2e refactory

RafilxTenfen:
- This PR creates new structures to handle e2e testing, it still uses docker to setup and run the chains + hermes 
- Adds option to wallet senders to wait for the tx to be included in the block and check the result of that transaction is `code == 0`
- Creates concept of acc seq and number to wallets sender (similar to replay tests)
- Sends the transactions via `sdk.Msg` types with signed transaction with GRPC instead of using docker CLI to execute transaction commands
- Allocates ports dynamically to allow multiple `e2ev2` tests to run in parallel 

## What is missing? 
- Migrate all the e2e tests to the new pattern
- Create structures of FinalityProvider and BtcStakers that contain an `WalletSender`


RafilxTenfen:
partially solves https://github.com/babylonlabs-io/babylon/issues/1395

test/e2ev2/govProps/resume-finality.json
```diff
@@ -0,0 +1,17 @@
+{
```
GAtom22, line 1: can we define a utility function to create the proposal json file as a temporary file?
So we don't have to keep these and we can make it flexible for any type of proposal

test/e2ev2/tmanager/chain.go
```diff
@@ -0,0 +1,234 @@
...
+
+// NewChainConfig creates a new chain configuration with default values
+func NewChainConfig(tempDir, chainID string) *ChainConfig {
+	return &ChainConfig{
+		ChainID: chainID,
+		Home:    filepath.Join(tempDir, chainID),
+		// starts with 2 nodes (1 val, one non-validator)
+		ValidatorCount:        1,
+		NodeCount:             1,
+		BlockTime:             5 * time.Second,
```
GAtom22, line 64: can we use lower block time (eg 1 or 2 sec) so tests don't take too long?

test/e2ev2/tmanager/chain.go
```diff
@@ -0,0 +1,234 @@
...
+		EpochLength:           10,
+		VotingPeriod:          30 * time.Second,
+		ExpeditedVotingPeriod: 15 * time.Second,
```
GAtom22, line 67: ditto here. If we can reduce block time, we can reduce these a bit too

GAtom22 (commented):
Great work @RafilxTenfen!! :raised_hands: 
Left a couple of comments

test/e2ev2/govProps/resume-finality.json
RafilxTenfen: actually, now gov props should be send as sdk.Msg, instead of using CLI to send, so the files will not be needed anymore

test/e2ev2/tmanager/chain.go
RafilxTenfen: move to 2s

GAtom22 (approved):

PR #1553, opened 2025-08-14
fix(evm): Remove default wrapped gas token pair

Vvaradinov:
Remove the default addition of an ERC20 precompile for wrapped BABY representation since we are going with a solid WETH9 contract for WBABY

### DIFF
--- a/cmd/babylond/cmd/genesis.go
+++ b/cmd/babylond/cmd/genesis.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"cosmossdk.io/math"
-	"github.com/babylonlabs-io/babylon/v4/app"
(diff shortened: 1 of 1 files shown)


canu0205:
LGTM!

one question - do we have any plans to deploy and manage WETH9 contract for BABY on the devnet?

canu0205 (approved):
LGTM!

one question - do we have any plans to deploy and manage WETH9 contract for BABY on the devnet?

Vvaradinov:
@canu0205 yes ideally we would have a script to deploy it and a script to run some tests on it that we will give to the devops team to run. Afterwards that contract address will be shared with our partners so they can use it in their contracts.

PR #1555, opened 2025-08-14
chore: Create mergify rule for release v4

Vvaradinov:


.mergify.yml
```diff
@@ -11,6 +11,23 @@ queue_rules:
       {{ body }}
     merge_method: squash
 pull_request_rules:
+  - name: auto-add v4.x backport label when v3.x label is added
+    conditions:
+      - base=main
+      - label=backport-to-release/v3.x
+      - -label=backport-to-release/v4.x
```
GAtom22, line 18: actually, this condition should be removed, no?
Cause we want to add the `backport-to-release/v4.x` label when there's a PR with these conditions:

```
 - base=main
 - label=backport-to-release/v3.x
 ```

Comment (RafilxTenfen) [inline .mergify.yml:18]:
```suggestion
      - label=backport-to-release/v4.x
```

PR #1556, opened 2025-08-14
chore: CLI for submitting equivocation evidence in x/finality module

samricotta:
closes:  #1541

### DIFF
--- a/x/finality/client/cli/tx.go
+++ b/x/finality/client/cli/tx.go
@@ -266,7 +266,10 @@ func AddEvidenceOfEquivocationCmd() *cobra.Command {
 	}
 	flags.AddTxFlagsToCmd(cmd)
 	cmd.Flags().String("signer", "", "signer address for the evidence (required)")
-	cmd.MarkFlagRequired("signer")
+	err := cmd.MarkFlagRequired("signer")
+	if err != nil {
+		return nil
+	}
 
 	return cmd
 }


x/finality/client/cli/tx.go
```diff
@@ -200,3 +201,59 @@ func NewUnjailFinalityProviderCmd() *cobra.Command {
...
+				Signer:                  clientCtx.FromAddress.String(),
+				FpBtcPkHex:              args[0],
+				BlockHeight:             blockHeight,
+				PubRandHex:              args[2],
+				CanonicalAppHashHex:     args[3],
+				ForkAppHashHex:          args[4],
+				CanonicalFinalitySigHex: args[5],
+				ForkFinalitySigHex:      args[6],
+				SigningContext:          args[7],
+			}
```
KonradStaniec, line 251: Below we should try to parse evidence to check all fields are valid:

Sth like:
```go
if _, err := msg.ParseToEvidence(); err != nil {
	return fmt.Errorf("failed to parse evidence: %v", err)
}
```

SebastianElvis: agreed

x/finality/client/cli/tx.go
```diff
@@ -200,3 +201,66 @@ func NewUnjailFinalityProviderCmd() *cobra.Command {
...
+
+			genOnly, _ := cmd.Flags().GetBool(flags.FlagGenerateOnly)
```
KonradStaniec, line 253: Why this check ? Imo it would be great to check validity of the evidence regardles whether we are broadcasting or generating

PR #1581, opened 2025-08-22
chore:  finality hooks for unbonded/active btc delegation

RafilxTenfen:
- Add hooks to finality to dispatch when an btc delegation is activated and when it is unbonded
- Modified incentives to be the first subscriber of the finality hooks, later coostaking will also subscribe


GAtom22 (approved):
LGTM! Great work @RafilxTenfen!! :tada: 
Please address the linter issues

PR #1589, opened 2025-08-25
imp(btcstkconsumer): add min bbn commission validation

GAtom22:
Closes BP3-CORE-001


RafilxTenfen (approved):
I feel like this could be in the parameters, but if we want to move with hardcoded, this PR looks good to me :tada:

Issue #1615, opened 2025-08-28
`x/incentives`: new parameter to give out % of `fee_collector` to FPs


RafilxTenfen:
- Create a new parameter to collect x% of the `fee_collector` balance each block and send to incentives module acc
- Distribute those amounts to FPs that correctly voted (can be done at the same point where we already [iterate](https://github.com/babylonlabs-io/babylon/blob/8cb9862b1d0631edd15df1039085afb4274823a9/x/incentive/keeper/btc_staking_gauge.go#L48) over the FPs to calculate commissions and accumulate in the reward gauge)

## Amounts to test
- Goal is to reduce inflation from 8% -> 5%
- Give out 0.07% of the 5% to FPs who correctly voted

PR #1631, opened 2025-08-28
imp(incentive): add FP direct rewards param

GAtom22:
Closes https://github.com/babylonlabs-io/babylon/issues/1615


x/incentive/migrations/v3/store.go
```diff
@@ -0,0 +1,27 @@
...
+
+// Keeper the expected keeper interface to perform the migration
+type Keeper interface {
+	SetParams(ctx context.Context, p types.Params) error
+	GetParams(ctx context.Context) (p types.Params)
+}
+
+// MigrateStore performs in-place store migrations.
+// Migration adds the default value for the new FpPortion param
+func MigrateStore(
```
GAtom22, line 19: added this migration, but in case we want to remove this and go with a gov prop after the change, [I added a nil check in the intercept_fee_collector](https://github.com/babylonlabs-io/babylon/pull/1631/files#diff-b6a48071fc2c94c8c20002a164ccf8ef1055ff5a07d5ab1d33ffd8dd58bbaeebR31)

x/incentive/keeper/btc_staking_gauge.go
```diff
@@ -61,28 +68,47 @@ func (k Keeper) RewardBTCStaking(ctx context.Context, height uint64, dc *ftypes.
...
 		// reward the rest of coins to each BTC delegation proportional to its voting power portion
 		coinsForBTCDels := coinsForFpsAndDels.Sub(coinsForCommission...)
 		if err := k.AddFinalityProviderRewardsForBtcDelegations(ctx, fp.GetAddress(), coinsForBTCDels); err != nil {
 			panic(fmt.Errorf("failed to add fp rewards for btc delegation %s at height %d: %w", fp.GetAddress().String(), height, err))
 		}
 	}
-	// TODO: prune unnecessary state (delete BTCStakingGauge after the amount is used)
+
+	// Prune gauges after rewards are distributed to free up storage
+	// k.pruneBTCStakingGauge(ctx, height) // this is used in e2e tests and queries, need to decide what to do
```
GAtom22, line 86: In the software upgrade test, we're querying the BTCStakingGauges to calculate the expected rewards.
Also, there's the module query `BTCStakingGauge`. A few questions here:
- do we want to still prune these and update the e2e test?
- should we add also a query for the  `FPDirectGauge`?
If we're pruning, the queries will work only for those heights where there haven't been any rewards distribution.
Is there a specific reason why we may want to keep the module query?

GAtom22: Other approach, if we want to have these numbers, is emitting an event. So the data will be available in nodes that did not prune the block

x/incentive/keeper/btc_staking_gauge.go
```diff
@@ -113,3 +139,45 @@ func (k Keeper) btcStakingGaugeStore(ctx context.Context) prefix.Store {
 	storeAdapter := runtime.KVStoreAdapter(k.storeService.OpenKVStore(ctx))
 	return prefix.NewStore(storeAdapter, types.BTCStakingGaugeKey)
 }
+
+func (k Keeper) SetFPDirectGauge(ctx context.Context, height uint64, gauge *types.Gauge) {
```
RafilxTenfen, line 143: Suggestion to use the collections map in keeper `collections.Map[[]byte, types.Gauge]` but no strong opinion, if you want to go with store just like the other gauge, all good to me as well

GAtom22: yes, I was following same style as the other gauge

x/incentive/keeper/btc_staking_gauge.go
GAtom22: sg, will add the events, so we can have these values for all blocks (the ones that pruned the gauge and the ones that don't)

PR #1643, opened 2025-09-01
fix: correct BIP-322 witness encoding comment to match implementation

sashass1315:
The comment incorrectly described using "1st byte" for lengths, but the code
actually uses Bitcoin consensus varint encoding via wire.ReadVarInt().


RafilxTenfen:
Main changed

Issue #1620, opened 2025-08-28
`x/coostaking`: upgrade handler to fill up initial data of baby and btc delegators


RafilxTenfen:
- New upgrade handler
- Iterate over baby and btc stakers
- Fill the information of `CoostakerRewardsTracker` for each coostaker
- Test

PR #1660, opened 2025-09-02
chore(coostaking): add upgrade handler

GAtom22:
Closes https://github.com/babylonlabs-io/babylon/issues/1620


app/upgrades/v4rc0/upgrades.go
```diff
@@ -0,0 +1,209 @@
...
+
+	"github.com/babylonlabs-io/babylon/v4/app/keepers"
+	"github.com/babylonlabs-io/babylon/v4/app/upgrades"
+	btcstkkeeper "github.com/babylonlabs-io/babylon/v4/x/btcstaking/keeper"
+	btcstktypes "github.com/babylonlabs-io/babylon/v4/x/btcstaking/types"
+	costktypes "github.com/babylonlabs-io/babylon/v4/x/coostaking/types"
+)
+
+// UpgradeName defines the on-chain upgrade name for the Babylon v4rc0 upgrade
+const UpgradeName = "v4rc0"
```
RafilxTenfen, line 28: costaking will be in v3 I believe :eyes: 

@KonradStaniec

app/upgrades/v4rc0/upgrades.go
```diff
@@ -0,0 +1,209 @@
...
+// InitializeCoStakerRwdsTracker initializes the coostaker rewards tracker
+// It looks for all BTC stakers that are also baby stakers
+func InitializeCoStakerRwdsTracker(
+	ctx context.Context,
+	cdc codec.BinaryCodec,
+	costkStoreService corestoretypes.KVStoreService,
+	stkKeeper *stkkeeper.Keeper,
+	btcStkKeeper btcstkkeeper.Keeper,
+) error {
+	btcStakers, err := getAllBTCStakers(ctx, btcStkKeeper)
```
RafilxTenfen, line 71: We could iterate over the data in incentives as well, might be faster than over all btc delegations since incentives already groups the btc stakers by babylon addr

GAtom22: But the incentives keeps track of the rewards only, no? There may be the case where there's a user that has pending rewards, but is not actively staking bitcoin.
The getAllBTCStakers returns all active BTC delegators

app/upgrades/v4rc0/upgrades.go
```diff
@@ -0,0 +1,209 @@
...
+		sb,
+		costktypes.CoostakerRewardsTrackerKeyPrefix,
+		"coostaker_rewards_tracker",
+		collections.BytesKey,
+		codec.CollValue[costktypes.CoostakerRewardsTracker](cdc),
+	)
+	// we're writing independent key-value
+	// pairs to storage, the order shouldn't affect the final state
+	for addr, val := range coStakers {
+		sdkAddr := sdk.MustAccAddressFromBech32(addr)
```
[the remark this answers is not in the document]
GAtom22: updated

PR #1664, opened 2025-09-03
costaking: hooks baby staking

RafilxTenfen:


x/costaking/keeper/hooks_staking.go
```diff
@@ -0,0 +1,98 @@
...
+
+// AfterDelegationModified implements types.StakingHooks.
+func (h HookStaking) AfterDelegationModified(ctx context.Context, delAddr sdk.AccAddress, valAddr sdk.ValAddress) error {
+	del, err := h.k.stkK.GetDelegation(ctx, delAddr, valAddr)
+	if err != nil { // we stop if the delegation is not found, because it must be found
+		return err
+	}
+
+	beforeAmount := h.k.stkCache.GetAndDeleteStakedAmount(delAddr, valAddr)
+	afterAmount := del.Shares
```
GAtom22, line 29: This does not represent the token amount, but the validator shares that is calculated using [this function](https://github.com/cosmos/cosmos-sdk/blob/e265bb9b42308dc70743b6200a70db9aafb70527/x/staking/types/validator.go#L322):
```
// SharesFromTokens returns the shares of a delegation given a bond amount. It
// returns an error if the validator has no tokens.
func (v Validator) SharesFromTokens(amt math.Int) (math.LegacyDec, error) {
	if v.Tokens.IsZero() {
		return math.LegacyZeroDec(), ErrInsufficientShares
	}

	return v.GetDelegatorShares().MulInt(amt).QuoInt(v.GetTokens()), nil
}
```
Should we backcalculate the corresponding baby amount here?
There's a [`validator.TokensFromSharesTruncated` function](https://github.com/cosmos/cosmos-sdk/blob/e265bb9b42308dc70743b6200a70db9aafb70527/x/staking/types/validator.go#L311-L314) that we could use

x/costaking/keeper/hooks_staking.go
```diff
@@ -0,0 +1,98 @@
...
+// BeforeDelegationSharesModified implements types.StakingHooks.
+func (h HookStaking) BeforeDelegationSharesModified(ctx context.Context, delAddr sdk.AccAddress, valAddr sdk.ValAddress) error {
+	del, err := h.k.stkK.GetDelegation(ctx, delAddr, valAddr)
+	if err != nil {
+		// probably is not found, but we don't want to stop execution for this
+		h.k.Logger(ctx).Error("hook costaking BeforeDelegationSharesModified: %v", err)
+		return nil
+	}
+
+	h.k.stkCache.SetStakedAmount(delAddr, valAddr, del.Shares)
```
GAtom22, line 46: ditto here, the del.Shares is not the staked coins amount

x/costaking/types/staking_cache.go
```diff
@@ -0,0 +1,57 @@
...
+
+	if sc.amtByValByDel[delAddrStr] == nil {
+		sc.amtByValByDel[delAddrStr] = make(map[string]math.LegacyDec)
+	}
+	sc.amtByValByDel[delAddrStr][valAddrStr] = amtStaked
+}
+
+// GetAndDeleteStakedAmount gets the value in the cache if it is found and deletes it after.
+// Note: If a value is not found it returns zero dec.
+func (sc *StakingCache) GetAndDeleteStakedAmount(delAddr sdk.AccAddress, valAddr sdk.ValAddress) math.LegacyDec {
```
GAtom22, line 35: I would separate the logic to delete the cache entries.
To avoid some weird side effects, I'd take the same approach as with the headerCache, and clean it up on every endBlock

GAtom22 (commented):
Great work @RafilxTenfen!! :tada: 
Left a few comments

x/costaking/types/staking_cache.go
RafilxTenfen: addressed to only clear at the end blocker

x/costaking/keeper/hooks_staking.go
```diff
@@ -0,0 +1,103 @@
...
```
GAtom22, line 51: The sequence in the[ Delegate function](https://github.com/cosmos/cosmos-sdk/blob/e265bb9b42308dc70743b6200a70db9aafb70527/x/staking/keeper/delegation.go):
  1. `BeforeDelegationCreated/BeforeDelegationSharesModified `hook  - new hook
  caches shares here
  2. Token transfers and pool management
  3. `AddValidatorTokensAndShares` called - This updates validator's total tokens
  and shares
  4. Delegation shares updated - `delegation.Shares = delegation.Shares.Add(newShares)`
  5. `AfterDelegationModified` hook  - New hook calculates delta here

The issue with this approach is that  when calling `valI.TokensFromShares(delSharesChange).TruncateInt()` in the `AfterDelegationModified` hook, we're using the [validator state after `AddValidatorTokensAndShares`](https://github.com/cosmos/cosmos-sdk/blob/e265bb9b42308dc70743b6200a70db9aafb70527/x/staking/keeper/delegation.go) has already updated the validator's tokens and total shares. This means the exchange rate (tokens/shares ratio) has changed between the before and after calculations.

One solution would be to cache the token amount directly rather than shares in this hook

PR #1670, opened 2025-09-04
chore: rename to costaking

GAtom22:


RafilxTenfen:
shit it breaks https://github.com/babylonlabs-io/babylon/pull/1664 
kkk but since you already did it, lets merge it sooner than latter

PR #1683, opened 2025-09-06
chore(costaking): finality hook

RafilxTenfen:
- Add finality hook to `x/costaking`
- Calculate previous status of finality provider based on the prev voting power distribution cache

Notes to consider
1. If the fp is active (in the active set) the btc delegations for that fp should count for the score
1. If the fp is inactive/jailed/slashed the btc delegations for that fp should NOT count for the score calculation
    - If the fp was already inactive and becomes jailed, it should NOT reduce the score again (same for slashed)
1. If the btc delegation becomes active but the fp is not active, the btc delegation should NOT count for score
1. If the btc delegation unbonds but the fp was already inactive, the btc delegation should NOT reduce the score
1. It doesn't matter if the FP voted for the block or not

Missing:

- Testing


x/finality/keeper/power_dist_change.go
```diff
@@ -95,42 +96,57 @@ func (k Keeper) HandleFPStateUpdates(ctx context.Context, prevDc, newDc *ftypes.
 
 	newlyActiveFPs := newDc.FindNewActiveFinalityProviders(prevDc)
 	for _, fp := range newlyActiveFPs {
+		// what was the status prior to update the status?
+		prevStatus := fp.Status()
```
RafilxTenfen, line 100: @GAtom22 what do you think about storing the previous state in FinalityProviderDistInfo and every time we emit an event also update there :eyes:

x/finality/types/finality.go
```diff
@@ -191,3 +191,19 @@ func (e *Evidence) ExtractBTCSK() (*btcec.PrivateKey, error) {
 		e.forkMsgToSign(e.SigningContext), e.ForkFinalitySig.ToModNScalar(), // msg and sig for fork block
 	)
 }
+
+func (fp *FinalityProviderDistInfo) Status() btcstktypes.FinalityProviderStatus {
```
RafilxTenfen, line 195: I am not very confident in defining the previous state like this, it might be usefull to use in the migration if we decide to store the previous state in the

x/finality/keeper/gov.go
```diff
@@ -138,7 +139,8 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
 		k.SetVotingPowerDistCache(ctx, blkHeight, newDc)
 
 		// ensure every active finality provider has signing info
-		k.HandleFPStateUpdates(ctx, dc, newDc, nil)
+		// TODO: check if we should emit events at every height
```
RafilxTenfen, line 142: @GAtom22 it is emitting the events from old heights every block until the current height, we need a way to avoid calling the hooks on this

PR #1690, opened 2025-09-08
chore (btcstkconsumer): add ConsumerActive as a query

samricotta:
Updates e2e tests to e2ev2 for btcstkconsumer for https://github.com/babylonlabs-io/babylon/pull/1549

Also integrated some of the tests from e2e to e2ev2 used for testing rollups

### DIFF
--- a/x/btcstkconsumer/keeper/grpc_query.go
+++ b/x/btcstkconsumer/keeper/grpc_query.go
@@ -167,7 +167,7 @@ func (k Keeper) ConsumerActive(goCtx context.Context,
 		if queryRes == nil {
 			active = false
 		} else {
-			active = queryRes.Label != ""
+			active = queryRes.Label == ""
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


x/btcstkconsumer/keeper/grpc_query.go
```diff
@@ -131,3 +132,53 @@ func (k Keeper) ConsumersRegistry(c context.Context, req *types.QueryConsumersRe
...
+		active = k.channelKeeper.ConsumerHasIBCChannelOpen(ctx, req.ConsumerId, channelID)
+	case types.ConsumerType_ROLLUP:
+		address := consumer.GetRollupConsumerMetadata().FinalityContractAddress
+		contractAddress, err := sdk.AccAddressFromBech32(address)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, "invalid contract address")
+		}
+
+		queryMsg := []byte(`{"config":{}}`)
+		queryRes, err := k.wasmKeeper.QuerySmart(goCtx, contractAddress,
```
KonradStaniec, line 167: I think we do not need to use `QuerySmart` smart here and it is enough if we use:
```go
GetContractInfo(ctx context.Context, contractAddress sdk.AccAddress) *wasmtypes.ContractInfo
```
as `ContractInfo` struct contrains:
```go
// Label is optional metadata to be stored with a contract instance.
Label string `protobuf:"bytes,4,opt,name=label,proto3" json:"label,omitempty"`
```
label field.

What we discussed with @SebastianElvis is that contract should be considered inactive if the `Label != ""` i.e if label is different than empty string.

KonradStaniec (approved):

PR #1713, opened 2025-09-15
chore(costaking): address pr comments and update finality hooks

RafilxTenfen:
Move finality hooks to:
```go
// FinalityHooks event hooks for finality btcdelegation actions
type FinalityHooks interface {
	AfterBtcDelegationUnbonded(ctx context.Context, fpAddr, btcDelAddr sdk.AccAddress, fpSecuresBabylon bool, isFpInActiveSet bool, sats uint64) error
	AfterBtcDelegationActivated(ctx context.Context, fpAddr, btcDelAddr sdk.AccAddress, fpSecuresBabylon bool, isFpInActiveSet bool, sats uint64) error
	AfterBbnFpEntersActiveSet(ctx context.Context, fpAddr sdk.AccAddress) error
	AfterBbnFpExistsActiveSet(ctx context.Context, fpAddr sdk.AccAddress) error
}
```

### DIFF
--- a/test/replay/costaking.go
+++ b/test/replay/costaking.go
@@ -24,10 +24,10 @@ func (d *BabylonAppDriver) CheckCostakerRewards(
 
 	del, err := costkK.GetCostakerRewards(d.Ctx(), addr)
 	require.NoError(d.t, err)
-	require.Equal(d.t, del.ActiveBaby.String(), expActiveBaby.String())
-	require.Equal(d.t, del.ActiveSatoshis.String(), expActiveSats.String())
-	require.Equal(d.t, del.StartPeriodCumulativeReward, expStartPeriod)
-	require.Equal(d.t, del.TotalScore.String(), expTotalScore.String())
+	require.Equal(d.t, expActiveBaby.String(), del.ActiveBaby.String(), "active baby")
+	require.Equal(d.t, expActiveSats.String(), del.ActiveSatoshis.String(), "active sats")
+	require.Equal(d.t, expStartPeriod, del.StartPeriodCumulativeReward, "start period cumulative rewards")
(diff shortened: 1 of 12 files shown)


x/finality/keeper/gov.go
```diff
@@ -151,14 +152,8 @@ func (k Keeper) HandleResumeFinalityProposal(ctx sdk.Context, fpPksHex []string,
...
-		// TODO: check if we should emit events at every height
-		// call HandleFPStateUpdates only at the current height
-		// to account for the last changes (jailed FPs on the proposal)
-		if isCurrHeight {
-			k.HandleFPStateUpdates(ctx, dc, newDc, state)
-			continue
-		}
-		k.HandleFPStateUpdates(ctx, dc, newDc, nil)
+		// Calls empty state until reaches the current block height
+		k.HandleFPStateUpdates(ctx, state, dc, newDc)
```
GAtom22, line 156: do we need to emit events at every height or should we only call the full HandleFPStateUpdates logic for the current height?
If we should only call this for the current height, we can remove the HandleFPStateUpdates call outside the loop (after it)
And with the current changes I believe we should do that, because otherwise we may be activating same FP several times and calling the hook every time in the `processActiveFp` func

GAtom22 (commented):
Great work @RafilxTenfen!! :rocket: 
Looks much cleaner!

Left one comment

x/finality/keeper/gov.go
RafilxTenfen: modified at https://github.com/babylonlabs-io/babylon/pull/1713/commits/0254c176718cea1bb16ecfe7d937546b0d9c9767

PR #1714, opened 2025-09-15
fix(cli): add bls flags to rollback and bootstrap-state cmds

GAtom22:


cmd/babylond/cmd/root.go
```diff
@@ -403,3 +403,57 @@ func appExport(
...
 	return babylonApp.ExportAppStateAndValidators(forZeroHeight, jailAllowedAddrs, modulesToExport)
 }
+
+// newCustomRollbackCmd creates a rollback command with custom BLS flags
+func newCustomRollbackCmd(appCreator servertypes.AppCreator, defaultNodeHome string) *cobra.Command {
+	cmd := server.NewRollbackCmd(appCreator, defaultNodeHome)
+
+	// Add custom BLS flags
+	cmd.Flags().Bool(flagNoBlsPassword, false, "Generate BLS key without password protection (suitable for RPC nodes)")
+	cmd.Flags().String(flagBlsPasswordFile, "", "Load a custom file path to the bls password (not recommended)")
```
RafilxTenfen, line 413: What if we only allow the `flagNoBlsPassword` flag and not `flagBlsPasswordFile` since it is not recommended any way

GAtom22: added it there to support all possible configurations. If we don't support this one, someone that started the node with that flag will not be able to execute the rollback or bootstrap-state cmds

PR #1727, opened 2025-09-18
chore: Bump CI workflows and pin to SHA hashes

filippos47:
This PR bumps all the reusable git workflows to the latest available version, and pins them all to commit hashes instead of git tags for additional security.


toroviej0x0 (approved):
lint ci is broken, but otherwise these changes make sense and LGTM

PR #1731, opened 2025-09-18
fix(costaking): fp inactive and unbond same block

RafilxTenfen:
FP is previously active, so when delegation is unbonded, we remove its sats from the co-staker reward tracker
FP becomes inactive in the same block -> we remove again the same sats

### DIFF
--- a/x/finality/types/power_table.go
+++ b/x/finality/types/power_table.go
@@ -37,11 +37,8 @@ func (dc *VotingPowerDistCache) AddFinalityProviderDistInfo(v *FinalityProviderD
 
 func (dc *VotingPowerDistCache) ActiveFpsByBtcPk() map[string]struct{} {
 	activeFpByBtcPk := make(map[string]struct{}, 0)
-	for idx, fp := range dc.FinalityProviders {
-		canBeActive := idx < int(dc.NumActiveFps)
-		if fp.FpStatus(canBeActive) == bstypes.FinalityProviderStatus_FINALITY_PROVIDER_STATUS_ACTIVE {
-			activeFpByBtcPk[fp.BtcPk.MarshalHex()] = struct{}{}
-		}
+	for _, fp := range dc.FinalityProviders[:dc.NumActiveFps] {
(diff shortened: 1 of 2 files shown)


x/finality/keeper/power_dist_change.go
```diff
@@ -347,6 +340,13 @@ func (k Keeper) ProcessAllPowerDistUpdateEvents(
 		}
 	}
 
+	// Execute the hooks logic based on the processed events
+	// First execute the hooks for BTC delegations
+	// and then for finality providers
+	if err := k.processBTCDelegationHooks(sdkCtx, state, newDc); err != nil {
+		panic(fmt.Errorf("failed to execute btc delegation hooks: %w", err))
```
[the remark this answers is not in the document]
RafilxTenfen: Good point, I will move to after RecordVotingPowerAndCache

x/finality/types/power_table.go
```diff
@@ -35,6 +35,17 @@ func (dc *VotingPowerDistCache) AddFinalityProviderDistInfo(v *FinalityProviderD
 	dc.FinalityProviders = append(dc.FinalityProviders, v)
 }
 
+func (dc *VotingPowerDistCache) ActiveFpsByBtcPk() map[string]struct{} {
+	activeFpByBtcPk := make(map[string]struct{}, 0)
+	for idx, fp := range dc.FinalityProviders {
+		canBeActive := idx < int(dc.NumActiveFps)
+		if fp.FpStatus(canBeActive) == bstypes.FinalityProviderStatus_FINALITY_PROVIDER_STATUS_ACTIVE {
```
KonradStaniec, line 42: I do not understand why do we need this 🤔 

In theory if we you want to get only active finality providers from `VotingPowerDistCache` it is enough to do `dc.FinalityProviders[:NumActiveFps]` i.e properly constructed distibution cache will have only active one in the frist `NumActiveFps` spots

GAtom22: > i.e properly constructed distibution cache will have only active one in the frist NumActiveFps spots
this is partially true, because you could have many jailed/slashed FPs that fit in this range no?

Issue #1662, opened 2025-09-03
Epoching module spam prevention


jhBharvest:
## Summary
RFC041(https://github.com/babylonlabs-io/pm/blob/main/rfc/rfc-041-epoching-spam-protection.md) proposes few measures to fight with epoching spam problem.

## Problem Definition

The epoching module batches staking-related messages (MsgDelegate, MsgUndelegate, MsgBeginRedelegate, etc.) and executes them at the last block of each epoch, at which point the validator set and voting power are updated.

###Vulnerabilities

**Unbounded queue size**

There is no cap on the per-epoch message queue. An attacker can enqueue arbitrarily many messages during an epoch, creating a spam/DoS vector.

**Incorrect gas accounting**

Gas is charged only when messages are enqueued. The (potentially heavy) execution at epoch end is not accounted for, enabling free execution and wasting validator resources.

**No immediate fund lock on MsgDelegate**

Delegation funds are not locked/escrowed at enqueue time. Accounts without sufficient tokens can submit unlimited MsgDelegate messages, enabling spam and deferred failures at execution time.

### Introduce an Epoch Delegation Pool

Create a per-epoch escrow pool that temporarily locks delegation funds at enqueue time, preventing spam by requiring stake up front.

Messages that deposit into this pool are checkpointing.WrappedCreateValidator and epoching.WrappedDelegate.

### Add a Minimum Delegation Amount

Introduce a chain parameter minDelegationAmount to reject dust-sized delegations and reduce micro-spam.

### Pre-charge Execution Gas

Compute the worst-case gas required to execute queued messages at the end of the epoch, and charge this amount upfront when the message is enqueued. This aligns costs with resource usage and mitigates execution-time DoS.

PR #1739, opened 2025-09-22
[backport] chore: epoching module spam prevention(#1663)(#1703)

jeoghyukLee123:
# Description

Closes: #1662 

Add upgrade handler logic to enable epoching spam prevention.

**Testnet:** upgrade handler applied in v4

**Mainnet:** upgrade handler applied in v4

## Notes
**Upgrade must occur at the first block of the epoch; if the block number is incorrect, the upgrade fails.**

- Add delegation_pool.go : The module manages fund lock / unlock related works
- Add delegation_pool_test.go : unit test for fund lock / unlock
- Add module parameters for execution gas
- Apply GasMeter().ConsumeGas in epoching / checkpointing module tx
- Add unit test for execution gas
- Modify Endblock of epoching module : When epoch ends, it unlocks fund and executes the message
- Add v4/upgrades.go: upgrade logic for testnet.
- Add v4/upgrades_test.go: upgrade test code.
- Add validation.go: common validation logic for upgrade.
- Modify e2e/configurer/upgrade.go: fix ChainConfig.UpgradeHeight to first block of epoch.

## Progress check list

- [x] Implementation upgrade handler for v4

### DIFF
--- a/app/include_upgrade_testnet.go
+++ b/app/include_upgrade_testnet.go
@@ -12,6 +12,7 @@ import (
 	v1rc9 "github.com/babylonlabs-io/babylon/v4/app/upgrades/v1rc9/testnet"
 	v2 "github.com/babylonlabs-io/babylon/v4/app/upgrades/v2"
 	v22 "github.com/babylonlabs-io/babylon/v4/app/upgrades/v2_2"
+	v23 "github.com/babylonlabs-io/babylon/v4/app/upgrades/v2_3"
 	v2rc4 "github.com/babylonlabs-io/babylon/v4/app/upgrades/v2rc4/testnet"
 	v4 "github.com/babylonlabs-io/babylon/v4/app/upgrades/v4"
 )
@@ -21,6 +22,8 @@ import (
 func init() {
 	Upgrades = []upgrades.Upgrade{
 		v4.Upgrade,
+		v23.Upgrade,
+		v22.Upgrade,
 		v2rc4.Upgrade,
 		v2.CreateUpgrade(true, map[string]struct{}{}),
 		v1_1.Upgrade,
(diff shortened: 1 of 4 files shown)


Lazar955 (approved):
lgtm! Let's wait for others reviews

app/upgrades/v4/upgrades.go
```diff
@@ -0,0 +1,67 @@
...
+			return nil, fmt.Errorf("failed to run migrations: %w", err)
+		}
+
+		if err := epoching.ValidateMigrationResults(ctx, keepers); err != nil {
+			return nil, fmt.Errorf("migration validation failed: %w", err)
+		}
+		// Log successful upgrade
+		sdkCtx.Logger().Info("spam prevention upgrade completed successfully",
+			"upgrade", UpgradeName,
+			"epoching_migration", "v3rc3->v3rc4",
```
KonradStaniec, line 60: I guess we should not hardcode version here

app/include_upgrade_testnet.go
```diff
@@ -13,12 +13,14 @@ import (
...
 	v2rc4 "github.com/babylonlabs-io/babylon/v4/app/upgrades/v2rc4/testnet"
+	v4 "github.com/babylonlabs-io/babylon/v4/app/upgrades/v4"
 )
 
 // init is used to include v1 upgrade testnet data
 // it is also used for e2e testing
 func init() {
 	Upgrades = []upgrades.Upgrade{
+		v4.Upgrade,
 		v2rc4.Upgrade,
```
KonradStaniec, line 24: I guess given that we inted to modify testnet to match mainnet, we will need to add
```go
		v23.Upgrade, // same as v3rc3 testnet
		v22.Upgrade,
```
upgrades here. Wydt @jhBharvest @RafilxTenfen @GAtom22 ?

GAtom22: yes, we'll need to add these at some point. Either in this PR or a separated one

PR #1763, opened 2025-09-29
fix: update upgrade logic for co-staker init

GAtom22:


app/upgrades/v4/upgrades.go
```diff
@@ -251,31 +294,42 @@ func saveCoStakersToStore(
...
+		rt = existing
+		rt.ActiveSatoshis = rt.ActiveSatoshis.Add(btcAmount)
+		rt.ActiveBaby = rt.ActiveBaby.Add(babyAmount)
 	}
 
-	// make sure current rewards is initialized
-	if _, err := k.GetCurrentRewardsInitialized(ctx); err != nil {
-		return err
+	// Update score
+	dp := costktypes.DefaultParams()
```
KonradStaniec, line 324: We can have one parameter instance per upgrade handler to avoid allocating so many of them

app/upgrades/v4/upgrades.go
```diff
@@ -209,40 +222,73 @@ func getBabyStakingAmount(ctx context.Context, stkQuerier stkkeeper.Querier, sta
...
-	return totalBaby, nil
+	return nil
 }
 
-// buildCoStakersMap builds a map of co-stakers from BTC stakers and their baby staking amounts
-func buildCoStakersMap(ctx context.Context, btcStakers map[string]math.Int, stkKeeper *stkkeeper.Keeper) (map[string]coStaker, error) {
-	stkQuerier := stkkeeper.NewQuerier(stkKeeper)
-	coStakers := make(map[string]coStaker)
+// saveBTCStakersRwdTracker retrieves all active BTC stakers with pagination and saves them to the costaker rewards tracker
+func saveBTCStakersRwdTracker(ctx context.Context,
```
RafilxTenfen, line 229:
```suggestion
func saveBTCStakersRwdTracker(
    ctx context.Context,
	cdc codec.BinaryCodec,
```

RafilxTenfen (approved):
LGTM, good job @GAtom22 🎉 

Can we add/modify e2e test that runs the upgrade where we simulate the error scenario:

1. Create an BTC delegation for an user that does not have baby delegations
2. Upgrade
3. Checks that reward tracking was created for that user
4. Jail/slash the finality provider 
5. Wait for a few blocks 

> It can be done in another PR as well if you think it is better

PR #1780, opened 2025-10-06
fix: claim rewards when updating score

RafilxTenfen:
Fix costaking MsgUpdateParams where it changes the score to baby ratio amount. Now it increments the current reward period, calculate and send rewards for those costakers that need and updates the current reward total score

V4-CORE-001 => Attacker can steal all rewards after score ratio update


KonradStaniec (approved):

GAtom22 (approved):
LGTM! Thanks @RafilxTenfen!! :pray:

PR #1783, opened 2025-10-06
fix: properly reduces the baby staking active amount after complete removal 

RafilxTenfen:
- Calls the same function as `AfterDelegationModified`


RafilxTenfen:
missing tests

PR #1790, opened 2025-10-07
fix: withdraw no rwd btc

RafilxTenfen:
Withdraw rewards of BTC_STAKE gauge also withdraws for COSTAKER gauge, but if there was no rewards available in BTC_STAKER it was throwing error to the user, when it should check if there isn't COSTAKER reward gauge. 

Modified to withdraw both and only error if there isn't any reward to be withdraw in the gauge


KonradStaniec (approved):
Looks good 👍 maybe we can add replay test for this ?

PR #1792, opened 2025-10-08
fix: baby staking hooks in costaking deletes StkCache after using

RafilxTenfen:
Issue case: 

1. Del1 baby stake to val1 10BABY
2. Del1 baby unbond from val1 10BABY
3. Del1 baby stake to val1 25BABY
4. Active BABY would result in 15BABY 

The stk cache was not being properly cleaned after using it. 

Fix is to clean the pair after using it


GAtom22 (approved):
LGTM! Thanks @RafilxTenfen!! :rocket:

PR #1801, opened 2025-10-11
chore: Update Testnet ID to bbn-test-6

filippos47:


RafilxTenfen:
@filippos47 the out-fps-testnet does not need to be changed, other changes will be done at https://github.com/babylonlabs-io/babylon/pull/1864

PR #1807, opened 2025-10-14
fix: remove dead nil-check for withdraw address and fix comment

radik878:
- Remove unreachable nil-check for withdraw address in withdrawReward
- Clarify comment to reflect defaulting behavior of GetWithdrawAddr


RafilxTenfen:
good catch, can you add changelog?

PR #1813, opened 2025-10-15
fix: add error and panic if there is any issue on removing delta sats

RafilxTenfen:
Audit V4-CORE-008


GAtom22 (approved):
LGTM! Thanks @RafilxTenfen!! :raised_hands:

PR #1859, opened 2025-11-06
Bump reusable workflows version

toroviej0x0:


.github/workflows/publish.yml
```diff
@@ -57,7 +57,7 @@ jobs:
               needs.lint_test.result == 'success'
             )
           ) }}
-    uses: babylonlabs-io/.github/.github/workflows/reusable_docker_pipeline.yml@5151754256060bf160c411d0784f831f29882106  # v0.13.4
```
RafilxTenfen: @toroviej0x0 can you use the commit hash of those versions? I remember there is some security recommendations for it

toroviej0x0: yes, tags are mutable

PR #1875, opened 2025-11-25
chore: ensure soft-deleted FPs cannot receive new/extended BTC stake, or commit pub rand

SebastianElvis:
## Summary

Fixes vulnerability where soft-deleted finality providers could still receive BTC delegations, stake expansions, and submit public randomness commitments, potentially halting finality.

Resolves https://github.com/babylonlabs-io/pm/issues/488

## Problem

Soft-deleted FPs were supposed to be excluded, yet Babylon still:
- Accepted new BTC delegations (validateStakedFPs never checked delete flag)
- Accepted stake expansions 
- Accepted pub rand commitments
- Assigned them active voting slots in power distribution

Since MsgAddFinalitySig correctly rejects deleted FP signatures, this created silent validators that couldn't vote, dropping quorum below ⅔ threshold and halting finality indefinitely.

## Solution

**Added delete check to 3 critical paths:**

1. **CreateBTCDelegation** - validateStakedFPs now checks IsFinalityProviderDeleted
2. **BtcStakeExpand** - validates FP is not deleted before accepting expansion
3. **CommitPubRandList** - rejects pub rand from deleted FPs

**New error:** ErrFpDeleted (code 1130) in x/btcstaking/types

**Tests:** Added unit test + updated all finality keeper tests with proper mocks

## Impact

- Prevents deleted FPs from gaining active voting slots
- Maintains network liveness when FPs are soft-deleted via governance
- Ensures soft-delete is effective at neutralizing misbehaving validators


GAtom22 (approved):
LGTM! Thanks @SebastianElvis!! :rocket:

PR #1877, opened 2025-11-25
test: improve e2ev2 multisig coverage

canu0205:
### Description

This pr targets to improve e2ev2 multisig test coverage, which contains:

- [x] complete multisig stake expansion flow (`CreateBTCDelegation`, `AddBTCDelegationInclusionProof`, `AddCovenantSigs`, `BtcStakeExpand`, `BTCUndelegate`)
- [x] multisig btc del -> single sig stake expansion (should fail)
- [x] single sig btc del -> multisig stake expansion (should fail)
- [ ] ~~multisig btc del slashed~~ (N/A)

### DIFF
--- a/test/e2ev2/multisig_test.go
+++ b/test/e2ev2/multisig_test.go
@@ -519,6 +519,8 @@ func startChainAndCreateFp(t *testing.T) (bbn2 *tmanager.Node, fpSK *btcec.Priva
 
 // buildMultisigDelegationWithSigCount construct multisig btc delegation msg
 // - sigsCount is the number of signatures
+//
+//nolint:unparam
 func buildMultisigDelegationMsgWithSigCount(
 	t *testing.T,
 	r *rand.Rand,


canu0205:
@RafilxTenfen so you mean FP is slashed and then check rewards of btc delegator who has delegated to the FP before slashing?

RafilxTenfen:
> @canu0205 lets also add one test where an multisig BTC delegation is slashed

What I mean is to slash the finality provider and composing the slashing transaction for this BTC multisig, thinking it right now this would not be possible in the e2e or e2ev2 as we don't have a real bitcoin chain running. 

We could add this case in https://github.com/babylonlabs-io/babylon-deployment, what do you think ?

canu0205:
yeah I also think sending bitcoin transaction in babylon e2e test is not available rn😅. 

sending multisig btc delegation slashing tx is done in vigilante e2e test tho. 

https://github.com/babylonlabs-io/vigilante/blob/feat/staker-multi-sig/e2etest/slasher_e2e_test.go#L209-L233

PR #1878, opened 2025-11-25
fix: remove unnecessary []byte allocations in IBC size checks

MozirDmitriy:
This change replaces len([]byte(...)) with len(...) for string size checks in validateIBCMsgTransfer and validateICAMsgSendTx within app/ante/ibc_msg_size.go. In Go, len(string) returns the number of bytes in the string’s UTF-8 representation without allocating, while converting to []byte creates a temporary slice and additional heap allocation. Since the limits are enforced in bytes and not runes, len(string) is both semantically correct and more efficient, eliminating avoidable allocations in the ante handler hot path.


canu0205 (commented):
good eye! @MozirDmitriy can u add a changelog entry?

RafilxTenfen (approved):

Issue #1870, opened 2025-11-19
[discussion] allow btc staker btc pks conversion when stake extension


canu0205:
### Summary

BTC Staker can extend their btc delegation by spending previous staking output on bitcoin. Stakers can extend their staking output with additional funding output, in addition to this, as btc multisig staking feature introduced, they can also convert signature scheme into multi signature from single signature and vice versa.

Purpose of this discussion is to decide on enabling signature scheme conversion while stake extension.

### Babylon signature scheme conversion

- `BtcStakeExpand` doesn't check if previous delegation is multisig or not.
- that is, both cases — prev delegation is single sig, stake extension is multi sig and vice versa — are available and btc delegation successfully stored in babylon.

### Bitcoin signature scheme conversion

- Stake extension is just a new transaction spending the old staking UTXO. As long as the new output script is valid and the inputs are correctly signed, bitcoin accepts it regardless of whether the previous output was single-sig or multi-sig.

PR #1882, opened 2025-11-27
fix: only allow stake extension when old and new delegation has the same staker btc pks

canu0205:
## Description

Closes: #1870


x/btcstaking/keeper/msg_server.go
```diff
@@ -165,6 +165,11 @@ func (ms msgServer) BtcStakeExpand(goCtx context.Context, req *types.MsgBtcStake
 		return nil, status.Errorf(codes.InvalidArgument, "the previous BTC staking transaction FP: %+v is not the same as FP of the stake expansion %+v", prevBtcDel.FpBtcPkList, req.FpBtcPkList)
 	}
 
+	// check that the previous delegation and the new expansion has the same staker btc pk
+	if err := validateStakerBtcPks(parsedMsg, prevBtcDel); err != nil {
```
RafilxTenfen: @GAtom22 shouldn't we validate that right now at main? the btc_pks of previous stk and extended stake should match

PR #1883, opened 2025-11-28
fix: validate btc pk is same with the prev del at btc stake exp

canu0205:
Closes: #1870 

related: #1882


canu0205:
after internal discussion, we've decided to include this pr in a new major version v5.x. cc. @RafilxTenfen @KonradStaniec

PR #1994, opened 2026-03-27
fix(finality): reject retroactive public randomness commitments

cgdusek:
## Summary

- Adds a lower-bound check in `CommitPubRandList` requiring `StartHeight > currentBlockHeight`, preventing finality providers from backfilling randomness for already-observed blocks
- Adds dedicated test cases (retroactive, equal-height, and future-height scenarios)
- Updates all e2e and unit tests to use dynamic/valid start heights

Closes #1984

## Test plan

- [x] New unit test case in `FuzzCommitPubRandList` verifying retroactive commits are rejected
- [x] New unit test case verifying `startHeight == currentHeight` is rejected
- [x] New unit test case verifying `startHeight == currentHeight + 1` succeeds
- [x] All existing test start heights updated to comply with the new constraint
- [ ] CI passes `make test` (blocked locally by pre-existing Go 1.26 / sonic dependency issue on main)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### State breaking
 
-- [#1984](https://github.com/babylonlabs-io/babylon/pull/1984) fix(finality): reject retroactive public randomness commitments in `CommitPubRandList`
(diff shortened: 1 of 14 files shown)


RafilxTenfen:
Authenticated commits are necessary to contribute

PR #2004, opened 2026-04-14
chore: add cl v4.2.7

RafilxTenfen:


.github/workflows/release.yml
```diff
@@ -26,12 +29,15 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GORELEASER_IMAGE: ghcr.io/goreleaser/goreleaser-cross:v1.27.0
+      RELEASE_TAG: ${{ inputs.tag || github.ref_name }}
+      RELEASE_DESCRIPTION: ${{ inputs.description || format('Release {0}', github.ref_name) }}
+      RELEASE_PRERELEASE: ${{ inputs.prerelease || contains(github.ref_name, '-rc') || contains(github.ref_name, '-alpha') || contains(github.ref_name, '-beta') }}
```
toroviej0x0: lets reserve `-rc` only, did we release `-alpha/-beta` before?

.github/workflows/release.yml
```diff
@@ -26,12 +29,15 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GORELEASER_IMAGE: ghcr.io/goreleaser/goreleaser-cross:v1.27.0
+      RELEASE_TAG: ${{ inputs.tag || github.ref_name }}
```
toroviej0x0, line 32: We should create release from ref. Only for tags.

toroviej0x0:
Remember to backport these changes to main.

.github/workflows/release.yml
RafilxTenfen: Those two requirements conflict — inputs.tag doesn't exist on push events. Two clean options:                                                                                               
                                                                       
  (a) Drop the push: tags trigger here; add a tiny dispatcher workflow that on tag push calls gh workflow run release.yml -f tag=$GITHUB_REF_NAME.                                            
                                                                                                                                                                                              
  (b) Keep push: tags here but accept that something has to resolve the tag on push — which means either a fallback expression or a step that sets it from github.ref_name.

toroviej0x0 (approved):

Lazar955 (approved):

